PR #27, opened 2019-09-17
Updating travis.yml so that checks pass

tanishqaggarwal:
This is so that checks pass when updating PRs.


tanishqaggarwal:
Closing because the unit tests also need to be fixed. I'll just fix this along with dev/usb-console.

PR #30, opened 2019-09-23
SBDWB Parsing issue resolved (Test/quake)

athena255:


lib/Drivers/QLocate.cpp
```diff
@@ -138,9 +146,7 @@ int QLocate::run_sbdix() {
 }
 
 // Parses the result buffer of sbdix into sbdix_r
-bool parse_ints(char const *c, int *i) {
-    Serial.println(c);
```
athena255: These were debug prints

lib/Drivers/QLocate.cpp
```diff
@@ -111,16 +111,24 @@ int QLocate::sbdwb(char const *c, int len) {
     Serial.print((char)(s >> 8), HEX);
     Serial.println((char)s, HEX);
 #endif
+
     // Process QLocate response
-    char buf[4];
-    len = port->readBytes(buf, 3);
-    buf[len] = '\0';
+    char buf[6];
```
athena255: There are three possible responses to SBDWB: 0\r\n0\r, 1\r\n, 2\r\n0\r, 3\r\n0\r
In all cases except for timeout, an extra 0\r is sent. I changed this to account for that. 
In the case of a timeout, 1\r\n will be sent after 60 seconds. The port->readBytes timeout is sooner than that, so if there is a timeout, the length of bytes written should be 0. We should still wait to consume the 1\r\n because otherwise, it will cause problems the next time we try to consume. 

An alternative solution would be to clear the buffer at the port before sending any command in which we expect to receive a response.

lib/Drivers/QLocate.cpp
```diff
@@ -111,16 +111,24 @@ int QLocate::sbdwb(char const *c, int len) {
...
+
 #ifdef DEBUG_ENABLED
     Serial.print("        > res=");
     for (int i = 0; i < len; i++) Serial.print(buf[i], HEX);
     Serial.println("\n        > return=" + String(*buf));
 #endif  
-    Serial.flush();
+    // Example response: 0\r\n0\r
+    if (len < 5) {
+        delay(60*1000);
```
athena255: Wait 60 seconds to consume 1\r\n

platformio.ini
```diff
@@ -140,59 +142,65 @@ src_filter = +<FCCode/> -<FCCode/main.cpp> +<native.cpp>
 # Quake
 ##########
 [env:teensy_36_test_quake_no_network]
-platform = teensy
 board = teensy36
-framework = arduino
+platform = ${teensy_cli.platform}
```
athena255: Changed these to be consistent with the rest of the environments

test/test_quake/test_quake.cpp
```diff
@@ -24,12 +24,12 @@ void test_sbdwb(void) {
     TEST_ASSERT_EQUAL(WRITE_OK, q.sbdwb(testString.c_str(), testString.length()));
     delay(DEFAULT_DELAY);
     // Send the same message again to verify that it's possible to do 
-    // (it should be, since MO queues the messages.)
+    // (it should be, since MO overwrites messages)
     TEST_ASSERT_EQUAL(WRITE_OK, q.sbdwb(testString.c_str(), testString.length()));
     delay(DEFAULT_DELAY);
 
-    std::string timeoutMsg ("Test write timeout");
```
athena255: Changed name since we cannot test timeout

lib/Drivers/QLocate.cpp
tanishqaggarwal: For flight control loop timing to be deterministic, I think it would be better to just not consume the 1 and instead clear the buffer. Were you having trouble with getting port->clear() to work?

tanishqaggarwal: Actually, scratch that. Since SBDIX might take some time, we have to let that thing sort of run in parallel anyway, so we might as well let SBDWB run in parallel as well. We'll have to restructure the driver a bit to achieve parallelism with the control loop.

PR #39, opened 2019-10-16
Dev/driver reorg

tanishqaggarwal:
Reorganizing how the drivers are written.


tanishqaggarwal:
@stewartaslan @kkrol27 I tagged you guys because all of this driver code needs a rigorous pinout review.

lib/Drivers/Gomspace.hpp
```diff
@@ -11,7 +11,8 @@ namespace Devices {
  */
 class Gomspace : public I2CDevice {
    public:
-    static constexpr unsigned char ADDRESS = 0x02; /**< I2C address of Gomspace device **/
+    static constexpr i2c_t3& wire = Wire; /**< I2C bus of Gomspace device **/
```
kkrol27: Does this not complain at all? I'm just curious because I remember we were having static initialization issue with the `Wire` objects before. Surprised the compiler doesn't complain about the `constexpr`.

lib/Drivers/PropulsionSystem.cpp
```diff
@@ -0,0 +1,66 @@
...
+
+float PropulsionSystem::get_pressure() {
+    return 0;
+}
+signed int PropulsionSystem::get_temp_inner() {
+    return 0;
+}
+signed int PropulsionSystem::get_temp_outer() {
+    return 0;
+}
```
kkrol27: We have this written down as a TODO and have a plan to get data to make lookup tables? I didn't realize this was still an open task.

lib/Drivers/PropulsionSystem.cpp
```diff
@@ -0,0 +1,66 @@
...
+
+bool PropulsionSystem::is_functional() { return is_enabled; }
+
+void PropulsionSystem::set_valves(const std::array<unsigned char, 4> &setting) {
+    for (unsigned char i = 2; i < 6; i++) { digitalWrite(valve_pins[i], setting[i]); }
+}
```
kkrol27: I'm assuming the finer control for the actual thruster firings happens outside of this loop? You then need to specify every valve each time you update?

**EDIT**: Also this needs to take a `std::array<unsigned char, 6> &` cuz we have six valves! (could make it `const` as well)

**EDIT**: It may be worth adding finer control here (i.e. the ability to actuate single valves). Most of the time we'll just be opening and closing a tank to tank valve to repress probably. @saa243 what do you think?

lib/Drivers/PropulsionSystem.hpp
```diff
@@ -0,0 +1,68 @@
...
+ *
+ * SpikeAndHold sph;
+ * sph.setup();
+ * **/
+class PropulsionSystem : public Device {
+   public:
+    //! Specifies which valve # corresponds to which physical valve
+    enum valve_ids {
+        intertank_main = 0,    // Main tank 1 to tank 2 valve
+        intertank_backup = 1,  // Backup tank 1 to tank 2 valve
```
kkrol27: We could also incorporate the logic to swap tank to tank valves into the driver? Although that could make `is_functional` checks a little confusing...

kkrol27:
Just a general comment as well, it may be worth taking a look at `std::string_view` if you are working with a bunch of compile time defined `char const *` variables. It's `constexpr` unlike `std::string` and doesn't require malloc.

lib/Drivers/PropulsionSystem.cpp
tanishqaggarwal: I think Stewart has this done but I lost the code

lib/Drivers/PropulsionSystem.cpp
tanishqaggarwal: Finer control happens outside the loop, yep. The actual execution of this function takes less than a microsecond, so it's safe to have an interrupt run every 1 or 2 ms to update the valve states.

I wanted to separate the tank valve control from the thrust valves, but I forgot to put in the tank valve control functions! I'll do that in a new commit

lib/Drivers/PropulsionSystem.hpp
tanishqaggarwal: It's better to have flight software control which tank valve is used, since the choice of tank valve should be operator-commandable.

lib/Drivers/Gomspace.hpp
tanishqaggarwal: No complaints...yeah now that you mention it, it is strange.

@fatimayousuf will be exploring the Gomspace driver in the next few weeks, so we should have answers on that soon. Thus far constexpr hasn't caused problems.

tanishqaggarwal:
Closing the PR for now due to significant changes required.

kkrol27:
@tanishqaggarwal What else needs to be updated?

tanishqaggarwal:
Tank valve control functions, replacing stubs for methods with actual content.

PR #46, opened 2019-10-25
Driver reorganization

tanishqaggarwal:
Reorganizes the propulsion, docking, and DCDC drivers so that they can be used as one cohesive unit by flight software. Also makes some cosmetic adjustments to Gomspace.


tanishqaggarwal:
I need to make changes to make the Docking System compatible with flight software. Currently it has a while loop that might take a lot of time to run.

PR #50, opened 2019-11-01
Controller/quake

athena255:
Functions not tested on Quake. 

- Controller implementation
- Removed QuakeMessage


test/test_quake_network/test_qlocate_network.cpp
```diff
@@ -1,79 +1,129 @@
...
 /* Tests that we can read messages from MT queue */
-void test_sbdrb_with_network(void) {
+void test_sbdrb_with_network(void)
+{
+    int numBytesRead;
+    char buf[16];
+    while (!Serial.available())
+        ;
+
+    // do
```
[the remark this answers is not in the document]
athena255: we need those if we want to use the python script

test/test_quake_no_network/test_qlocate_no_network.cpp
```diff
@@ -6,50 +6,184 @@
 #include "core_pins.h"
 #include "usb_serial.h"
 
+// #define DEFAULT_DELAY 10
```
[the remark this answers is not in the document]
athena255: The commented out stuff is the original no network test. The network test would not compile it

lib/Drivers/QLocate.hpp
```diff
@@ -29,32 +95,34 @@ class QLocate : public Device {
     /*! Sets the QLocate serial port and serial timeout value. Do not Initialize
      *  the serial port with begin(), it will be done in the constructor.
      */
+#ifndef DESKTOP
     QLocate(const std::string &name, HardwareSerial *const port, unsigned char nr_pin, int timeout);
-
+#else 
+    using String = std::string;
```
shihaocao: what is the purpose of `using String = std::string;`?

src/FCCode/QuakeControlTask.h
```diff
@@ -0,0 +1,77 @@
+#pragma once
```
shihaocao: Why is this file QuakeControlTask.h instead of QuakeControlTask.hpp?

tanishqaggarwal: It doesn't matter, but it would have been more consistent to name it QuakeControlTask.hpp

lib/Drivers/QLocate.hpp
tanishqaggarwal: It just makes declaring variables an epsilon less cumbersome, but @athena255, do you use this anywhere?

tanishqaggarwal: If not, I think it'd be better if any std:: objects were explicitly named as being part of std::

PR #51, opened 2019-11-01
Dev/piksi test

shihaocao:
This merge request will merge in the new Piksi driver, which is able to successfully read about half of all Piksi message bursts (assuming a 120 ms overall control cycle). The main new method: read_buffer() processes data in the serial RX buffer within 300 microseconds (experimentally guaranteed). You will still need to call getters to update state variables.


platformio.ini
```diff
@@ -131,6 +131,19 @@ test_filter = test_fsw_*
...
 
+##########
+# Piksi
+##########
+[env:teensy_36_test_piksi]
+board = teensy36
+platform = ${teensy_cli.platform}
+framework = ${teensy_cli.framework}
+lib_deps = ${teensy_cli.lib_deps}
+build_flags = -std=c++14 -Wall -D UNITY_INCLUDE_DOUBLE -D PAN_LEADER -D SERIAL4_RX_BUFFER_SIZE=1024
```
tanishqaggarwal: Can you add the buffer size compiler flag to the other non-test environments too? I think you'd just need to modify teensy_common

src/piksi_test.cpp
```diff
@@ -0,0 +1,16 @@
+#include <Arduino.h>
+#include <Piksi.hpp>
+#include <HardwareSerial.h>
+
+Devices::Piksi piksi("piksi", Serial2);
```
tanishqaggarwal: Serial4?

tanishqaggarwal: Actually can you just delete this file? It doesn't do anything.

lib/Drivers/Piksi.hpp
```diff
@@ -38,69 +38,135 @@ class Piksi : public Device {
     /** @brief Runs read over UART buffer to process values sent by Piksi into
      * memory.
      *  @returns Whether or not any data was processed. **/
-    virtual bool process_buffer();
```
tanishqaggarwal: None of these functions don't need to be virtual anymore. The reason they were virtual was to allow the functions to be overwritten in a base class for a FakePiksi class, but this isn't necessary anymore.

lib/Drivers/Piksi.hpp
```diff
@@ -38,69 +38,135 @@ class Piksi : public Device {
...
+    virtual signed char process_buffer();
```
tanishqaggarwal: The process_buffer functions shouldn't be called anymore, right? Can you make them protected functions, in order to make this clear to a user?

lib/Drivers/Piksi.hpp
```diff
@@ -38,69 +38,135 @@ class Piksi : public Device {
...
+     * @return return code, 0 if buffer was successfully processed (to the knowledge of the driver)
+     * 1 if buffer already has parts of multiple bursts of data, or an incomplete burst
+     * 2 if buffer is interrupted with new data
+     */
+    virtual unsigned char read_buffer();
 
     /** @brief Gets GPS time.
      *  @return GPS time, as nanoseconds since the epoch. **/
-    virtual void get_gps_time(gps_time_t *time);
+    // virtual void get_gps_time(gps_time_t *time);
```
tanishqaggarwal: Remove this commented line.

lib/Drivers/Piksi.hpp
```diff
@@ -38,69 +38,135 @@ class Piksi : public Device {
...
      *  @return Number of satellites used for determining GPS velocity. **/
     virtual unsigned char get_vel_ecef_nsats();
+
     /** @brief Get status flags of GPS velocity measurement.
      *  @return Status flags of GPS velocity measurement. **/
     unsigned char get_vel_ecef_flags();
 
     /** @brief Gets base station position in ECEF coordinates.
-     *  @param position A pointer to an array for storing the x,y,z coordinates of
+     *  @param position A pointer to an std::array of doubles for storing the x,y,z coordinates of
```
tanishqaggarwal: Improve argument description

test/test_piksi/test_piksi.cpp
```diff
@@ -0,0 +1,382 @@
+#include <core_pins.h>
```
tanishqaggarwal: Can you clean this file up by a) removing unnecessary comments, and b) adding documentation for the helper functions and variables that you use in the test file?

Also, it would be helpful to move the variables closer to where they're actually used.

test/test_piksi/test_piksi.cpp
```diff
@@ -0,0 +1,382 @@
...
+#include <unity.h>
+#include <usb_serial.h>
+#include <array>
+#include "../lib/Drivers/Piksi.hpp"
+
+#define PIKSI_READ_ALLOTED 300
```
tanishqaggarwal: Don't use macros; use static const unsigned ints instead.

tanishqaggarwal:
Closing PR due to more development work needed.

PR #52, opened 2019-11-05
Deadline manager

tanishqaggarwal:
This contains extensions to control task for timing, and some basic unit testing for the capability.


tanishqaggarwal:
NOTE: I removed the testing for this item from CI because it doesn't work. Travis CI's virtual machines make the timing too unreliable. For testing, try it on your local machine instead.

src/FCCode/TimedControlTask.hpp
```diff
@@ -0,0 +1,127 @@
...
+     * 
+     * @param registry State field registry
+     * @param offset Time offset of start of this task from the beginning of a
+     *               control cycle, in microseconds.
+     */
+    TimedControlTask(
+                      StateFieldRegistry& registry,
+                      const unsigned int _offset) :
+        ControlTask<T>(registry), offset(us_to_duration(_offset + 1))
+    {}
```
kkrol27: Just wanna make sure this formatting is intentional - doesn't really matter, just thought it was a bit goofy.

src/FCCode/TimedControlTask.hpp
```diff
@@ -0,0 +1,127 @@
...
+    /**
+     * @brief Cause the system to pause operation until a system time is reached.
+     * 
+     * TODO check for errors that could happen on Teensy due to integer overflow.
+     * 
+     * @param time Time until which the system should pause.
+     */
+    void wait_until_time(const sys_time_t& time) const {
+      #ifdef DESKTOP
+        while((signed int) duration_to_us(time - get_system_time()) > 0);
```
shihaocao: To clarify, this code will let us execute the control cycle as soon as possible. If for some ungodly reason 1 control cycle takes 160 ms. Do we want to wait until the next possible in sync time arrives (an additional 80 ms) for a perfect 240 ms spacing?

tanishqaggarwal: I realized that we don't need to if nothing is significantly affected. Integrator stepsizes should not be preturbed significantly for one funky control cycle. Only thing that might care a bit is Piksi--will it be fine?

src/FCCode/TimedControlTask.hpp
tanishqaggarwal: Goofy formatting indeed

src/FCCode/ClockManager.cpp
```diff
@@ -1,12 +1,28 @@
...
     control_cycle_count_sr(0, 4294967295, 32),
     control_cycle_count_f("pan.cycle_no", control_cycle_count_sr)
 {
     add_readable_field(control_cycle_count_f);
 }
 
 void ClockManager::execute() {
+    if (has_executed) {
+        sys_time_t earliest_start_time = control_cycle_start_time + control_cycle_size;
+        wait_until_time(earliest_start_time);
```
athena255: why are we waiting during the execute?

src/FCCode/ClockManager.hpp
```diff
@@ -1,14 +1,58 @@
...
+     * @param control_cycle_size Control cycle time, in nanoseconds if on desktop,
+     * and in microseconds if on Teensy.
+     */
+    ClockManager(StateFieldRegistry& registry, const unsigned int control_cycle_size);
+
+    /**
+     * @brief Increments the control cycle count. 
+     * 
+     * Also ensures that the control cycle count can only be incremented 
+     * at least control_cycle_size microseconds after the last execution
```
athena255: Why does it matter that it can only be incremented control_cycle_size microseconds after the last execution? Won't we just increment it whenever we finish the entire cycle?

tanishqaggarwal: The clock manager is controlling the length of a control cycle. We want to ensure that control cycles are at least their minimum length in order to ensure timing guarantees.

src/FCCode/ClockManager.cpp
tanishqaggarwal: Answered on your other comment.

src/FCCode/TimedControlTask.hpp
shihaocao: Should be fine. If we get one stale packet from Piksi, we'll have clean data within a few control cycles.

PR #55, opened 2019-11-06
ADCS Control Tasks

tanishqaggarwal:
This is a draft of the interface for all of the control tasks that are needed for attitude determination and control.

I need your input for any fields I might be missing or that might be extraneous.


lib/Devices/I2CDevice.hpp
```diff
@@ -37,7 +37,9 @@
 /* Note that I2C_AUTO_RETRY should be enabled - I2CDevice makes no calls to
```
tanishqaggarwal: Just modifying I2C to work with desktop compilations of Flight Software here.

src/FCCode/ADCSBoxController.hpp
```diff
@@ -0,0 +1,38 @@
...
+   void execute() override;
+
+  protected:
+   /** ADCS Driver. **/
+   Devices::ADCS& adcs_system;
+
+   /**
+    * @brief Outputs to send to ADCS box.
+    */
+   ReadableStateField<bool> adcs_box_mode_f;
```
tanishqaggarwal: As a reminder: Readable means that the field is read-only **from the ground**. Other control tasks are free to modify the value.

tanishqaggarwal: I think it would be nuts to allow us to be able to do that from the ground. In fact I'm not able to remember why we have this option on the ADCS box. @kkrol27 thoughts?

shihaocao: I thought there might be a condition where the EPS disables the ADCS box, so wouldn't it be useful to be able to turn it back on? Or would this enabling/disabling take place somewhere else?

src/FCCode/ADCSBoxController.hpp
```diff
@@ -0,0 +1,43 @@
...
+   ReadableStateField<f_vector_t> rwa_torque_vec_f;
+   ReadableStateField<f_vector_t> rwa_speed_vec_f;
+   ReadableStateField<f_vector_t> mtr_cmd_vec_f;
+   ReadableStateField<float> gyr_temp_f;
```
stewartaslan: This kinda seems like we're sending a gyro temp to the box, when we're actually sending a setpoint. Let's change the name of this to be gyr_temp_set_f or something.

src/FCCode/ADCSBoxController.hpp
```diff
@@ -0,0 +1,43 @@
...
+   // We don't expect these values to change very often, but we
+   // send them on every control cycle anyway.
+   WritableStateField<float> rwa_momentum_filter_f;
```
stewartaslan: What is this?

src/FCCode/ADCSBoxController.hpp
```diff
@@ -0,0 +1,43 @@
...
+   WritableStateField<float> ramp_filter_f;
+   WritableStateField<float> ssa_voltage_filter_f;
```
stewartaslan: If this is the voltage threshold for sun sensors, that's good. If it's something else, we should add a field for the voltage threshold.

src/FCCode/ADCSBoxController.hpp
```diff
@@ -0,0 +1,43 @@
...
+   WritableStateField<float> mag_filter_f;
+   WritableStateField<float> gyr_filter_f;
+   WritableStateField<float> gyr_temp_kp_f;
+   WritableStateField<float> gyr_temp_ki_f;
+   WritableStateField<float> gyr_temp_kd_f;
+   WritableStateField<float> mtr_limit_f;
+   WritableStateField<unsigned int> imu_mode_f;
```
stewartaslan: Other things we need to add (if they're not already here with different names)
- Wheel speed setting (like the target speed that the wheels are ramping to)
- MTR PWM value to control them (unless this is handled by the MTR driver)
- MTR direction (same deal as above, really the question is what are we passing to the MTR driver)

src/FCCode/ADCSBoxMonitor.hpp
```diff
@@ -0,0 +1,48 @@
...
+
+   /**
+    * @brief Inputs to get from ADCS box.
+    */
+   //! Speed reads off of the reaction wheels.
+   ReadableStateField<f_vector_t> rwa_rd_f;
+   //! Vector to the sun in the body frame.
+   ReadableStateField<f_vector_t> ssa_vec_f;
+   //! Confidence in sun vector estimation.
+   ReadableStateField<bool> ssa_vec_valid_f;
```
stewartaslan: Is this enough info to properly communicate the validity of the sun vector? Or do we want to send all 20 voltage values? @kkrol27 @nhz2

src/FCCode/ADCSBoxController.hpp
tanishqaggarwal: I don't think we have a driver function for setting this threshold. Can't this value just be a static constant on the ADCS box?

stewartaslan: No we need to be able to modify it, sort of like changing the gain on the sun sensors.

src/FCCode/ADCSBoxMonitor.hpp
```diff
@@ -0,0 +1,48 @@
...
+   //! Magnetic field vector in the body frame.
+   ReadableStateField<f_vector_t> mag_vec_f;
```
stewartaslan: How are we handling the two magnetometers? Shouldn't we get the vector from both of them?

src/FCCode/ADCSBoxController.hpp
kkrol27: It's called like `gyr_temp_eq_f` (equilibrium on ADCS).

src/FCCode/ADCSBoxController.hpp
kkrol27: The exponential filter constant. Sets up the filters on ADCS.

src/FCCode/ADCSBoxController.hpp
kkrol27: Taking this conversation into slack

src/FCCode/ADCSBoxController.hpp
kkrol27: MTR PWM isn't needed. And the MTR command vec handles direction.

src/FCCode/ADCSBoxMonitor.hpp
kkrol27: We have the ability to read in all the voltage values. That's just for telemetry purposes however.

src/FCCode/ADCSBoxMonitor.hpp
kkrol27: No, you select which one to enable. We were treating it as more of a redundant sensor in case of a hardware failure.

src/FCCode/ADCSBoxController.hpp
kkrol27: The 'disabled' mode only prevents actuator operation; all passive sensors are still running. We would be turning off the ADCS box potentially in low power/safehold mode. I don't think we would ever want to only turn off the ADCS box but try and run a nominal mission otherwise. So no, I'd say it doesn't need to be toggled from the ground.

kkrol27: We may want to power cycle ADCS however, in which case it would boot up in it's disabled mode again.

src/FCCode/ADCSBoxController.hpp
kkrol27: Also this should be an `unsigned char` (at least that's what it's sent as over I2C).

src/FCCode/ADCSBoxController.hpp
```diff
@@ -0,0 +1,68 @@
...
+#define ADCS_BOX_CONTROLLER_HPP_
+
+#include <ADCS.hpp>
+#include <ControlTask.hpp>
+
+/**
+  * @brief Writes outputs to the ADCS box from the values in the state fields
+  * listed below.
+  */
+class ADCSBoxController : public ControlTask<void> {
```
kkrol27: What's the template argument to `ControlTask<...>` for?

src/FCCode/ADCSBoxController.hpp
```diff
@@ -0,0 +1,68 @@
...
+
+   /**
+    * @brief Outputs to send to ADCS box.
+    */
+   //! Box mode (enabled or disabled.)
+   ReadableStateField<bool> adcs_box_mode_f;
+   //! Torque command to apply to reaction wheels.
+   ReadableStateField<f_vector_t> rwa_torque_vec_f;
+   //! Speed command to apply to reaction wheels.
+   ReadableStateField<f_vector_t> rwa_speed_vec_f;
```
kkrol27: I would maybe suggest combining these into a single vector and then having a `ReadableStateField<unsigned char> rwa_cmd_mode`?

kkrol27: That would more closely mirror how ADCS operates, and reduce memory usage. How do you know which type of command we want to send now?

src/FCCode/ADCSBoxMonitor.hpp
```diff
@@ -0,0 +1,48 @@
...
+  */
+class ADCSBoxMonitor : public ControlTask<void> {
+  public:
+   /**
+    * @brief Construct a new ADCSBoxMonitor
+    * 
+    * @param registry 
+    * @param adcs 
+    */
+   ADCSBoxMonitor(StateFieldRegistry& registry, Devices::ADCS& adcs);
```
kkrol27: Unsure if the ADCS driver with this style of `const` in mind, but could this input be `const Devices::ADCS& adcs`?

src/FCCode/ADCSBoxMonitor.hpp
```diff
@@ -0,0 +1,48 @@
...
+
+   /**
+    * @brief Gets inputs from the ADCS box and dumps them into the state
+    * fields listed below.
+    */
+   void execute() override;
+
+  protected:
+   /** ADCS Driver. **/
+   Devices::ADCS& adcs_system;
```
kkrol27: Same `const` questions here.

src/FCCode/ADCSBoxMonitor.hpp
kkrol27: After reading through this again, I'd say we do want to periodically read out all twenty voltage values. It shouldn't be every time, however, because that would take too long. Perhaps making this a separate control task is an easier way forward?

src/FCCode/AttitudeComputer.hpp
```diff
@@ -0,0 +1,72 @@
...
+
+    /**
+     * @brief Inputs, a.k.a estimator outputs.
+     */
+    //! True if the satellite is in eclipse.
+    std::shared_ptr<ReadableStateField<bool>> in_eclipse_f;
+    //! Spacecraft angular momentum vector, in the body frame.
+    std::shared_ptr<ReadableStateField<f_vector_t>> h_vec_body_f;
+    //! Quaternion representing the rotation from the body frame to the ECI frame.
+    std::shared_ptr<ReadableStateField<f_quat_t>> body_to_eci_f;
```
kkrol27: What's up with the use of `std::shared_ptr<...>` here and not anywhere else?

src/FCCode/AttitudeComputer.hpp
```diff
@@ -0,0 +1,72 @@
...
+     * @brief Computer outputs.
+     */
+    //! Desired attitude.
+    ReadableStateField<f_quat_t> cmd_attitude_f;
+    //! Reaction wheel torques to apply on the next timestep in order to start
+    //! reaching the desired attitude.
+    std::shared_ptr<ReadableStateField<f_vector_t>> rwa_torque_vec_f;
+    //! Reaction wheel speeds to apply on the next timestep in order to start
+    //! reaching the desired attitude.
+    std::shared_ptr<ReadableStateField<f_vector_t>> rwa_speed_vec_f;
```
kkrol27: Whatever the ADCS RWA command ends up looking like, that change/no change will need to be mirrored here.

src/FCCode/AttitudeComputer.hpp
```diff
@@ -0,0 +1,72 @@
...
+    //! Magnetorquer moment commands to apply on the next timestep in order to start
+    //! reaching the desired attitude.
+    std::shared_ptr<ReadableStateField<f_vector_t>> mtr_cmd_vec_f;
```
kkrol27: `std::shared_ptr<...>` again?

src/FCCode/AttitudeEstimator.hpp
```diff
@@ -0,0 +1,62 @@
...
+    std::shared_ptr<ReadableStateField<f_vector_t>> r1_vec_eci_fp;
+    //! Position of other satellite, in the ECI frame.
+    std::shared_ptr<ReadableStateField<f_vector_t>> r2_vec_eci_fp;
+    
+    /**
+     * @brief Constant parameters for calculation.
+     */
+    static const f_vector_t gyro_bias;   /**< Gyroscope bias **/
+    static const f_vector_t mag_bias;    /**< Magnetometer bias **/
+    static const std::array<float, 6> I; /**< Moment of inertia **/
```
kkrol27: I'm thinking some of this stuff may live a little deeper in GNC (at the very least the moment of inertia probably doesn't need to be defined here). Also, regardless, the biases aren't going to be `const`.

src/FCCode/AttitudeEstimator.hpp
```diff
@@ -0,0 +1,62 @@
...
+    //! Magnetic field vector in the body frame.
+    std::shared_ptr<ReadableStateField<f_vector_t>> mag_vec_body_rd_fp;
+    //! Gyroscope vector in the body frame.
+    std::shared_ptr<ReadableStateField<f_vector_t>> gyr_vec_body_rd_fp;
+    //! Reaction wheel speeds.
+    std::shared_ptr<ReadableStateField<f_vector_t>> rwa_rd_fp;
+    //! Position of current satellite, in the ECI frame.
+    std::shared_ptr<ReadableStateField<f_vector_t>> r1_vec_eci_fp;
+    //! Position of other satellite, in the ECI frame.
+    std::shared_ptr<ReadableStateField<f_vector_t>> r2_vec_eci_fp;
```
kkrol27: I think I may understand the `std::shared_ptr<...>` now. We're referencing the fields from other tasks (whatever is producing the value). Why not just make them references, however?

src/FCCode/ADCSBoxController.hpp
tanishqaggarwal: It's the return type of execute(). For most control tasks this is void, but for some it's useful for execute() to return a value (see: Athena's Quake controller PR.)

src/FCCode/ADCSBoxMonitor.hpp
tanishqaggarwal: It seems so, given that there are no member variables for the ADCS driver. I'll do that.

src/FCCode/AttitudeComputer.hpp
tanishqaggarwal: The convention is to use `shared_ptr`s to fields that are inputs (i.e. not created by the ControlTask).

src/FCCode/AttitudeComputer.hpp
tanishqaggarwal: They're inputs to the control task

src/FCCode/AttitudeEstimator.hpp
tanishqaggarwal: In order for them to be references, we'd have to pass in the ControlTasks in which they're created as arguments to the constructor of `AttitudeEstimator`. Moreover, we'd have to make the StateFields public members of the other ControlTasks. This is cumbersome, bad OOP practice, and makes the class less unit-testable (the most important point.)

Instead, by making them pointers, the addresses of these `StateField`s can be stored in a centralized object like `StateFieldRegistry` and initialized at runtime. This makes code more prone to runtime initialization errors, but honestly, these are so easy to catch by just doing some good unit testing.

Check out the structure for MissionManager, as an example.

src/FCCode/ADCSBoxController.hpp
tanishqaggarwal: Sorry, which field should be unsigned char?

src/FCCode/ADCSBoxMonitor.hpp
tanishqaggarwal: Reading 20 unsigned chars takes 312 microseconds on an ADCS bus. I think we can budget for that. I'm wary of doing things "periodically" because it makes each control cycle less deterministic.

src/FCCode/ADCSBoxMonitor.hpp
tanishqaggarwal: Marking as resolved.

src/FCCode/ADCSBoxController.hpp
tanishqaggarwal: Don't we want to send both speed and torque commands on every control cycle? Maybe I'm mistaken.

src/FCCode/AttitudeEstimator.hpp
tanishqaggarwal: Ok, I'll make the biases writable state fields and remove moment-of-inertia.

tanishqaggarwal: Ok, I'll make the biases writable state fields and remove moment-of-inertia.

src/FCCode/ADCSBoxController.hpp
tanishqaggarwal: Might as well make it a WritableStateField then. It'll be a "blacklisted" command but I guess we should still have the option of commanding the field.

src/FCCode/ADCSBoxController.hpp
kkrol27: The IMU mode I think.

src/FCCode/ADCSBoxMonitor.hpp
kkrol27: Okay yeah no perfect! If we can do it at every time step we might as well!

src/FCCode/ADCSBoxController.hpp
kkrol27: We’ll be interested din reading speed and accel ramp at every cycles however as far as commanding goes we’ll only every command speed or accel. You specify which you’re commanding by writing an rwa mode byte to adcs.

src/FCCode/ADCSBoxController.hpp
tanishqaggarwal: Gotcha. I made it an int because I haven't made a Serializer specialization for chars

src/FCCode/ADCSBoxController.hpp
tanishqaggarwal: Right, ok, makes sense. I'll make some adjustments for this interface

src/FCCode/AttitudeEstimator.hpp
kkrol27: I guess I just don't see how references/raw pointers couldn't accomplish the same things but it doesn't matter too much. The big benefit to `std::shard_ptr<...>` is to leverage RAII to handle memory allocation for use via reference counting in an allocated control block. We never deallocate anything so we could've just used raw pointers and skipped the extra mallocs and overhead?

src/FCCode/AttitudeEstimator.hpp
```diff
@@ -44,9 +44,8 @@ class AttitudeEstimator : public ControlTask<void> {
     /**
      * @brief Constant parameters for calculation.
      */
-    static const f_vector_t gyro_bias;   /**< Gyroscope bias **/
-    static const f_vector_t mag_bias;    /**< Magnetometer bias **/
-    static const std::array<float, 6> I; /**< Moment of inertia **/
+    WritableStateField<f_vector_t> gyro_bias_fp;   /**< Gyroscope bias **/
+    WritableStateField<f_vector_t> mag_bias_fp;    /**< Magnetometer bias **/
```
kkrol27: And writable here means the the attitude estimator can adjust it but nothing can change it from the outside right?

src/FCCode/AttitudeEstimator.hpp
tanishqaggarwal: Yeah...it was a big mistake to use shared_ptrs for referencing state fields, but now it's so deep-baked into CommonSoftware and FlightSoftware that ripping it out would be a huge time investment.

There's no significant malloc going on aside from the control block, which is minimal. And it happens only once upon initialization.

One (very minor) benefit of shared_ptrs is that it makes it abundantly clear if a state element is an input or an output to a control task. Sometimes that's easy to miss with a raw pointer.

src/FCCode/AttitudeEstimator.hpp
tanishqaggarwal: No, writable and readable refer to the **ground**'s ability to read or write to the field. This is intended to be writable so that we can change the biases if necessary.

As many fields should be writable as possible. Some are unavoidably readable, such as sensor readings and computational outputs, but variables that directly affect control decisions should be writable.

tanishqaggarwal:
RFC is over.

PR #57, opened 2019-11-06
Driver reorganization

tanishqaggarwal:
Reopening this PR. Has updates to all drivers and in particular fleshes out the propulsion and docking drivers.


test/test_docking/test_docking.cpp
```diff
@@ -0,0 +1,18 @@
+#include <unity.h>
```
tanishqaggarwal: Will be populated in #58

test/test_prop/test_prop.cpp
```diff
@@ -0,0 +1,18 @@
+#include <unity.h>
```
tanishqaggarwal: Will be populated in #58

test/test_docking/test_docking.cpp
tanishqaggarwal: Populating this also depends on @fatimayousuf's updates to the Docking System driver in order to get it to be compatible with flight software.

lib/Drivers/DockingSystem.cpp
```diff
@@ -0,0 +1,89 @@
...
+
+    enable();
+
+    return true;
+}
+
+bool DockingSystem::is_functional() {
+    return digitalRead(DCDC::dcdc_sph_enable_pin)
+        && digitalRead(motor_sleep_pin);
+}
```
kkrol27: I know this is all we can technically do, but just want to make sure we're all aware this isn't fool proof.

lib/Drivers/DockingSystem.cpp
```diff
@@ -0,0 +1,89 @@
...
+    unsigned int current_step = 0;
+
+    // step at a frequency of 500Hz
+    while (current_step < steps){
+        digitalWrite(motor_step_pin, LOW);
+        delayMicroseconds(1000);
+        digitalWrite(motor_step_pin, HIGH);
+        delayMicroseconds(1000);
+        current_step++;
+    }
```
kkrol27: This is doing an individual step at each iteration triggered by a rising or falling edge correct?

tanishqaggarwal: It's just briefly setting the motor PWM duty cycle to 100% and then to 0%.

This while loop is actually going to be removed on dev/docking_controller_2, in order to make the driver compatible with Flight Software.

PR #64, opened 2019-11-08
GNC Control Tasks

tanishqaggarwal:
Also creating this PR for awareness. Here's a first draft of an interface between GNC functions and FSW functions. Looking for recommendations!

### DIFF
--- a/src/FCCode/GNCNavigator.hpp
+++ b/src/FCCode/GNCNavigator.hpp
@@ -1,0 +1,40 @@
+#ifndef GNC_NAVIGATOR_HPP_
+#define GNC_NAVIGATOR_HPP_
+
+#include <ControlTask.hpp>
+
+class GNCNavigator : public ControlTask<void> {
+   public:
+    /**
+     * @brief Construct a new Orbit Estimator.
+     * 
+     * @param registry 
+     */
+    GNCNavigator(StateFieldRegistry& registry);
+
+    /**
+     * @brief Determine the magnitude, time, and direction of the next
+     * propulsive manuever.
+     */
+    void execute() override;
+
+   protected:
+    //! Estimated position and velocity of this satellite in the ECI frame.
+    std::shared_ptr<ReadableStateField<d_vector_t>> r1_vec_eci_fp;
+    std::shared_ptr<ReadableStateField<d_vector_t>> v1_vec_eci_fp;
+    //! Estimated position and velocity of other satellite in the ECI frame.
+    std::shared_ptr<ReadableStateField<d_vector_t>> r2_vec_eci_fp;
+    std::shared_ptr<ReadableStateField<d_vector_t>> v2_vec_eci_fp;
+
+    //! Navigator mode. If in "auto" mode (true), it publishes recommended firings to
+    //! the propulsion_cmd fields below as soon as it determines them. If in
+    //! "manual" mode, the ground has all authority on propulsive manuevers.
+    WritableStateField<bool> navigator_mode_f;
+
+    //! Either a propulsion manuever computed by the rendezvous algorithm, or a
+    //! propulsion manuever manually commanded by the ground.
+    WritableStateField<d_vector_t> propulsion_cmd_f;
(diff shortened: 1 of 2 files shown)


src/FCCode/OrbitEstimator.hpp
```diff
@@ -0,0 +1,45 @@
...
+
+    //! Position and velocity of this satellite, as read directly from the Piksi in the
+    //! ECEF frame.
+    std::shared_ptr<ReadableStateField<d_vector_t>> r1_vec_ecef_rd_fp;
+    std::shared_ptr<ReadableStateField<d_vector_t>> v1_vec_ecef_rd_fp;
+
+    //! Position and velocity of this satellite, as (either) read directly from the Piksi
+    //! or updated by the ground.
+    std::shared_ptr<ReadableStateField<d_vector_t>> r2_vec_ecef_rd_fp;
+    std::shared_ptr<ReadableStateField<d_vector_t>> v2_vec_ecef_rd_fp;
```
kkrol27: This will probably just come up in ECI. I'd don't see why we would pass it around in ECEF. Also it's very important these things be timestamped!

src/FCCode/OrbitEstimator.hpp
```diff
@@ -0,0 +1,45 @@
...
+
+    // TODO insert other state fields we need to know in order to determine orbital state.
+
+    //! Estimated position and velocity of this satellite in the ECI frame.
+    ReadableStateField<d_vector_t> r1_vec_eci_f;
+    ReadableStateField<d_vector_t> v1_vec_eci_f;
+    //! Estimated position and velocity of other satellite in the ECI frame.
+    ReadableStateField<d_vector_t> r2_vec_eci_f;
+    ReadableStateField<d_vector_t> v2_vec_eci_f;
```
kkrol27: We'll also want an estimated position in ECEF for this satellite as well (eventually at least). @nhz2 @stewartaslan you think it's too soon for that?

src/FCCode/GNCNavigator.hpp
```diff
@@ -0,0 +1,40 @@
...
+    //! Navigator mode. If in "auto" mode (true), it publishes recommended firings to
+    //! the propulsion_cmd fields below as soon as it determines them. If in
+    //! "manual" mode, the ground has all authority on propulsive manuevers.
+    WritableStateField<bool> navigator_mode_f;
+
+    //! Either a propulsion manuever computed by the rendezvous algorithm, or a
+    //! propulsion manuever manually commanded by the ground.
+    WritableStateField<d_vector_t> propulsion_cmd_f;
+    WritableStateField<gps_time_t> propulsion_cmd_time_f;
+};
```
kkrol27: This, for now at least, may take in directly the position of the other satellite in the hill frame of the first. Again something that may be a bit farther down the line but something to keep in mind.

src/FCCode/OrbitEstimator.hpp
kkrol27: Actually idk if this is possible, but it may be easier to house orbital state in something like this:

    struct Orbit {
      lin::Vector3d r_eci;
      lin::Vector3d v_eci;
      lin::Vector3d r_ecef;  // The ECEF frame's would may this state overrefined
      lin::Vector3d v_ecef;  // b/c you could convert between the two given t
      double t;
    };

tanishqaggarwal: Forgot about the timestamps! Thanks

src/FCCode/OrbitEstimator.hpp
tanishqaggarwal: Why do we want that?

src/FCCode/OrbitEstimator.hpp
tanishqaggarwal: Also, we have to pass it around in ECEF because the Piksi sends data in ECEF.

tanishqaggarwal:
RFC seems over.

PR #65, opened 2019-11-08
Dev/docking controller 2

fatimayousuf:


test/test_fsw_docking/test_fsw_docking.cpp
```diff
@@ -0,0 +1,79 @@
...
+#include <unity.h>
+
+class TestFixture {
+  public:
+    StateFieldRegistryMock registry;
+    Devices::DockingSystem docksys;
+
+    std::shared_ptr<WritableStateField<bool>> docking_config_cmd_fp;
+
+    std::unique_ptr<DockingController> docking_controller;
```
shihaocao: Why are control tasks unique pointers? I forgot

test/test_fsw_docking/test_fsw_docking.cpp
```diff
@@ -0,0 +1,79 @@
...
+    std::shared_ptr<ReadableStateField<bool>>is_turning_fp;
+
+    TestFixture() : registry() {
+        docking_config_cmd_fp = registry.create_writable_field<bool>("docksys.config_cmd");
+        docking_config_cmd_fp->set(true);
+
+        docking_controller = std::make_unique<DockingController>(registry, docksys); 
+
+        docked_fp = std::static_pointer_cast<ReadableStateField<bool>>(registry.find_readable_field("docksys.docked"));
+        docked_fp->set(true);
```
shihaocao: Why are we initializing docked?

shihaocao: Nvm i see you set to false in the actual init

test/test_fsw_docking/test_fsw_docking.cpp
```diff
@@ -0,0 +1,79 @@
...
+    TestFixture tf;
+    TEST_ASSERT_EQUAL(true, tf.docking_config_cmd_fp->get());
+    TEST_ASSERT_EQUAL(true, tf.docked_fp->get());
+    TEST_ASSERT_EQUAL(false, tf.dock_config_fp->get());
+}
+
+void test_task_execute() {
+    TestFixture tf;
+    tf.docked_fp->set(false);
+    TEST_ASSERT_EQUAL(false, tf.docked_fp->get());
```
tanishqaggarwal: This just tests the state field going false, which it will definitely do. Instead, the more useful test is to see if the docked_fp maintains its value after a call to execute(). I don't think it will with the current software since DockingSystem always returns docked() as true.

tanishqaggarwal: In order to set the docking system sometimes report docked() to false, you might want to create a desktop-only function inside the DockingSystem that sets docked() to either true or false. We can then call this desktop-only function within the test fixture.

PR #75, opened 2019-11-17
Dev/gomspace controller

fatimayousuf:


src/FCCode/GomspaceController.cpp
```diff
@@ -0,0 +1,274 @@
...
+    curout6_sr(0,10,10), curout6_f("gomspace.curout.output6", curout6_sr),
+
+    output1_sr(0,10,10), output1_f("gomspace.output.output1", output1_sr),
+    output2_sr(0,10,10), output2_f("gomspace.output.output2", output2_sr),
+    output3_sr(0,10,10), output3_f("gomspace.output.output3", output3_sr),
+    output4_sr(0,10,10), output4_f("gomspace.output.output4", output4_sr),
+    output5_sr(0,10,10), output5_f("gomspace.output.output5", output5_sr),
+    output6_sr(0,10,10), output6_f("gomspace.output.output6", output6_sr),
+    output7_sr(0,10,10), output7_f("gomspace.output.output7", output7_sr),
+    output8_sr(0,10,10), output8_f("gomspace.output.output8", output8_sr),
```
tanishqaggarwal: These outputs are boolean values (1 or 0) and so should the state fields. So we can reduce a lot of complexity here.

src/FCCode/GomspaceController.cpp
```diff
@@ -0,0 +1,274 @@
+#include "GomspaceController.hpp"
+
+GomspaceController::GomspaceController(StateFieldRegistry &registry, unsigned int offset,
+    Devices::Gomspace &_gs)
+    : TimedControlTask<void>(registry, offset), gs(_gs), 
+    vboost1_sr(0,10,10), vboost1_f("gomspace.vboost.output1", vboost1_sr),
+    vboost2_sr(0,10,10), vboost2_f("gomspace.vboost.output2", vboost2_sr),
+    vboost3_sr(0,10,10), vboost3_f("gomspace.vboost.output3", vboost3_sr),
```
tanishqaggarwal: vboost is from 0 to 3700 and we only care about the values in intervals of like 100 mV.

src/FCCode/GomspaceController.cpp
```diff
@@ -0,0 +1,274 @@
...
+
+    vbatt_sr(0,10,10), vbatt_f("gomspace.vbatt", vbatt_sr),
```
tanishqaggarwal: Number is in range 5000 to 9000, we care about the values in intervals of 50 mV

src/FCCode/GomspaceController.cpp
```diff
@@ -0,0 +1,274 @@
...
+    output_off_delta6_sr(0,10,10), output_off_delta6_f("gomspace.output_off_delta.output6", output_off_delta6_sr),
+    output_off_delta7_sr(0,10,10), output_off_delta7_f("gomspace.output_off_delta.output7", output_off_delta7_sr),
+    output_off_delta8_sr(0,10,10), output_off_delta8_f("gomspace.output_off_delta.output8", output_off_delta8_sr),
+
+    latchup1_sr(0,10,10), latchup1_f("gomspace.latchup.output1", latchup1_sr),
+    latchup2_sr(0,10,10), latchup2_f("gomspace.latchup.output2", latchup2_sr),
+    latchup3_sr(0,10,10), latchup3_f("gomspace.latchup.output3", latchup3_sr),
+    latchup4_sr(0,10,10), latchup4_f("gomspace.latchup.output4", latchup4_sr),
+    latchup5_sr(0,10,10), latchup5_f("gomspace.latchup.output5", latchup5_sr),
+    latchup6_sr(0,10,10), latchup6_f("gomspace.latchup.output6", latchup6_sr),
```
tanishqaggarwal: I don't think we care about these values, so you can just straight up delete all these serializers and fields. I'm so sorry for not mentioning this before, it slipped my mind.

src/FCCode/GomspaceController.cpp
tanishqaggarwal: Since these state fields all have the same values, they can share the same serializer (the serializer is copied into the state field upon construction.) No need to construct more than 1 serializer for fields with the same value.

src/FCCode/GomspaceController.cpp
```diff
@@ -0,0 +1,274 @@
...
+
+    wdt_i2c_time_left_sr(0,10,10), wdt_i2c_time_left_f("gomspace.wdt_i2c_time_left", wdt_i2c_time_left_sr),
```
tanishqaggarwal: This is a number from 0 to 1000 and we care about high precision (we need 10 bits).

src/FCCode/GomspaceController.cpp
```diff
@@ -0,0 +1,274 @@
...
+
+    wdt_gnd_time_left_sr(0,10,10), wdt_gnd_time_left_f("gomspace.wdt_gnd_time_left", wdt_gnd_time_left_sr),
+
+    wdt_csp_pings_left1_sr(0,10,10), wdt_csp_pings_left1_f("gomspace.wdt_csp_pings_left.output1", wdt_csp_pings_left1_sr),
+    wdt_csp_pings_left2_sr(0,10,10), wdt_csp_pings_left2_f("gomspace.wdt_csp_pings_left.output2", wdt_csp_pings_left2_sr),
```
tanishqaggarwal: We don't need these three values.

src/FCCode/GomspaceController.cpp
```diff
@@ -0,0 +1,274 @@
...
+
+    counter_wdt_i2c_sr(0,10,10), counter_wdt_i2c_f("gomspace.counter_wdt_i2c", counter_wdt_i2c_sr),
```
tanishqaggarwal: Number is from 0 to 2^32 - 1, we care about every bit of precision. But unsigned int has a default constructor that doesn't accept any arguments and captures the entire value of the unsigned int, so no need to specify any arguments.

src/FCCode/GomspaceController.cpp
```diff
@@ -0,0 +1,274 @@
...
+
+    counter_wdt_gnd_sr(0,10,10), counter_wdt_gnd_f("gomspace.counter_wdt_gnd", counter_wdt_gnd_sr),
+
+    counter_wdt_csp1_sr(0,10,10), counter_wdt_csp1_f("gomspace.counter_wdt_csp.output1", counter_wdt_csp1_sr),
+    counter_wdt_csp2_sr(0,10,10), counter_wdt_csp2_f("gomspace.counter_wdt_csp.output2", counter_wdt_csp2_sr),
```
tanishqaggarwal: We don't care about these values.

src/FCCode/GomspaceController.cpp
```diff
@@ -0,0 +1,274 @@
...
+
+    counter_boot_sr(0,10,10), counter_boot_f("gomspace.counter_boot", counter_boot_sr),
```
tanishqaggarwal: No need for serializer arguments; we care about the full value of the number, so we can use the default constructor.

src/FCCode/GomspaceController.cpp
```diff
@@ -0,0 +1,274 @@
...
+
+    temp1_sr(0,10,10), temp1_f("gomspace.temp.output1", temp1_sr),
+    temp2_sr(0,10,10), temp2_f("gomspace.temp.output2", temp2_sr),
+    temp3_sr(0,10,10), temp3_f("gomspace.temp.output3", temp3_sr),
+    temp4_sr(0,10,10), temp4_f("gomspace.temp.output4", temp4_sr),
+    temp5_sr(0,10,10), temp5_f("gomspace.temp.output5", temp5_sr),
+    temp6_sr(0,10,10), temp6_f("gomspace.temp.output6", temp6_sr),
```
tanishqaggarwal: Min: -40, Max: 125, Bits: can leave that argument blank and let the serializer constructor auto-calculate it.

src/FCCode/GomspaceController.cpp
```diff
@@ -0,0 +1,274 @@
...
+
+    bootcause_sr(0,10,10), bootcause_f("gomspace.bootcause", bootcause_sr),
+
+    battmode_sr(0,10,10), battmode_f("gomspace.battmode", battmode_sr),
+
+    pptmode_sr(0,10,10), pptmode_f("gomspace.pptmode", pptmode_sr),
+
+    reserved2_sr(0,10,10), reserved2_f("gomspace.reserved2", reserved2_sr)
```
tanishqaggarwal: Don't need this field

src/FCCode/GomspaceController.cpp
```diff
@@ -0,0 +1,274 @@
...
+    counter_boot_sr(0,10,10), counter_boot_f("gomspace.counter_boot", counter_boot_sr),
+
+    temp1_sr(0,10,10), temp1_f("gomspace.temp.output1", temp1_sr),
+    temp2_sr(0,10,10), temp2_f("gomspace.temp.output2", temp2_sr),
+    temp3_sr(0,10,10), temp3_f("gomspace.temp.output3", temp3_sr),
+    temp4_sr(0,10,10), temp4_f("gomspace.temp.output4", temp4_sr),
+    temp5_sr(0,10,10), temp5_f("gomspace.temp.output5", temp5_sr),
+    temp6_sr(0,10,10), temp6_f("gomspace.temp.output6", temp6_sr),
+
+    bootcause_sr(0,10,10), bootcause_f("gomspace.bootcause", bootcause_sr),
```
tanishqaggarwal: 0 to 8, no need for the third argument.

src/FCCode/GomspaceController.cpp
```diff
@@ -0,0 +1,274 @@
...
+
+    battmode_sr(0,10,10), battmode_f("gomspace.battmode", battmode_sr),
```
tanishqaggarwal: 0 to 4, no need for third argument.

src/FCCode/GomspaceController.cpp
```diff
@@ -0,0 +1,274 @@
...
+
+    pptmode_sr(0,10,10), pptmode_f("gomspace.pptmode", pptmode_sr),
```
tanishqaggarwal: 1 to 2, no need for third argument.

src/FCCode/GomspaceController.hpp
```diff
@@ -0,0 +1,227 @@
...
+    ReadableStateField<unsigned int> output5_f;
+
+    Serializer<unsigned int> output6_sr;
+    ReadableStateField<unsigned int> output6_f;
+
+    Serializer<unsigned int> output7_sr;
+    ReadableStateField<unsigned int> output7_f;
+
+    Serializer<unsigned int> output8_sr;
+    ReadableStateField<unsigned int> output8_f;
```
tanishqaggarwal: Fields can be bools.

tanishqaggarwal: We don't need the last two outputs because there are only 6 possible outputs on the Gomspace. The I2C specification just allows for 8, maybe because they share the specification across more robust battery systems.

src/FCCode/GomspaceController.hpp
```diff
@@ -0,0 +1,227 @@
...
+    ReadableStateField<unsigned int> temp3_f;
+
+    Serializer<unsigned int> temp4_sr;
+    ReadableStateField<unsigned int> temp4_f;
+
+    Serializer<unsigned int> temp5_sr;
+    ReadableStateField<unsigned int> temp5_f;
+
+    Serializer<unsigned int> temp6_sr;
+    ReadableStateField<unsigned int> temp6_f;
```
tanishqaggarwal: These need to be signed ints.

src/FCCode/GomspaceController.hpp
```diff
@@ -0,0 +1,227 @@
...
+
+    Serializer<unsigned int> bootcause_sr;
+    ReadableStateField<unsigned int> bootcause_f;
+
+    Serializer<unsigned int> battmode_sr;
+    ReadableStateField<unsigned int> battmode_f;
+
+    Serializer<unsigned int> pptmode_sr;
+    ReadableStateField<unsigned int> pptmode_f;
+
```
tanishqaggarwal: These can be unsigned chars. Using unsigned char saves a little memory on the flight computer.

src/FCCode/GomspaceController.cpp
```diff
@@ -0,0 +1,274 @@
...
+
+    curin1_sr(0,10,10), curin1_f("gomspace.curin.output1", curin1_sr),
+    curin2_sr(0,10,10), curin2_f("gomspace.curin.output2", curin2_sr),
+    curin3_sr(0,10,10), curin3_f("gomspace.curin.output3", curin3_sr),
+
+    cursun_sr(0,10,10), cursun_f("gomspace.cursun", cursun_sr),
+
+    cursys_sr(0,10,10), cursys_f("gomspace.cursys", cursys_sr),
+
+    reserved1_sr(0,10,10), reserved1_f("gomspace.reserved1", reserved1_sr),
```
tanishqaggarwal: No need for reserved1 or reserved2.

src/FCCode/GomspaceController.cpp
```diff
@@ -3,93 +3,66 @@
...
+    curin2_f("gomspace.curin.output2", curin_sr),
+    curin3_f("gomspace.curin.output3", curin_sr),
+
+    cursun_sr(0,10,10), 
+    cursun_f("gomspace.cursun", cursun_sr),
+
+    cursys_sr(0,10,10), 
+    cursys_f("gomspace.cursys", cursys_sr),
+
+    curout_sr(0,10,10),
```
tanishqaggarwal: We need to figure out values for these serializers experimentally. Can you create an issue ticket for this so that we don't forget about the work?

lib/Drivers/Gomspace.cpp
```diff
@@ -1,17 +1,25 @@
...
+#endif
 
 using namespace Devices;
 
+#ifdef DESKTOP
+#define I2C_INITIALIZATION I2CDevice("gomspace", 10000)
+#else
+#define I2C_INITIALIZATION I2CDevice("gomspace", Gomspace::wire, Gomspace::address, 10000)
+#endif
+
```
tanishqaggarwal: After the constructor can you add an `#undef I2C_INITIALIZATION`? That way we keep the global namespace clean

test/test_fsw_gomspace/test_fsw_gomspace.cpp
```diff
@@ -0,0 +1,271 @@
...
+    ReadableStateField<unsigned int>* counter_wdt_csp2_fp;
+
+    ReadableStateField<unsigned int>* counter_boot_fp;
+
+    ReadableStateField<unsigned int>* temp1_fp;
+    ReadableStateField<unsigned int>* temp2_fp;
+    ReadableStateField<unsigned int>* temp3_fp;
+    ReadableStateField<unsigned int>* temp4_fp;
+    ReadableStateField<unsigned int>* temp5_fp;
+    ReadableStateField<unsigned int>* temp6_fp;
```
tanishqaggarwal: I forgot that these temperature values in the housekeeping packet don't actually work. Can you remove them?

Also, all of the temperature values need to be SIGNED ints, even in the test.

test/test_fsw_gomspace/test_fsw_gomspace.cpp
```diff
@@ -0,0 +1,271 @@
...
+        battmode_fp = registry.find_readable_field_t<unsigned char>("gomspace.battmode");
+
+        pptmode_fp = registry.find_readable_field_t<unsigned char>("gomspace.pptmode");
+    }
+};
+
+void test_task_initialization() {
+    TestFixture tf;
+    //check that it is initialized correctly. tests pass
+    TEST_ASSERT_EQUAL(1, tf.gs.hk->vboost[0]);
```
tanishqaggarwal: Since you're testing exactly the same (long) list of values twice, I think it makes sense to break out the assert statements into a separate function, maybe call it something like `test_hk_packet_ok`, and then call that function in each of these two test cases that you have.

test/test_fsw_gomspace/test_fsw_gomspace.cpp
```diff
@@ -0,0 +1,271 @@
...
+        temp3_fp = registry.find_readable_field_t<unsigned int>("gomspace.temp.output3");
+        temp4_fp = registry.find_readable_field_t<unsigned int>("gomspace.temp.output4");
+        temp5_fp = registry.find_readable_field_t<unsigned int>("gomspace.temp.output5");
+        temp6_fp = registry.find_readable_field_t<unsigned int>("gomspace.temp.output6");
+
+        bootcause_fp = registry.find_readable_field_t<unsigned char>("gomspace.bootcause");
+
+        battmode_fp = registry.find_readable_field_t<unsigned char>("gomspace.battmode");
+
+        pptmode_fp = registry.find_readable_field_t<unsigned char>("gomspace.pptmode");
```
tanishqaggarwal: Can you add assert(field) statements after here to ensure that the field named `field` is actually created by the controller?

platformio.ini
```diff
@@ -72,6 +72,7 @@ extends = native_common
 build_flags = ${native_common.build_flags} -O0 -g --coverage -D CI
 extra_scripts = get_code_coverage.py
 src_filter = +<FCCode/> -<FCCode/main.cpp> +<native_dummy.cpp>
+test_filter = test_fsw_gomspace
```
athena255: Should this be test_fsw_*?

lib/Drivers/Gomspace.cpp
```diff
@@ -52,17 +66,23 @@ bool Gomspace::get_hk() {
     // Serial.println();
     i2c_finish();
 
+    #ifndef DESKTOP
     if (buffer[0] != PORT_BYTE && buffer[1] != 0)
```
athena255: Why is this && instead of ||?

lib/Devices/I2CDevice.hpp
```diff
@@ -41,6 +41,7 @@
 typedef unsigned int i2c_stop;
 typedef unsigned int i2c_t3;
 #define I2C_STOP 0
+#define I2C_NOSTOP 0
```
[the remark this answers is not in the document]
tanishqaggarwal: Yeah, this is OK since this is  purely a desktop hack to get the driver to compile.

lib/Drivers/Gomspace.cpp
```diff
@@ -120,18 +157,26 @@ bool Gomspace::get_hk_wdt() {
     i2c_end_transmission(I2C_NOSTOP);
 
     size_t struct_size = sizeof(eps_hk_wdt_t);
+    #ifndef DESKTOP
     unsigned char buffer[struct_size + 2];
```
athena255: might want to zero this out in case the entire buffer is not filled

i2c_read returns the number of bytes successfully read. It does not guarantee that it will read all the requested bytes. It might be good idea to check the return value of i2c_read in case of read failure.

PR #76, opened 2019-11-17
Dev/piksi controller v2

shihaocao:
This is a PR that encapsulates the Piksi Driver and its Corresponding Control Task.

Within the PiksiController, read_all() is called once, and no memory is stored from cycle to cycle, except an error counter that is reset upon a good reading.

Essentially, barring an extremely long chain of failures, the state of PiksiController is fully determined by the current state of the Piksi. Hopefully, this keeps things simple.

I will make a flow chart of read_all() and the PiksiController execute() soon.


tanishqaggarwal:
There are too many file changes unrelated to your actual change. Closing the PR until you're able to fix this.

PR #78, opened 2019-11-17
Dev/piksi controller v2

shihaocao:
This is a PR that encapsulates the Piksi Driver and its Corresponding Control Task.

Within the PiksiController, read_all() is called once, and no memory is stored from cycle to cycle, except an error counter that is reset upon a good reading.

Essentially, barring an extremely long chain of failures, the state of PiksiController is fully determined by the current state of the Piksi. Hopefully, this keeps things simple.

I will make a flow chart of read_all() and the PiksiController execute() soon.

### DIFF
--- a/lib/Drivers/Piksi.cpp
+++ b/lib/Drivers/Piksi.cpp
@@ -6,8 +6,6 @@
 #endif
 
 using namespace Devices;
-//using namespace std;
-
 
 // Initialize callback nodes so that C++ doesn't complain about them not being
 // defined.
@@ -30,8 +28,6 @@ Piksi::Piksi(const std::string &name, HardwareSerial &serial_port)
     : Device(name), _serial_port(serial_port) {}
 #else
 Piksi::Piksi(const std::string &name) {}
-    //: Device(name) {}
-//Piksi::Piksi();
 #endif
 
 
@@ -281,15 +277,23 @@ unsigned char Piksi::read_all() {
     
     if(bytes_available()){
         bool crc_error = false;
-        while(bytes_available() && (micros() - initial_time < 900)){
+        while(bytes_available() && (micros() - initial_time < READ_ALL_LIMIT)){
             int p_b_return = process_buffer();
             if(p_b_return < 0)
                 crc_error = true;
         }
-        if(micros()-initial_time >= 900){
+
+        //ensure that if the while loop terminated because of exceeding the READ_ALL_LIMIT
+        //it will enter the clear bytes condition below
+        delayMicroseconds(5);
+
+        if(micros()-initial_time >= READ_ALL_LIMIT){
             clear_bytes();
             return 5;
         }
+        
+        //by this point in the code, it is guarenteed that there are no more bytes in buffer
+
         if(crc_error)
             return 3;
         else if(_gps_time_update && _pos_ecef_update && _vel_ecef_update && !_baseline_ecef_update)
@@ -299,7 +303,7 @@ unsigned char Piksi::read_all() {
             //Something RTK
             return 1;
         else
-            //no relevant callbacks or crc_errors -> NO_FIX
+            //no relevant callbacks -> NO_FIX
             return 2;
     }
     else
--- a/lib/Drivers/Piksi.cpp
+++ b/lib/Drivers/Piksi.cpp
@@ -278,8 +278,8 @@ unsigned char Piksi::read_all() {
     if(bytes_available()){
         bool crc_error = false;
         while(bytes_available() && (micros() - initial_time < READ_ALL_LIMIT)){
-            int p_b_return = process_buffer();
-            if(p_b_return < 0)
+            //call process_buffer() to process data, and check if crc_error happened
+            if(process_buffer() < 0)
                 crc_error = true;
         }
 
--- a/lib/Drivers/Piksi.cpp
+++ b/lib/Drivers/Piksi.cpp
@@ -290,16 +290,16 @@ unsigned char Piksi::read_all() {
             clear_bytes();
             return 5;
         }
-  
-        if(_gps_time_update && _pos_ecef_update && _vel_ecef_update && !_baseline_ecef_update)
+        if(crc_error)
+            return 3;
+        else if(_gps_time_update && _pos_ecef_update && _vel_ecef_update && !_baseline_ecef_update)
+            //SPP
             return 0;
         else if(_gps_time_update && _pos_ecef_update && _vel_ecef_update && _baseline_ecef_update)
+            //Something RTK
             return 1;
-        else if(crc_error)
-            return 3;
         else
-            //error condition, perhaps serial data bad
-            //or time looped around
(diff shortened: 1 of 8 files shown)


lib/Drivers/Piksi.cpp
```diff
@@ -1,344 +1,435 @@
...
+u32 Piksi::_uart_read(u8 *buff, u32 n, void *context) {
+    #ifndef DESKTOP
+    Piksi *piksi = (Piksi *)context;
+    
+    HardwareSerial &sp = piksi->_serial_port;
+
+    u32 i;
+    for (i = 0; i < n; i++) {
+        if (sp.available())
+            buff[i] = sp.read();
```
athena255: Is n == the size of buff? If not, then this function could possibly overflow buff.

lib/Drivers/Piksi.cpp
```diff
@@ -1,344 +1,435 @@
...
+        _latest_log = &_logbook[0];
+    }
+    memcpy(_latest_log, msg, sizeof(msg_log_t));
+    if (_logbook_size < _logbook_max_size) _logbook_size++;
+}
+void Piksi::dump_log(char *destination) {
+    memcpy(destination, (char *)_logbook, _logbook_size * sizeof(msg_log_t));
+}
+void Piksi::clear_log() {
+    _latest_log = &_logbook[0] - 1;
```
athena255: These logbook methods are difficult to understand, which makes them prone to error. Are there tests for these methods? A grep of the FlightSoftware folder indicates there are no tests for these.

lib/Drivers/Piksi.cpp
```diff
@@ -1,344 +1,435 @@
...
+    Piksi *piksi = (Piksi *)context;
+    memcpy((u8 *)(&(piksi->_base_pos_ecef)), msg, sizeof(msg_base_pos_ecef_t));
+}
+void Piksi::_iar_callback(u16 sender_id, u8 len, u8 msg[], void *context) {
+    Piksi *piksi = (Piksi *)context;
+    memcpy((u8 *)(&(piksi->_iar)), msg, sizeof(msg_iar_state_t));
+}
+void Piksi::_settings_read_resp_callback(u16 sender_id, u8 len, u8 msg[], void *context) {
+    Piksi *piksi = (Piksi *)context;
+    memcpy((u8 *)(&(piksi->_base_pos_ecef)), msg, sizeof(msg_settings_read_resp_t));
```
athena255: Why do we copy msg_settings_read_resp_t length into _base_poc_ecef since msg_settings_read_resp_t is a different type than _base_pos_ecef. msg_settings_read_resp_t also seems to be 0 bytes.

src/FCCode/PiksiControlTask.cpp
```diff
@@ -0,0 +1,134 @@
...
+    //3 means CRC error on serial
+    //5 means timing error exceed
+    if(read_out == 3|| read_out == 4|| read_out == 5)
+        since_good_cycles += 1;
+    else 
+        since_good_cycles = 0;
+        
+    //if we haven't had a good reading in ~120 seconds the piksi is probably dead
+    //eventually replace with HAVT logic
+    if(since_good_cycles > 1000){
```
athena255: Might be good  to define this somewhere

test/test_fsw_piksi_control_task/test_piksi_control_task.cpp
```diff
@@ -0,0 +1,298 @@
...
+        TestFixture tf;
+        assert_piksi_mode(piksi_mode_t::NO_FIX);
+}
+void test_read_errors(){
+        TestFixture tf;
+
+        //read out == 3 means a CRC error occurred
+        tf.set_read_return(3);
+        tf.execute();
+        assert_piksi_mode(piksi_mode_t::CRC_ERROR);
```
athena255: Might be good to assert the precondition before execute as well.

lib/Drivers/Piksi.cpp
```diff
@@ -1,344 +1,435 @@
...
+void Piksi::piksi_reset() {
+    sbp_send_message(&_sbp_state, SBP_MSG_RESET, SBP_SENDER_ID, 0, nullptr, &Piksi::_uart_write);
+}
+void Piksi::send_user_data(const msg_user_data_t &data) {
+    sbp_send_message(&_sbp_state, SBP_MSG_USER_DATA, SBP_SENDER_ID, sizeof(msg_user_data_t),
+                     (unsigned char *)&data, &Piksi::_uart_write);
+}
+
+signed char Piksi::process_buffer() {
+    signed char status = ((signed char)sbp_process(&_sbp_state, Piksi::_uart_read));
```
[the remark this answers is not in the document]
tanishqaggarwal: In C++ there's no difference; it automatically resolves a pointer to a function. For consistency  I think it'd be better to not use the ampersand

lib/Drivers/Piksi.cpp
tanishqaggarwal: Yep, it's the size of the buf and the input arguments are supplied by libsbp, so they're supposed to be correct

lib/Drivers/Piksi.cpp
tanishqaggarwal: This is a typo, good catch

test/test_fsw_piksi_control_task/test_piksi_control_task.cpp
```diff
@@ -0,0 +1,298 @@
...
+        ReadableStateField<d_vector_t>* pos_fp;
+        ReadableStateField<d_vector_t>* vel_fp;
+        ReadableStateField<d_vector_t>* baseline_fp;
+        ReadableStateField<gps_time_t>* time_fp;
+
+        std::unique_ptr<PiksiControlTask> piksi_task;
+
+        Piksi piksi;
+        // Create a TestFixture instance of PiksiController with pointers to statefields
+        #ifndef DESKTOP
```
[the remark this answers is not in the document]
shihaocao: I need to conditionally compile because of the Serial port

lib/Drivers/Piksi.cpp
```diff
@@ -1,344 +1,435 @@
...
+    _read_return = out;
+}
+#endif
+
+unsigned int Piksi::get_iar() { return _iar.num_hyps; }
+
+char *Piksi::get_settings_read_resp() { return _settings_read_resp.setting; }
+
+unsigned int Piksi::get_heartbeat() { return _heartbeat.flags; }
+bool Piksi::is_system_healthy() { return !(_heartbeat.flags & 0x0001); }
```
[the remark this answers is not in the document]
shihaocao: heartbeat will be addressed fully in the future

test/test_fsw_piksi_control_task/test_piksi_control_task.cpp
shihaocao: asserted in test_initialization :D

shihaocao:
Finished the Piksi Driver and Control Task :)

PR #79, opened 2019-11-19
Remove libsbp

tanishqaggarwal:
We don't need it anymore


tanishqaggarwal:
Apparently there are dependencies, nevermind.

PR #84, opened 2019-11-25
Fix the debug task to wait for the sim

tanishqaggarwal:
In order for the sim to work properly, the debug task should wait for the simulation's signal that it's done updating values before allowing the Flight Software to continue executing its cycle. This PR implements that feature.

### DIFF
--- a/lib/Drivers/Piksi.hpp
+++ b/lib/Drivers/Piksi.hpp
@@ -404,7 +404,6 @@ class Piksi {
     bool _pos_ecef_update;
     bool _baseline_ecef_update;
     bool _vel_ecef_update;
-    bool _iar_update;
     bool _heartbeat_update;
     bool _user_data_update;
 
--- a/test/test_real_fsw.py
+++ b/test/test_real_fsw.py
@@ -10,7 +10,7 @@ import serial
 import traceback
 import unittest
 
-class TestDummyFlightSoftwareBinary(unittest.TestCase):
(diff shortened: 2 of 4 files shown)


lib/Drivers/Piksi.hpp
```diff
@@ -404,7 +404,6 @@ class Piksi {
     bool _pos_ecef_update;
     bool _baseline_ecef_update;
     bool _vel_ecef_update;
-    bool _iar_update;
```
tanishqaggarwal: Had to do this to resolve compile errors

shihaocao: makes sense, never ended up implementing

test/test_real_fsw.py
```diff
@@ -20,25 +20,38 @@ class TestDummyFlightSoftwareBinary(unittest.TestCase):
 
     def setUp(self):
         master_fd, slave_fd = pty.openpty()
-        self.dummy_fsw = subprocess.Popen([self.binary_dir], stdin=master_fd, stdout=master_fd)
+        self.fsw = subprocess.Popen([self.binary_dir], stdin=master_fd, stdout=master_fd)
         self.console = serial.Serial(os.ttyname(slave_fd), 9600, timeout=1)
 
-    def read_cycle_no(self):
+    def readCycleNumber(self):
```
shihaocao: Note that this is outdated in comparison to the quake integration branch.
Talked with Tanishq, and merge conflict will be addressed.

PR #88, opened 2019-12-02
Downlink Producer

tanishqaggarwal:
This implements v1 of the downlink producer, along with some drive-by fixes

As a reminder, here's the general structure of each downlink frame:
- Control cycle count
- For each flow definition, the flow ID and then the bits in each flow field
- Packet headers (1 bit) at the beginning of each 70-byte packet. The header bit is a 1 if it's the first packet in the downlink frame, and 0 if its not.


platformio.ini
```diff
@@ -11,8 +11,8 @@ lib_deps =
   https://github.com/pathfinder-for-autonomous-navigation/psim
   https://github.com/pathfinder-for-autonomous-navigation/CommonSoftware
   https://github.com/nox771/i2c_t3
-build_flags = -std=c++14 -Werror -Wall -D UNITY_INCLUDE_DOUBLE -O3 -D NDEBUG -DLIN_RANDOM_SEED=358264
-src_filter = +<FCCode/> +<teensy.cpp>
+build_flags = -std=c++14 -Wno-narrowing -Werror -Wall -D UNITY_INCLUDE_DOUBLE -O3 -D NDEBUG -DLIN_RANDOM_SEED=358264
```
tanishqaggarwal: This compiler flag is required in order to allow the unit test for Downlink Producer to pass. Otherwise, it complains that the initiailzation of arrays of chars can't be done by setting the individual bytes.

src/FCCode/MainControlLoop.cpp
```diff
@@ -44,6 +46,13 @@ void MainControlLoop::execute() {
 
     piksi_control_task.execute_on_time(control_cycle_start);
     mission_manager.execute_on_time(control_cycle_start);
+    downlink_producer.execute_on_time(control_cycle_start);
     quake_manager.execute_on_time(control_cycle_start);
     docking_controller.execute_on_time(control_cycle_start);
 }
+
+#ifdef GSW
```
tanishqaggarwal: Downlink parser software on the ground uses this function to get access to the downlink producer. Ground software instantiates flight software internally in order to populate the state field registry, then uses the DownlinkProducer to access all of the flows machinery.

src/FCCode/MainControlLoop.cpp
```diff
@@ -44,6 +46,13 @@ void MainControlLoop::execute() {
...
+DownlinkProducer* MainControlLoop::get_downlink_producer() {
+    return &downlink_producer;
+}
+#endif
```
tanishqaggarwal: Downlink parser software instantiates flight software in order to populate the state field registry, then uses this function to get access to the flows machinery of DownlinkProducer.

src/FCCode/DownlinkProducer.hpp
```diff
@@ -0,0 +1,130 @@
...
+    /**
+     * @brief Toggle a flow on or off.
+     */
+    void toggle_flow(unsigned char id);
+
+    /**
+     * @brief Swap the priorities of the flows with the given IDs. Also swaps
+     * their active status.
+     */
+    void swap_flow_priorities(unsigned char id1, unsigned char id2);
```
tanishqaggarwal: These functions still need to get tested. I'll do this when I implement a facility for actually running these functions based on uplink commands.

test/test_fsw_quake_control_task/test_quake_controller.cpp
```diff
@@ -19,7 +19,7 @@ void exec_check(QuakeControlTask& task, int expectedState, int expectedFnNum)
 void test_task_initialization()
```
tanishqaggarwal: The folder names of the quake manager and quake control task tests weren't right for the unit test regex, so the tests weren't running and so I had to modify the folder names. Once I modified them I found that there were compilation bugs inside the unit test, which I fixed here

test/test_fsw_downlink_producer/packet_gen.py
```diff
@@ -0,0 +1,73 @@
+# Scratch space for generating manual packets for downlink testing
```
tanishqaggarwal: This file produces the expected outputs in the test_fsw_downlink_producer test file

src/FCCode/DownlinkProducer.cpp
```diff
@@ -0,0 +1,247 @@
...
+        field_list.push_back(field_ptr);
+    }
+
+    assert(get_packet_size() <= 560 - 1 - 32); // Flow should fit within one downlink packet
+}
+
+size_t DownlinkProducer::Flow::get_packet_size() const {
+    // Get bitcount of all fields in the flow
+    size_t packet_size = 0;
+    packet_size += id_sr.bitsize();
```
athena255: Is id_sr the index of the field? If so, why is it only added once to the packet, shouldn't each field have an id_sr?

tanishqaggarwal: id_sr is the serializer that contains the _flow's_ ID. We don't have a need for field IDs.

tanishqaggarwal:
I'm going to break this PR down into two to make it easier to review.

tanishqaggarwal:
Nevermind, turns out I'd made this the simplest it could be. Sorry :/

PR #91, opened 2019-12-02
Attitude Estimator Control Task

shihaocao:
This PR creates the AttitudeEstimator Control Task. It finds necessary input state fields, calls Kyle's gnc estimate attitude function and dumps the output into output state fields.


src/FCCode/AttitudeEstimator.cpp
```diff
@@ -0,0 +1,77 @@
+#include "AttitudeEstimator.hpp"
+
+AttitudeEstimator::AttitudeEstimator(StateFieldRegistry &registry,
+    unsigned int offset) 
+    : TimedControlTask<void>(registry, offset),
+    data(),
+    state(),
+    estimate(),
+    q_body_eci_sr(0, 1, 22),
```
tanishqaggarwal: No need for arguments on a quaternion serializer constructor

src/FCCode/AttitudeEstimator.cpp
```diff
@@ -0,0 +1,77 @@
...
+
+AttitudeEstimator::AttitudeEstimator(StateFieldRegistry &registry,
+    unsigned int offset) 
+    : TimedControlTask<void>(registry, offset),
+    data(),
+    state(),
+    estimate(),
+    q_body_eci_sr(0, 1, 22),
+    q_body_eci_f("attitude_estimator.q_body_eci", q_body_eci_sr),
+    w_body_sr(0, 1000, 22),
```
tanishqaggarwal: @kkrol27 are these the actual bounds on body angular rate?

test/test_fsw_attitude_estimator/test_attitude_estimator.cpp
```diff
@@ -0,0 +1,63 @@
...
+        
+        }
+};
+
+void test_task_initialization()
+{
+        TestFixture tf;
+}
+
+void test_execute(){
```
tanishqaggarwal: Don't need this function and the test_execute() below

src/FCCode/AttitudeEstimator.hpp
```diff
@@ -0,0 +1,60 @@
+#ifndef ATTITUDE_ESTIMATOR_HPP_
+#define ATTITUDE_ESTIMATOR_HPP_
+
+//#include "../../.pio/libdeps/teensy35_hootl/psim/include/gnc_attitude_estimation.hpp"
```
tanishqaggarwal: Remove commented line

test/test_fsw_attitude_estimator/test_attitude_estimator.cpp
```diff
@@ -0,0 +1,63 @@
+#include "../StateFieldRegistryMock.hpp"
```
tanishqaggarwal: Note to other reviewers: there's no need for unit testing for this class, because it's just a wrapper around another algorithm. Instead, the class will be functional-tested with psim

src/FCCode/AttitudeEstimator.cpp
```diff
@@ -0,0 +1,77 @@
...
+    }
+
+void AttitudeEstimator::execute(){
+    set_data();
+    gnc::estimate_attitude(state, data, estimate);
+    set_estimate();
+}
+
+void AttitudeEstimator::set_data(){
+    data.t = ((double)(unsigned long)(piksi_time_fp->get() - pan_epoch_fp->get()))/(10e9);
```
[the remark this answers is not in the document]
shihaocao: Oof yes, will fix

src/FCCode/AttitudeEstimator.cpp
kkrol27: Uhh.... I don't think we really have any technical bounds - the estimated omega can be arbitrarily large because its a finite difference. I know we discussed at some point coming up with an artificial max omega that's worth reporting however. This would be an @nhz2 @stewartaslan question.
Sort of on that note, are the float arguments clamped prior to being serialized @tanishqaggarwal?

test/test_fsw_adcs/test_fsw_adcs.cpp
```diff
@@ -0,0 +1,30 @@
+#include <unity.h>
```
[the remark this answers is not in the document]
shihaocao: Boilerplate, it was actually already there before I started work on this branch XD

src/FCCode/AttitudeEstimator.cpp
shihaocao: Got it, I'll remove them

test/test_fsw_adcs/test_fsw_adcs.cpp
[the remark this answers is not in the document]
tanishqaggarwal: Feel free to remove it

src/FCCode/AttitudeEstimator.cpp
shihaocao: Absolutely not oops, I'll dig into this

src/FCCode/AttitudeEstimator.cpp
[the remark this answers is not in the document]
kkrol27: Also, the following flags must be included for teensy builds to ensure that `1.0e9` is treated as a `double` and not a `float`:

    build_unflags = -fsingle-precision-constant -fmath-errno
    build_flags = -fno-math-errno

Not sure if that has trickled into FSW yet.

test/test_fsw_attitude_estimator/test_attitude_estimator.cpp
shihaocao: I thought we'd leave this as boilerplate until this function could actually do stuff based on mocked inputs and outputs?

tanishqaggarwal: I think this class is sufficiently simple that we can forgo unit testing and use functional testing with psim instead

src/FCCode/AttitudeEstimator.hpp
```diff
@@ -0,0 +1,60 @@
...
+    ReadableStateField<gps_time_t>* pan_epoch_fp;
+
+    //! Time from Piksi
+    ReadableStateField<gps_time_t>* piksi_time_fp;
+    //! Position of this satellite, Vector of doubles
+    ReadableStateField<d_vector_t>* pos_vec_ecef_fp;
+    //! Sun vector of this satellite, in the body frame.
+    ReadableStateField<f_vector_t>* ssa_vec_rd_fp;
+    //! Magnetic field vector of this satellite in the body frame.
+    ReadableStateField<f_vector_t>* mag_vec_fp;
```
kkrol27: Would we want to make these either `ReadableStateField<...> *const` or `ReadableStateField<...> &`as they don't change during the lifetime of the object? They'd just get initialized in the constructor's initializer list as well. I'd say that'd be a more "proper" way of treating these member variables.

@tanishqaggarwal because this may be a more general design question as well.

test/test_fsw_adcs/test_fsw_adcs.cpp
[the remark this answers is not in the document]
kkrol27: Wait isn't this a test file...? or I'm just totally misunderstanding something.

src/FCCode/AttitudeEstimator.hpp
tanishqaggarwal: I like that idea.

tanishqaggarwal: I like that idea.

kkrol27: I'd recommend references if possible because it'd prevent something like `*(x + 1)` as well as `*(x++)` while `*const`'s only prevents the latter - even though the former would sorta be ridiculous to miss lol

src/FCCode/AttitudeEstimator.hpp
```diff
@@ -0,0 +1,60 @@
...
+
+    /**
+     * @brief Using raw sensor inputs, determine the attitude and angular state
+     * of the spacecraft.
+     */
+    void execute() override;
+
+    void set_data();
+
+    void set_estimate();
```
kkrol27: Also, is there any point to making these functions `public` as opposed to `protected`/`private`?

tanishqaggarwal: For unit testing this would be nice. But since we're not unit testing this control task I agree with making the functions protected

src/FCCode/AttitudeEstimator.cpp
```diff
@@ -0,0 +1,77 @@
...
+        //Find pan epoch
+        pan_epoch_fp = find_readable_field<gps_time_t>("pan.epoch", __FILE__, __LINE__);
+
+        //Find piksi inputs
+        piksi_time_fp = find_readable_field<gps_time_t>("piksi.time", __FILE__, __LINE__);
+        pos_vec_ecef_fp = find_readable_field<d_vector_t>("piksi.pos", __FILE__, __LINE__);
+
+        //find ADCSBoxMonitor inputs
+        ssa_vec_rd_fp = find_readable_field<f_vector_t>("adcs_box.sun_vec", __FILE__, __LINE__);
+        mag_vec_fp = find_readable_field<f_vector_t>("adcs_box.mag_vec", __FILE__, __LINE__);
```
tanishqaggarwal: You should assert false if these fields aren't found. Since you're still working on them, I suggest creating the fields inside the `FieldCreatorTask` as a temporary measure

kkrol27: Could you maybe add a function called `assert_readable_filed<...>(...)` that wraps the assertion up in the function call? Not really sure if that'd do what you want in all scenarios though.

src/FCCode/AttitudeEstimator.cpp
nhz2: I think the largest omega we can measure from finite differences is if the sat is rotating 180 degrees every control cycle. If it rotates faster, there are aliasing effects. So max omega with a x2 safety factor is about 2*pi/0.12 rad/s.

src/FCCode/AttitudeEstimator.cpp
tanishqaggarwal: The problem with doing that is that it hides where the assertion actually occurred--it'll report some file and line inside ControlTask.hpp, not in AttitudeEstimator.cpp.

src/FCCode/AttitudeEstimator.cpp
tanishqaggarwal: Yeah seems like 55 rad/s would be a very safe upper bound

shihaocao: I'll go from -55 to 55 with full precision on the serializer

src/FCCode/AttitudeEstimator.cpp
```diff
@@ -35,7 +35,7 @@ void AttitudeEstimator::execute(){
 }
 
 void AttitudeEstimator::set_data(){
-    data.t = ((double)(unsigned long)(piksi_time_fp->get() - pan_epoch_fp->get()))/(10e9);
+    data.t = ((double)(unsigned long)(piksi_time_fp->get() - pan_epoch_fp->get()))/(1e9);
```
nhz2: To get 1e9 to be a `double` literal, you need to append an L like `1e9L` This is because, by default on teensy the compiler treats `double` literals as `float`, adding the L makes a `long double`, which is the most precise floating point type.

src/FCCode/AttitudeEstimator.hpp
```diff
@@ -0,0 +1,67 @@
...
+     */
+    AttitudeEstimator(StateFieldRegistry& registry, unsigned int offset);
+
+   protected:
+
+    /**
+     * @brief Using raw sensor inputs, determine the attitude and angular state
+     * of the spacecraft.
+     */
+    void execute() override;
```
nhz2: I think `execute` should be public.

src/FCCode/AttitudeEstimator.hpp
```diff
@@ -0,0 +1,67 @@
...
+     */
+    void set_data();
+
+    /**
+     * @brief Set the estimate object from the outputs of the gnc estimate
+     * 
+     */
+    void set_estimate();
+    
+    const gps_time_t pan_epoch;
```
nhz2: This should be initialized to the value from psim.

src/FCCode/AttitudeEstimator.cpp
```diff
@@ -0,0 +1,78 @@
+#include "AttitudeEstimator.hpp"
+
+AttitudeEstimator::AttitudeEstimator(StateFieldRegistry &registry,
+    unsigned int offset) 
+    : TimedControlTask<void>(registry, offset),
+    pan_epoch(),
```
nhz2: You don't need to default initialize `pan_epoch` here.

src/FCCode/AttitudeEstimator.hpp
shihaocao: yep fixed on next commit

src/FCCode/AttitudeEstimator.hpp
```diff
@@ -0,0 +1,67 @@
...
+     * @brief Inputs collected from Piksi and ADCSBoxMonitor.
+     */
+    //! Time from Piksi
+    const ReadableStateField<gps_time_t>* piksi_time_fp;
+    //! Position of this satellite, Vector of doubles
+    const ReadableStateField<d_vector_t>* pos_vec_ecef_fp;
+    //! Sun vector of this satellite, in the body frame.
+    const ReadableStateField<f_vector_t>* ssa_vec_rd_fp;
+    //! Magnetic field vector of this satellite in the body frame.
+    const ReadableStateField<f_vector_t>* mag_vec_fp;
```
kkrol27: I think we want this to be a constant pointer to a readable state field and not a pointer to a constant readable state field - i.e. `ReadableStateField<f_vector_t> *const` and not `ReadableStateField<f_vector_t> const *`?

This is why a always put `const` to the right of the variable name for consistency lol

tanishqaggarwal: I disagree. We don't want to be able to change the value of the state field. We could perhaps make the pointer const as well, but then the initialization must move to the initializer list of the constructor.

src/FCCode/AttitudeEstimator.cpp
kkrol27: We'll be using the compiler flag to fix that though... Or should be for the sake of libraries.

src/FCCode/AttitudeEstimator.hpp
```diff
@@ -0,0 +1,67 @@
...
+     */
+    void set_data();
+
+    /**
+     * @brief Set the estimate object from the outputs of the gnc estimate
+     * 
+     */
+    void set_estimate();
+    
+    const gps_time_t pan_epoch;
```
kkrol27: Also, this is going to need to be used in multiple control tasks. Would there be a better place to hold this constant potentially?

src/FCCode/AttitudeEstimator.hpp
kkrol27: Ohhh I see - I was talking originally about turning this in to a `*const` or reference to be more correct. I guess it should be plane `const` as well.

src/FCCode/AttitudeEstimator.hpp
tanishqaggarwal: I feel like it's OK for each control task to just build the field again. It's not that much overhead

src/FCCode/AttitudeEstimator.hpp
kkrol27: If it was me, for the sake of being thorough I'd mark it as a constant pointer and constant value but I'll approve either way. Sorry we misunderstood each other before.

tanishqaggarwal: I think I'll drive-by fix this when I work on MissionManager

Nah, `Readable` just means that the field isn't writable via uplink. It's still writable by other control tasks.

tanishqaggarwal: I wish there was a cleaner way to name Readable state fields but something like `GroundReadableStateField` is too many letters, I feel?

kkrol27: Ah okay got it. And I'll approve then - if @shihaocao wants to fix it now and @nhz2 is still around to review super quick we can do that (I'll still be up too). Otherwise fixing it later is fine too.

src/FCCode/AttitudeEstimator.hpp
shihaocao: Currently, I can't think of a unified place to put it. Perhaps on a future pr, I can put pan_epoch in gnc_constants itself?

PR #94, opened 2019-12-04
Bitstream

athena255:


test/test_fsw_bitstream/test_bitstream.cpp
```diff
@@ -0,0 +1,363 @@
...
+  char* myarr = (char*)"\xde\xad\xbe\xef\xab\xcd\xef";
+  uint8_t res[64];
+  uint32_t actual [7] = {0xde, 0xad, 0xbe, 0xef, 0xab, 0xcd, 0xef};
+  bitstream bs(myarr, 7);
+  bs.nextN(64, res);
+  for (int i = 0; i < 7; ++i)
+  {
+    TEST_ASSERT_EQUAL(actual[i], (uint32_t)res[i]);
+  }
+}
```
tanishqaggarwal: Are you counting on null-termination of myarr here? There are only 7 bytes in your character array but you're writing 8 bytes to res

test/test_fsw_bitstream/test_bitstream.cpp
```diff
@@ -0,0 +1,363 @@
...
+void test6()
+{
+  char* myarr = (char*)"\xde\xad\xbe\xef\xab\xcd\xef";
+  uint8_t res[64];
+  uint32_t actual [7] = {0xde, 0xad, 0xbe, 0xef, 0xab, 0xcd, 0xef};
+  bitstream bs(myarr, 7);
+  bs.nextN(64, res);
+  for (int i = 0; i < 7; ++i)
+  {
+    TEST_ASSERT_EQUAL(actual[i], (uint32_t)res[i]);
```
tanishqaggarwal: I'm also confused by how casting res[i] to uint32_t results in the exact same results as the original char array. I think my confusion is hard to explain so I'll ask you about it on Friday?

test/test_fsw_bitstream/test_bitstream.cpp
```diff
@@ -0,0 +1,363 @@
...
+    bit_arr[i] = (0xdeadbeef >> i) & (0x1);
+  }
+
+  char arr [ (bit_arr.size() + 7)/8 ];
+  bitstream bs(bit_arr, arr); 
+  uint8_t res[4];
+  bs.nextN(18, res);
+
+  uint8_t expect[4] = {0xef, 0xbe, 0x1};
+  for (int i = 0; i < 3; ++i)
```
tanishqaggarwal: i = 0..2, so you check for 24 bits being equal, but you only write 18 bits to res. How come the expected and actual end up matching up, then?

PR #101, opened 2019-12-10
Added uplink consumer files

athena255:


src/FCCode/UplinkConsumer.cpp
```diff
@@ -0,0 +1,110 @@
+#include "UplinkConsumer.h"
+#include<bitstream.h>
+
+UplinkConsumer::UplinkConsumer(StateFieldRegistry& registry, unsigned int offset) :
+    TimedControlTask<void>(registry, offset),
+    radio_mt_packet_len_f("uplink_consumer.mt_len"),
+    radio_mt_packet_f("uplink_consumer.mt_ptr"),
```
tanishqaggarwal: Best to make these names as short as possible in order to save memory. Maybe something like "uplink.len" and "uplink.ptr"?

tanishqaggarwal: Relatedly if you could drive-by fix the downlink producer names, that would be awesome.

src/FCCode/UplinkConsumer.cpp
```diff
@@ -0,0 +1,110 @@
...
+    size_t field_index = 0, field_len = 0, bits_consumed = 0;
+    while (bits_consumed < radio_mt_packet_len_f.get())
+    {
+        // Get index from the bitstream
+        bits_consumed += bs.nextN(index_size, reinterpret_cast<uint8_t*>(&field_index));
+
+        // Get field length from the index
+        field_len = get_field_length(field_index);
+
+        auto field_p = registry.writable_fields[field_index];
```
tanishqaggarwal: Shouldn't this be `field_index - 1`? We want to offset the field index by 1 when it's written in the uplink packet, to distinguish the 0th field from empty 0 bits at the end of the packet.

src/FCCode/UplinkConsumer.cpp
```diff
@@ -0,0 +1,110 @@
...
+        std::vector<bool> field_bit_arr = field_p->get_bit_array();
```
tanishqaggarwal: You're making a copy here; make a reference instead!!! Otherwise we're allocating memory.

src/FCCode/UplinkConsumer.cpp
```diff
@@ -0,0 +1,110 @@
...
+
+        // Clear field's bit array
+        field_bit_arr.clear();
+        field_bit_arr.resize(field_len);
```
tanishqaggarwal: This causes memory reallocation!!!

src/FCCode/UplinkConsumer.cpp
```diff
@@ -0,0 +1,110 @@
...
+    }
+}
+
+bool UplinkConsumer::validate_packet()
+{
+    size_t packet_bytes = (radio_mt_packet_len_f.get() + 7)/8;
+    bitstream bs (radio_mt_packet_f.get(), packet_bytes);
+    size_t field_index = 0, field_len = 0, bits_checked = 0, bits_seeked;
+    // Keep a bit map to prevent updating the same field twice
+    vector<bool> bit_map(registry.writable_fields.size(), 0);
```
tanishqaggarwal: This causes heap fragmentation since it's called on every control cycle. Use an array that's allocated once upon construction, instead.

src/FCCode/UplinkConsumer.cpp
```diff
@@ -0,0 +1,110 @@
...
+ 
+    while (bits_checked < radio_mt_packet_len_f.get())
+    {
+        // Get indices form bitstream
+        bits_checked += bs.nextN(index_size, reinterpret_cast<uint8_t*>(&field_index));
+
+        // Check if index is within writable_fields        
+        field_len = get_field_length(field_index);
```
tanishqaggarwal: Should be field_index - 1

src/FCCode/UplinkConsumer.cpp
tanishqaggarwal: I feel like this can be named something more intuitive? maybe something like `is_field_updated`?

src/FCCode/UplinkConsumer.cpp
tanishqaggarwal: If the concern is that get_bit_array() returns a const reference, go ahead and either ignore the constness with reinterpret_cast, or make a PR to commonsoftware to make it return a non-const reference instead. I would prefer reinterpret_cast since this is the only situation where we want non-constness of the returned bit array.

tanishqaggarwal: Also, it's best to use the `bit_array` type because it deletes all of the functions from `vector` that cause memory allocation/deallocation.

src/FCCode/UplinkConsumer.cpp
```diff
@@ -0,0 +1,110 @@
...
+
+        // Return in this case because indicates that packet is not aligned since
+        // we reached the end earlier than expected
+        if (bits_seeked != field_len) 
+            return false;
+
+        bits_checked += bits_seeked;
+    }
+    if (bs.byte_offset + 1 < bs.max_len) 
+        return false;
```
tanishqaggarwal: What does this check mean?

src/FCCode/UplinkConsumer.h
```diff
@@ -0,0 +1,121 @@
...
+
+#define change_bit_arr(w, x){\
+    w->set_bit_array(x);\
+    w->deserialize();\
+}
+
+#define read_bit_arr(w, x){\
+    w->serialize();\
+    x = w->get_bit_array();\
+}
```
tanishqaggarwal: These could be member functions.

src/FCCode/UplinkConsumer.cpp
athena255: no

test/test_fsw_uplink_consumer/test_fsw_uplink_consumer.cpp
```diff
@@ -0,0 +1,432 @@
...
+void setup() {
+    delay(2000);
+    Serial.begin(9600);
+    test_uplink_consumer();
+}
+
+void loop() {}
+#endif
+
+//cat test/test_uplink_consumer//test_uplink_consumer.cpp | grep "void test_" | sed 's/^void \(.*\)$/\1/' | sed 's/()/);/g'| sed -e 's/^/RUN_TEST(/'
```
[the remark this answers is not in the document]
athena255: it is a useful line in case i want to add new tests

test/test_fsw_uplink_consumer/test_fsw_uplink_consumer.cpp
```diff
@@ -0,0 +1,432 @@
...
+        radio_mt_packet_fp->set(mt_buffer);
+        field_map = std::map<std::string, size_t>();
+
+        // Setup field_map and assign some values
+        for (size_t i = 0; i < registry.writable_fields.size(); ++i)
+        {
+            auto w = registry.writable_fields[i];
+            field_map[w->name().c_str()] = i;
+            w->name();
+            from_ull(w, rand()); // no seed so should be the same each time
```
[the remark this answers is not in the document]
athena255: I added srand(1);

PR #102, opened 2019-12-10
ADCS Box Monitor

shihaocao:
This a control task that uses the ADCS Driver to pull data from the ADCS Box. It also checks to make sure the readings are within validity bounds. If they are outside validity bounds, state fields will still be populated, but out of bounds flags will also be raised.

There are also some small changes to the ADCS Driver to use the momentum->speed changes coming from Common Software.

There will be a later control task that issues commands to the ADCS Box.


lib/Drivers/ADCS.hpp
```diff
@@ -32,7 +30,11 @@ class ADCS : public I2CDevice {
      * @param i2c_wire The assoicated i2c wire
      * @param address The address on i2c bus
      */
+    #ifndef DESKTOP
     ADCS(const std::string &name, i2c_t3 &i2c_wire, unsigned char address);
+    #else
+    ADCS(const std::string &name, unsigned char address);
```
tanishqaggarwal: We shouldn't need an I2C address in a desktop constructor for ADCS

tanishqaggarwal: Honestly, best to get rid of the name in the constructor too. That's a relic from a previous Flight Software era, sorry

test/test_fsw_adcs_box/test_adcs_box.cpp
```diff
@@ -0,0 +1,166 @@
...
+class TestFixture {
+    public:
+        StateFieldRegistryMock registry;
+
+        // pointers to output statefields for easy access
+        ReadableStateField<f_vector_t>* rwa_speed_rd_fp;
+        ReadableStateField<f_vector_t>* rwa_torque_rd_fp;
+        ReadableStateField<int>* ssa_mode_fp;
+        ReadableStateField<f_vector_t>* ssa_vec_fp;
+        std::vector<ReadableStateField<float>*> ssa_voltages_fp{};
```
tanishqaggarwal: We don't need the braces here, I think? std::vector has a default initializer and you're just calling it explicitly.

src/FCCode/ADCSBoxMonitor.cpp
```diff
@@ -0,0 +1,136 @@
...
+    ssa_vec_f("adcs_monitor.ssa_vec", ssa_vec_sr),
+    ssa_voltage_sr(ssa::min_voltage_rd, ssa::max_voltage_rd, 8),
+    ssa_voltages_f(),
+    mag_vec_sr(imu::min_rd_mag, imu::max_rd_mag, 16*3), //referenced from I2C_Interface.doc
+    mag_vec_f("adcs_monitor.mag_vec", mag_vec_sr),
+    gyr_vec_sr(imu::min_rd_omega, imu::max_rd_omega, 16*3), //referenced from I2C_Interface.doc
+    gyr_vec_f("adcs_monitor.gyr_vec", gyr_vec_sr),
+    gyr_temp_sr(ssa::min_voltage_rd, ssa::max_voltage_rd, 16), //referenced from I2C_Interface.doc
+    gyr_temp_f("adcs_monitor.gyr_temp", gyr_temp_sr),
+    flag_sr(),
```
tanishqaggarwal: Good use of a shared serializer

src/FCCode/ADCSBoxMonitor.cpp
```diff
@@ -0,0 +1,136 @@
...
+    rwa_torque_rd_flag("adcs_monitor.torque_rd_flag", flag_sr),
+    ssa_vec_flag("adcs_monitor.ssa_vec_flag", flag_sr),
+    mag_vec_flag("adcs_monitor.mag_vec_flag", flag_sr),
+    gyr_vec_flag("adcs_monitor.gyr_vec_flag", flag_sr),
+    gyr_temp_flag("adcs_monitor.gyr_temp_flag", flag_sr)
+    {
+        //fill vector of statefields
+        char buffer [3];
+        for(unsigned int i = 0; i<num_sun_sensors;i++){
+            ssa_voltages_f.emplace_back(ReadableStateField<float>("adcs_monitor.ssa_voltage"+sprintf(buffer, "%u", i), ssa_voltage_sr));
```
tanishqaggarwal: This doesn't work. sprintf returns the number of characters written to the buffer, not the char*. You need to instead call sprintf in one line, and then add the _buffer_ to the base name string.

src/FCCode/ADCSBoxMonitor.cpp
```diff
@@ -0,0 +1,136 @@
...
+    adcs_system.get_rwa(&rwa_speed_rd,&rwa_torque_rd);
+    adcs_system.get_ssa_mode(&ssa_mode);
+    adcs_system.get_ssa_vector(&ssa_vec);
+    adcs_system.get_ssa_voltage(&ssa_voltages);
+    adcs_system.get_imu(&mag_vec, &gyr_vec, &gyr_temp);
+
+    //set statefields from internal containers
+    rwa_speed_rd_f.set(rwa_speed_rd);
+    rwa_torque_rd_f.set(rwa_torque_rd);
+    ssa_mode_f.set(ssa_mode);
```
tanishqaggarwal: There needs to be some more conditional logic on the SSA mode. Specifically, when the SSA mode goes from IN_PROGRESS to COMPLETE or FAILURE, we should set it back to IN_PROGRESS so that we can get the next sun vector reading.

In addition, maybe we should have a field that keeps track of the control cycle # of the last successful SSA measurement.

src/FCCode/ADCSBoxMonitor.hpp
```diff
@@ -0,0 +1,90 @@
...
+     */
+    ADCSBoxMonitor(StateFieldRegistry &registry, unsigned int offset, Devices::ADCS &_adcs);
+
+    /**
+    * @brief Gets inputs from the ADCS box and dumps them into the state
+    * fields listed below.
+    */
+    void execute() override;
+
+    static constexpr unsigned int num_sun_sensors = 20;
```
tanishqaggarwal: Is this not a constant available in adcs_constants.hpp? I'm surprised.

test/test_fsw_adcs_box/test_adcs_box.cpp
```diff
@@ -0,0 +1,166 @@
...
+            // initialize pointers to statefields
+            rwa_speed_rd_fp = registry.find_readable_field_t<f_vector_t>("adcs_monitor.rwa_speed_rd");
+            rwa_torque_rd_fp = registry.find_readable_field_t<f_vector_t>("adcs_monitor.rwa_torque_rd");
+            ssa_mode_fp = registry.find_readable_field_t<int>("adcs_monitor.ssa_mode");
+            ssa_vec_fp = registry.find_readable_field_t<f_vector_t>("adcs_monitor.ssa_vec");
+            
+            // fill vector of pointers to statefields
+            char buffer[3];
+            for(unsigned int i = 0; i<ADCSBoxMonitor::num_sun_sensors; i++){
+                ssa_voltages_fp.emplace_back(registry.find_readable_field_t<float>("adcs_monitor.ssa_voltage"+sprintf(buffer, "%u", i)));
```
tanishqaggarwal: Here we would probably use push_back, because there's no construction required for a pointer; it's a primitive type. Either `emplace` or `push` works though.

`emplace` really means "construct in-place", i.e. don't construct and _then_ copy into the vector.

lib/Drivers/ADCS.cpp
```diff
@@ -158,27 +164,42 @@ void ADCS::set_imu_gyr_temp_desired(const float desired){
 void ADCS::get_who_am_i(unsigned char* who_am_i) {
     i2c_point_and_read(WHO_AM_I, who_am_i, 1);
 }
-void ADCS::get_rwa(std::array<float, 3>* rwa_momentum_rd, std::array<float, 3>* rwa_ramp_rd) {
+void ADCS::get_rwa(std::array<float, 3>* rwa_speed_rd, std::array<float, 3>* rwa_ramp_rd) {
     unsigned char readin[12];
+    std::memset(readin, 0, sizeof(readin));
+
+    #ifndef DESKTOP
```
tanishqaggarwal: I think we should make the dummy values happen not only if the software is running on desktop, but also if the software is being unit tested. This way the unit tests can work on a Teensy with hardware disconnected. So you can replace this with

`#if !defined(DESKTOP) && !defined(UNIT_TEST)`

Also, for the sake of readability it's better to use _positive_ if statements as much as possible, e.g. use `#ifdef` instead of `#ifndef`. It's easy to read here, but as a general note it avoids the mental gymnastics of inverting a statement in your head.

test/test_fsw_adcs_box/test_adcs_box.cpp
```diff
@@ -0,0 +1,166 @@
...
+            assert(gyr_temp_fp);
+
+            //assert flag statefields have been found
+            assert(rwa_speed_rd_flag_p);
+            assert(rwa_torque_rd_flag_p);
+            assert(ssa_vec_flag_p);
+            assert(mag_vec_flag_p);
+            assert(gyr_vec_flag_p);
+            assert(gyr_temp_flag_p);
+
```
tanishqaggarwal: extra empty line

lib/Drivers/ADCS.cpp
```diff
@@ -198,11 +219,26 @@ void ADCS::get_imu(std::array<float,3>* mag_rd,std::array<float,3>* gyr_rd,float
...
     i2c_point_and_read(SSA_SUN_VECTOR, readin,6);
+    #else
+    for(int i = 0;i<6;i++){
+        readin[i] = 255;
+    }
+    #endif
+
+
```
tanishqaggarwal: Extra newline

test/test_fsw_adcs_box/test_adcs_box.cpp
```diff
@@ -0,0 +1,166 @@
...
+    //verify that all flags are set to true
+    //since temp bounds are all max - 1
+    //mocking using max output sets all flags to true
+    TEST_ASSERT_TRUE(tf.rwa_speed_rd_flag_p->get());
+    TEST_ASSERT_TRUE(tf.rwa_torque_rd_flag_p->get());
+    TEST_ASSERT_TRUE(tf.ssa_vec_flag_p->get());
+    TEST_ASSERT_TRUE(tf.mag_vec_flag_p->get());
+    TEST_ASSERT_TRUE(tf.gyr_vec_flag_p->get());
+    TEST_ASSERT_TRUE(tf.gyr_temp_flag_p->get());
+
```
tanishqaggarwal: Extra newline

tanishqaggarwal:
When you do add this task to the MainControlLoop, I think we should also add ADCS setup within the main control loop. I say the main control loop and not a control task because multiple control tasks share the same ADCS system.

lib/Drivers/ADCS.hpp
shihaocao: Do you want me to remove the name from all devices?

tanishqaggarwal: Let's just do it for ADCS for now

src/FCCode/ADCSBoxMonitor.hpp
shihaocao: would you like it to be?

tanishqaggarwal: If its not its totally ok

test/test_fsw_adcs_box/test_adcs_box.cpp
shihaocao: Cool, is there ever a case to use push over emplace?

tanishqaggarwal: If the copy overhead is minimal, I think it's less efficient to use emplace over push

tanishqaggarwal: The biggest thing is that emplace requires the space to already be allocated and push doesn't. So you would use push if the array hasn't been resized to capacity

src/FCCode/ADCSBoxMonitor.cpp
shihaocao: Big oops :(

shihaocao: I suppose this means technically you can add multiple statefields with the same name to the registry? (not that we would ever want to)

tanishqaggarwal: Yeah I'm not sure why this didn't just fail. There's supposed to be a safeguard against this

src/FCCode/ADCSBoxMonitor.hpp
nhz2: Yeah, this should be in `adcs_constants.hpp`. Do you if there is some way to join two PRs on common software and flight software? Maybe, flight software should depend on releases of common software, instead of the master branch, that way, an update to common software won't break everything on flight software.

tanishqaggarwal: That would be nice except we edit CommonSoftware so little that I don't think the extra overhead is worth it. @shihaocao can you make a PR to CommonSoftware and get that approved, and then fix this here?

shihaocao: Yeah I'll bake that edit into my common software PR with momentum -> speed

src/FCCode/ADCSBoxMonitor.cpp
shihaocao: @tanishqaggarwal @kkrol27 I'm thinking about pushing off the setting of the SSAMode to the ADCSControllerTask, that way GNC code can use the SSAMode as a flag to see if there's been a valid SSAVector calculation.

As a follow up question, is there/should there, be logic that holds off an attitude estimation calculation if there's not a new SSAVector? Should that go in AttitudeEstimator?

tanishqaggarwal: Pushing off the setting of the SSA flag makes sense to me. And I think the SSA vector should be updated by ADCSBoxMonitor if and only if the SSA mode is COMPLETE--that'll handle validity for us.

shihaocao: And what about AttitudeEstimator code? Currently, it'll chug away at a new calculation even if SSAVector wasn't new/Complete

tanishqaggarwal: This is true... @kkrol27 thoughts?

src/FCCode/MainControlLoop.cpp
```diff
@@ -27,6 +29,8 @@ MainControlLoop::MainControlLoop(StateFieldRegistry& registry,
       debug_task(registry, debug_task_offset),
       PIKSI_INITIALIZATION,
       piksi_control_task(registry, piksi_control_task_offset, piksi),
+      ADCS_INITIALIZATION,
+      adcs_monitor(registry, adcs_monitor_offset, adcs),
```
tanishqaggarwal: Need to call adcs.setup() inside MainControlLoop constructor body

test/test_fsw_adcs_box/test_adcs_box.cpp
```diff
@@ -0,0 +1,164 @@
...
+            ssa_vec_flag_p = registry.find_readable_field_t<bool>("adcs_monitor.ssa_vec_flag");
+            mag_vec_flag_p = registry.find_readable_field_t<bool>("adcs_monitor.mag_vec_flag");
+            gyr_vec_flag_p = registry.find_readable_field_t<bool>("adcs_monitor.gyr_vec_flag");
+            gyr_temp_flag_p = registry.find_readable_field_t<bool>("adcs_monitor.gyr_temp_flag");
+
+            //aseert the state fields are found
+            assert(rwa_speed_rd_fp);
+            assert(rwa_torque_rd_fp);
+            assert(ssa_mode_fp);
+            assert(ssa_vec_fp);
```
tanishqaggarwal: No need for these asserts anymore because find_readable_field_t takes care of them, as of my latest PR. At least, I think so? Can you double check this

tanishqaggarwal: Nevermind, it's not taken care of.

tanishqaggarwal: Yep, just checked--it's taken care of, so these assert statements can be removed.

src/FCCode/MainControlLoop.cpp
shihaocao: I don't think ADCS has a setup()

tanishqaggarwal: Resolved over PM, it does

src/FCCode/MainControlLoop.hpp
```diff
@@ -56,9 +61,10 @@ class MainControlLoop : public ControlTask<void> {
     #else
         static constexpr unsigned int debug_task_offset          =   5500;
         static constexpr unsigned int piksi_control_task_offset  =   6000;
-        static constexpr unsigned int attitude_estimator_offset  =  85500;
-        static constexpr unsigned int gomspace_controller_offset =  57500;
-        static constexpr unsigned int mission_manager_offset     =  62600;
+        static constexpr unsigned int adcs_monitor_offset        =  20500;
+        static constexpr unsigned int attitude_estimator_offset  =  35500;
+        static constexpr unsigned int gomspace_controller_offset =  56500;
+        static constexpr unsigned int mission_manager_offset     =  61600;
```
shihaocao: changed to make things slightly more consistent/not wrong

src/FCCode/ADCSBoxMonitor.cpp
```diff
@@ -0,0 +1,143 @@
...
+    ssa_vec_flag("adcs_monitor.ssa_vec_flag", flag_sr),
+    mag_vec_flag("adcs_monitor.mag_vec_flag", flag_sr),
+    gyr_vec_flag("adcs_monitor.gyr_vec_flag", flag_sr),
+    gyr_temp_flag("adcs_monitor.gyr_temp_flag", flag_sr)
+    {
+        //fill vector of statefields
+        //char buffer [3];
+        char buffer[50];
+        for(unsigned int i = 0; i<num_sun_sensors;i++){
+            sprintf(buffer,"adcs_monitor.ssa_voltage");
```
tanishqaggarwal: You need to memset this buffer to 0 on every iteration. Otherwise, your assumption below that you can just use `strlen` to get to the end of the string you've written is false, since the previous name will be inside the buffer

src/FCCode/ADCSBoxMonitor.cpp
```diff
@@ -0,0 +1,143 @@
...
+    if(input < min || input > max)
+        return true;
+    return false;
+}
+
+void ADCSBoxMonitor::execute(){
+
+    //create internal containers to read data
+    f_vector_t rwa_speed_rd{};
+    f_vector_t rwa_torque_rd{};
```
tanishqaggarwal: No need for the braces; std::array has a default initializer

src/FCCode/MainControlLoop.cpp
```diff
@@ -35,7 +39,17 @@ MainControlLoop::MainControlLoop(StateFieldRegistry& registry,
...
+{
+    //setup I2C bus for Flight Controller
+    #ifndef DESKTOP
+    Wire.begin(I2C_MASTER, 0x00, I2C_PINS_18_19, I2C_PULLUP_EXT, 400000, I2C_OP_MODE_IMM);
+    #endif
+    
+    //setup I2C devices
+    adcs.setup();
+    gomspace.setup();
+
```
tanishqaggarwal: extra newline

test/test_fsw_adcs_box/test_adcs_box.cpp
```diff
@@ -0,0 +1,144 @@
...
+            rwa_torque_rd_fp = registry.find_readable_field_t<f_vector_t>("adcs_monitor.rwa_torque_rd");
+            ssa_mode_fp = registry.find_readable_field_t<int>("adcs_monitor.ssa_mode");
+            ssa_vec_fp = registry.find_readable_field_t<f_vector_t>("adcs_monitor.ssa_vec");
+            
+            // fill vector of pointers to statefields
+            char buffer[50];
+            for(unsigned int i = 0; i<ADCSBoxMonitor::num_sun_sensors; i++){
+                sprintf(buffer,"adcs_monitor.ssa_voltage");
+                sprintf(buffer + strlen(buffer), "%u", i);
+                ssa_voltages_fp.push_back(registry.find_readable_field_t<float>(buffer));
```
tanishqaggarwal: Very surprised that this worked. See above comment

src/FCCode/ADCSBoxMonitor.hpp
kkrol27: Lol yeah I guess on the ADCS end the number of sun sensors is just hard coded everywhere

src/FCCode/ADCSBoxMonitor.cpp
kkrol27: Ahhh yeah. So basically I had imagined the sun vector on the flight computer only being updated when the SSA flag gets updated. It'll never really take longer than a control cycle or two for this to happen. If the flag comes back as a calculation failure this means we're in eclipse or the sun is on the docking face. In that case I'd say set the sun vector to NaN. If the SSA flag gets set to success we obviously just updated the vector.

kkrol27: I'd say too that using a slightly old sun vector isn't a big deal as we should be turning pretty slow.

src/FCCode/ADCSBoxMonitor.hpp
tanishqaggarwal: I guess it's probably not worth it then. Not something we can change anyways

src/FCCode/ADCSBoxMonitor.cpp
tanishqaggarwal: @shihaocao can you add logic for setting the sun vector to NaN upon failure?

shihaocao: For the sake of clarity, I'm going to make another PR that implements setting the sun_vector to NaN this later instead of putting it into this one.

shihaocao:
About SSA_Vec to NaN in the case of in progress or failure, I'll do that in another PR

PR #105, opened 2019-12-14
Psim estimator integration

tanishqaggarwal:
Mostly cosmetic improvements. The real changes are in `MainControlLoop.cpp`: adding estimator to the control cycle loop
I commented out some verbose print statements from QuakeManager. @athena255 just want your opinion on these

### DIFF
--- a/src/FCCode/AttitudeEstimator.cpp
+++ b/src/FCCode/AttitudeEstimator.cpp
@@ -15,7 +15,6 @@ AttitudeEstimator::AttitudeEstimator(StateFieldRegistry &registry,
     w_body_sr(-55, 55, 32*3),
     w_body_f("attitude_estimator.w_body", w_body_sr)
     {
-
         piksi_time_fp = find_readable_field<gps_time_t>("piksi.time", __FILE__, __LINE__),
         pos_vec_ecef_fp = find_readable_field<d_vector_t>("piksi.pos", __FILE__, __LINE__),
         ssa_vec_rd_fp = find_readable_field<f_vector_t>("adcs_box.sun_vec", __FILE__, __LINE__),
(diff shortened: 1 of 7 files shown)


src/FCCode/AttitudeEstimator.cpp
```diff
@@ -35,41 +35,27 @@ void AttitudeEstimator::execute(){
 void AttitudeEstimator::set_data(){
     data.t = (double)(((unsigned long)(piksi_time_fp->get() - pan_epoch))/(1e9L));
 
-    data.r_ecef = lin::Vector3d({
-        pos_vec_ecef_fp->get()[0], 
-        pos_vec_ecef_fp->get()[1], 
-        pos_vec_ecef_fp->get()[2]
-    });
+    const d_vector_t r_ecef = pos_vec_ecef_fp->get();
+    data.r_ecef = lin::Vector3d({r_ecef[0], r_ecef[1], r_ecef[2]});
```
kkrol27: This can just be `data.r_ecef = {r_ecef[0], r_ecef[1], r_ecef[2]};`

src/FCCode/AttitudeEstimator.cpp
```diff
@@ -35,41 +35,27 @@ void AttitudeEstimator::execute(){
...
-    data.s_body = lin::Vector3f({
-        ssa_vec_rd_fp->get()[0],
-        ssa_vec_rd_fp->get()[1],
-        ssa_vec_rd_fp->get()[2],
-    });
+    const f_vector_t mag_vec = mag_vec_fp->get();
+    data.b_body = lin::Vector3f({mag_vec[0], mag_vec[1], mag_vec[2]});
 
+    const f_vector_t f_vec = ssa_vec_rd_fp->get();
+    data.s_body = lin::Vector3f({f_vec[0], f_vec[1], f_vec[2]});
```
kkrol27: Same with these two, no need for the constructor!

kkrol27: Also why do you capture `mag_vec` by value and not just as a reference?

src/FCCode/AttitudeEstimator.hpp
```diff
@@ -37,7 +37,7 @@ class AttitudeEstimator : public TimedControlTask<void> {
      */
     void set_estimate();
     
-    const gps_time_t pan_epoch;
+    static const gps_time_t pan_epoch;
```
kkrol27: `extern gps_time_t const` here I think?

kkrol27: Also not sure if we want this to live somewhere more central given it'll be used by almost everything interfacing with `::gnc::*` stuff

kkrol27:
Also, I think it'd be good to have a compiler flag talk at some point. I found where platformio defines all of the default flags and we have some overlap atm.

src/FCCode/AttitudeEstimator.cpp
tanishqaggarwal: Oh nice, I didn't realize there was an implicit initializer-list copy constructor

src/FCCode/AttitudeEstimator.cpp
tanishqaggarwal: `get()` returns a copy, not a reference

src/FCCode/AttitudeEstimator.hpp
tanishqaggarwal: Not extern. It's a static class member. The value gets initialized by gnc constants so it's already centralized; this is just a copy for the class

kkrol27: > `extern gps_time_t const` here I think?

Nvm I didn't see this was in the class!

src/FCCode/AttitudeEstimator.cpp
kkrol27: Yep! Thought that'd be cleaner than having to construct objects all the time.

PR #108, opened 2019-12-16
Clocking updates

tanishqaggarwal:
- Added timing and memory statistics as state fields. The memory statistics don't work quite well yet but I'll test them and improve them if we need to. Currently flight software is using up about 32 KB on a Teensy 3.6, out of 256 KB RAM, which is pretty good.
- GPS time is now "propagated" so that the orbit estimator can get the most current GPS time.


kkrol27:
I'm not quite sure if we want to propagate the time forward like that (or if it even matters tbh). The timescale of our control loop vs what's important for estimation and control tasks is so tiny. I also prefer keeping the timestamp that came with the GPS reported position and velocity and not messing with it further if that makes sense. Propegating the GPS time forward would cause a mismatch there would it not?

tanishqaggarwal:
We still have a timestamp for incoming GPS position and velocity. I thought the purpose of the propagated time is for an estimator to be able to tell us where we are exactly _right now_. Relying on an old measurement of GPS time means that the estimated position could be hundreds of meters away from our actual position.

kkrol27:
> We still have a timestamp for incoming GPS position and velocity. I thought the purpose of the propagated time is for an estimator to be able to tell us where we are exactly _right now_. Relying on an old measurement of GPS time means that the estimated position could be hundreds of meters away from our actual position.

While all of this is true, I'd argue it doesn't matter and introduces complexity we just don't need. Thinking about relies on position measurements:
 * The magnetic field model is fairly constant over many kilometers (otherwise the B-dot scheme wouldn't really work).
 * The rendezvous algorithm relies on relative position and velocity which again is much more "constant" over short time scales. If we have an accurate position that's a second old and use that for the rendezvous algorithm, our results would change marginally (if at all).
 * If we worried about time differences on the order of a second or less we'd also need to revamp the position estimation scheme greatly (in terms of more inputs, perhaps a filter, and other little bells and whistles that wouldn't make a difference in terms of the mission in my opinion).

I'm not sure @nhz2 if you have a different opinion on this, but I think it'd just convolute things a little too much.

Also, the estimator is supposed to tell us where we are at a particular time, but for our purposes it's not super important that it's _right now_.

tanishqaggarwal:
I'm going to make some changes to the PR in light of @kkrol27's comments.

tanishqaggarwal:
Nevermind, actually, it's good to go.

I should be more precise: the code isn't actually propagating the time forward. It's just timestamping the GPS time with the system time at which it was collected. It also creates a readable state field that represents how much time has elapsed since the last successful GPS reading. I imagine the former item _might_ be useful; if not, it's not a big deal. The latter item is useful as diagnostic data on the ground.

tanishqaggarwal:
Just kidding, turns out Shihao already has a field for this thing. I'm going to just use his

kkrol27:
> Nevermind, actually, it's good to go.
> 
> I should be more precise: the code isn't actually propagating the time forward. It's just timestamping the GPS time with the system time at which it was collected. It also creates a readable state field that represents how much time has elapsed since the last successful GPS reading. I imagine the former item _might_ be useful; if not, it's not a big deal. The latter item is useful as diagnostic data on the ground.

Oh, okay, yeah for sure. Essentially we should pass the GPS time to the orbit estimator if GPS data was collected during a control loop. If data wasn't collected, then we pass the time sense the last GPS reading plus the last GPS readings.

tanishqaggarwal:
Yep, makes sense. Here's an updated version.

lib/utils/memuse.h
```diff
@@ -0,0 +1,129 @@
+#ifndef memuse_h_
```
tanishqaggarwal: This script doesn't actually work. Lol. Surprising how hard it is to find a decent function to compute RAM use

kkrol27: Lol oh well. It's alright.

src/FCCode/MissionManager.hpp
```diff
@@ -45,10 +45,6 @@ class MissionManager : public TimedControlTask<void> {
     WritableStateField<float>* adcs_min_stable_ang_rate_fp;
 
     // Fields that control overall mission state.
-    /**
-     * @brief Control cycle count, provided by ClockManager.
-     */
-    ReadableStateField<unsigned int>* control_cycle_count_fp;
```
shihaocao: Is the removal here because we are not making any decisions based on how many control cycles have passed?

src/FCCode/PiksiControlTask.cpp
```diff
@@ -1,50 +1,59 @@
...
-    baseline_pos_f("piksi.baseline_pos", baseline_pos_sr),
-    current_state_sr(0, 4, 2),
-    current_state_f("piksi.state", current_state_sr),
-    time_sr(),
-    time_f("piksi.time", time_sr)
+    pos_f("piksi.pos", Serializer<d_vector_t>(0,100000,100)),
+    vel_f("piksi.vel", Serializer<d_vector_t>(0,100000,100)),
+    baseline_pos_f("piksi.baseline_pos", Serializer<d_vector_t>(0,100000,100)),
+    current_state_f("piksi.state", Serializer<unsigned int>(4)),
+    since_good_cycles_f("piksi.last_good_cycle", Serializer<unsigned int>(1001)),
```
shihaocao: this should maybe be
`    since_good_cycles_f("piksi.since_good_cycles", Serializer<unsigned int>(1001)),`
for clarity

src/FCCode/PiksiControlTask.cpp
```diff
@@ -1,50 +1,59 @@
...
     //4 means no bytes
     //3 means CRC error on serial
     //5 means timing error exceed
     if(read_out == 3|| read_out == 4|| read_out == 5)
-        since_good_cycles += 1;
-    else 
-        since_good_cycles = 0;
+        since_good_cycles_f.set(since_good_cycles_f.get() + 1);
+    else {
+        last_good_reading_time_f.set(get_system_time());
```
shihaocao: this line sets the last fix time even if a sync_error or nsat_error occurs later

src/FCCode/PiksiControlTask.cpp
```diff
@@ -1,50 +1,59 @@
...
         //register callbacks and begin the serial port
         piksi.setup();
 
+        // Set initial values
+        const double nan = std::numeric_limits<double>::quiet_NaN();
         current_state_f.set(static_cast<unsigned int>(piksi_mode_t::no_fix));
+        pos_f.set({nan, nan, nan});
+        vel_f.set({nan, nan, nan});
+        baseline_pos_f.set({nan, nan, nan});
+        last_good_reading_time_f.set(get_system_time());
```
shihaocao: Should this be initialized to something else? Maybe nan too?
This seems to indicate that at the time of initialization we had a fix, which is not true

src/FCCode/PiksiControlTask.cpp
tanishqaggarwal: What's the difference? The suggestion looks the same to me as the code

src/FCCode/PiksiControlTask.cpp
tanishqaggarwal: Thanks, good catch

tanishqaggarwal: Relatedly, are you sure since_good_cycles is set in the right places? Currently it's set to zero even when there's a possibility of a no_fix, nsat, or sync error

src/FCCode/PiksiControlTask.cpp
tanishqaggarwal: Thinking about this further, I'm going to leave the value uninitialized, since that seems to be the most logically correct way to handle this situation.

src/FCCode/QuakeManager.cpp
```diff
@@ -14,22 +14,21 @@
...
-    control_cycle_count_fp = find_readable_field<unsigned int>("pan.cycle_no", __FILE__, __LINE__);
     snapshot_size_fp = find_internal_field<size_t>("downlink_producer.snap_size", __FILE__, __LINE__);
     radio_mo_packet_fp = find_internal_field<char*>("downlink_producer.mo_ptr", __FILE__, __LINE__);
     radio_mt_packet_fp = find_internal_field<char*>("uplink_consumer.mt_ptr", __FILE__, __LINE__);
     radio_err_fp = find_readable_field<int>("downlink_producer.radio_err_ptr", __FILE__, __LINE__);
     radio_mt_ready_fp = find_internal_field<bool>("uplink_consumer.mt_ready", __FILE__, __LINE__);
 
     // Initialize Quake Manager variables
-    last_checkin_cycle = control_cycle_count_fp->get();
+    last_checkin_cycle = control_cycle_count;
```
shihaocao: Is this ever incremented anywhere? Shouldn't this be pulled from a state field?

src/FCCode/MissionManager.hpp
tanishqaggarwal: It's because control_cycle_count is now a static member of the `TimedControlTask` class, so there's no need to create a state field to access the value.

src/FCCode/QuakeManager.cpp
tanishqaggarwal: See comment above about control cycle count

tanishqaggarwal: It's incremeneted in ClockManager

src/FCCode/PiksiControlTask.cpp
shihaocao: Using the new name, consecutive_fault_errors is a measure of how many consecutive errors that indicate something is DEFINITELY wrong with the piksi. And thus it's used to update a HAVT value later. It's main purpose is HAVT related, and thus is set to 0 so long as not read_out is not 3,4,5

test/test_fsw_piksi_control_task/test_piksi_control_task.cpp
```diff
@@ -250,11 +247,20 @@ void test_dead(){
...
+        //simulate that the piksi is not sending any data for 1000 control cycles.
+        //Make sure that the counter state fields are set correctl.
         tf.set_read_return(4);
-        for(int i = 0;i<1000;i++){
+        for(int i = 0;i<1000;i++) {
+                if (i % 100 == 0) {
+                        TEST_ASSERT_EQUAL(i, tf.fix_error_count_fp->get());
+                }
                 tf.execute();
+                TimedControlTaskBase::wait_duration(1);
```
shihaocao: This test runs a total of 1ms with a gap of 1 us between execute()'s. I don't think this breaks anything, but this was intended right? (intended to only span 1ms)

tanishqaggarwal: Yep. I do this so that the test assertion immediately afterwards on `delta_t` works

PR #110, opened 2019-12-16
Quake bug fixes

athena255:


src/FCCode/QuakeManager.cpp
```diff
@@ -121,6 +128,7 @@ bool QuakeManager::dispatch_write() {
         }
         // load the current 70 bytes of the buffer
        qct.set_downlink_msg(mo_buffer_copy + (packet_size*mo_idx), packet_size);
+       assert(max_snapshot_size/packet_size != 0);
```
tanishqaggarwal: This assertion will currently fail because the snapshot size produced by DownlinkProducer is less than 70 bytes. To make Flight Software not crash, can you edit Downlink Producer to pad the snapshot size

tanishqaggarwal: Hm...your build passed, I'm surprised. Probably because not enough control cycles are stepped for the assertion to fail.

Can you modify `test_real_fsw.py` to send "cycle.start" a bunch of times, to catch this?

tanishqaggarwal: Nevermind

PR #111, opened 2019-12-18
Quake/updates

tanishqaggarwal:
- Makes radio mode a state field so that mission manager can access it
- Eliminates Quake Manager dependence on other control tasks to produce its fields by allowing Downlink Producer to initialize its flows in the _body_ of the MainControlLoop's constructor rather than MainControlLoop's initializer list
- Updates here and there to unit tests to make everything pass


tanishqaggarwal:
@shihaocao The Quake radio mode is now accessible at "radio.mode". You might have to tweak the initialization order of the PiksiControlTask relative to QuakeManager in order to get your radio-consideration updates to work.

PR #112, opened 2019-12-19
Radio testing

athena255:


lib/Drivers/QLocate.cpp
```diff
@@ -50,27 +50,17 @@ int QLocate::query_is_functional_1()
...
     int status = consume(F("0\r"));
-    if (status != OK)
-        return status;
-    return OK;
+    return (status != OK) ? status : OK;
 }
 
 int QLocate::query_config_1()
 {
-#ifndef DESKTOP
```
tanishqaggarwal: Don't we need these #ifndef's? Otherwise the flight software won't run on a desktop.

tanishqaggarwal: Don't we need these #ifndef's? Otherwise the flight software won't run on a desktop.

PR #116, opened 2019-12-24
Dev/eeprom

fatimayousuf:


src/FCCode/EEPROMController.cpp
```diff
@@ -0,0 +1,67 @@
...
+    // copy the string name of the statefield into a char array
+    char field[statefields.at(i).length() + 1];
+    strcpy(field, statefields.at(i).c_str());
+    // get the pointer to that statefield and add it to the pointer array
+    pointers.push_back(find_readable_field<unsigned int>(field, __FILE__, __LINE__));
+    // add the address of the pointer to the address array
+    addresses.push_back(i*5);
+  }
+
+  control_cycle_count_fp = find_readable_field<unsigned int>("pan.cycle_no", __FILE__, __LINE__);
```
tanishqaggarwal: We actually don't need this state field anymore, the control cycle count is available via an internal counter in `TimedControlTask` called `control_cycle_count`.

test/test_fsw_EEPROM/test_EEPROM.cpp
```diff
@@ -0,0 +1,136 @@
...
+    tf.sat_designation_fp->set(2);
+    tf.control_cycle_count_fp->set(50);
+
+    // The period is set to 5. At the 50th control cycle, the EEPROM
+    // should write the values to the EEPROM
+    tf.eeprom_controller->execute();
+    #ifndef DESKTOP
+    for (size_t i = 0; i<tf.eeprom_controller->addresses.size(); i++){
+        // Expect 2 0 2 50
+        Serial.println(EEPROM.read(tf.eeprom_controller->addresses.at(i)));
```
tanishqaggarwal: Can you use a test assertion here rather than a serial print? Here and everywhere else in the file that you're printing

test/test_fsw_EEPROM/test_EEPROM.cpp
```diff
@@ -0,0 +1,136 @@
...
+        Serial.println(EEPROM.read(tf2.eeprom_controller->addresses.at(i)));
+    }
+    #endif
+
+}
+
+int test_control_task() {
+    UNITY_BEGIN();
+    RUN_TEST(test_task_initialization);
+    RUN_TEST(test_task_execute);
```
tanishqaggarwal: Can you do some intra-period testing? I.e. making sure new values are not written between ticks of the period

src/FCCode/EEPROMController.cpp
tanishqaggarwal: In your tests you can access the control cycle count with `TimedControlTaskBase::control_cycle_count`

src/FCCode/EEPROMController.cpp
```diff
@@ -0,0 +1,67 @@
+#include "EEPROMController.hpp"
+#ifndef DESKTOP
+#include <EEPROM.h>
+#endif
+
+EEPROMController::EEPROMController(StateFieldRegistry &registry, unsigned int offset, std::vector<std::string>& statefields)
```
tanishqaggarwal: Can you add the EEPROMController to the main control loop?

src/FCCode/EEPROMController.hpp
```diff
@@ -0,0 +1,58 @@
...
+     * respective address in EEPROM
+     */
+    void updateEEPROM();
+
+    /**
+     * @brief Checks if the EEPROM is empty or not. The default value
+     * of the EEPROM is 0xFF, or 255 in decimal. Returns true if empty
+     * and false if values are stored in the EEPROM.
+     */
+    bool checkEmpty();
```
tanishqaggarwal: Would be nice if we can use snake-case notation to be consistent across the board

src/FCCode/EEPROMController.cpp
tanishqaggarwal: Relatedly, you might have to move this constructor body into an `init()` function and call the init() function in the body of the MainControlLoop, similar to how DownlinkProducer currently does for the instantiation of its flows. This is to allow all control tasks to construct state fields and dump them into the registry before this Control Task starts doing things with state field pointers.

platformio.ini
```diff
@@ -86,10 +87,10 @@ src_filter = +<FCCode/> -<FCCode/main.cpp> +<native.cpp> +<flow_data.cpp>
...
-lib_deps = ${teensy_cli.lib_deps}
-build_flags = ${teensy_cli.build_flags} -D FUNCTIONAL_TEST -D UNITY_INCLUDE_DOUBLE
+platform = ${teensy_common.platform}
+framework = ${teensy_common.framework}
+lib_deps = ${teensy_common.lib_deps}
+build_flags = ${teensy_common.build_flags} -D FUNCTIONAL_TEST -D UNITY_INCLUDE_DOUBLE
 src_filter = +<teensy_stub.cpp> +<flow_data.cpp>
 upload_protocol = teensy-cli
-test_filter = test_fsw_*
+test_filter = test_fsw_EEPROM
```
tanishqaggarwal: I think teensy ci should run all FSW tests, not just EEPROM.

PR #117, opened 2019-12-25
Radio updates

tanishqaggarwal:
**This will need a careful review.**

This PR adds a "radio mode" in addition to the "radio state" that was originally the only state variable being tracked for Quake Manager. The radio mode controls whether or not the state machine that the "radio state" is based on is currently running. If radio mode is DISABLED, then the radio is fully disabled and the state machine does not execute. If radio mode is ACTIVE, then the radio steps through its state machine, beginning with config.

This change is needed in order to make mission manager compatible with the Quake Manager. The mission manager should be able to turn the radio on and off, but it can't do this by setting the state, because then it's interrupting the flow of the Quake Manager's state machine. Instead, the radio mode provides an easy mechanism for the mission manager to turn the radio on or off.


src/FCCode/QuakeManager.cpp
```diff
@@ -52,35 +61,50 @@ QuakeManager::~QuakeManager()
...
     // printf(debug_severity::info, "[Quake Info] Executing Quake Manager "
     //     "current radio_state %d, current control task state %d", 
-    //         radio_mode_f.get(), 
+    //         radio_state_f.get(), 
     //         qct.get_current_state());
+
     const radio_mode_t radio_mode = static_cast<radio_mode_t>(radio_mode_f.get());
     switch(radio_mode) {
-        case radio_mode_t::config:
+        case radio_mode_t::disabled:
```
athena255: Why would we request config state if the mode is disabled?

tanishqaggarwal: This ensures that when the mode is set back to "active", the Quake Manager starts at a reasonable initial state, like config.

PR #118, opened 2019-12-25
Cleanup items in preparation for Mission Manager

tanishqaggarwal:
Inevitably, there's some repository-wide cleanup to do before installing something as central as the mission manager.


src/FCCode/DockingController.cpp
```diff
@@ -2,17 +2,17 @@
...
+      is_turning_f("docksys.is_turning", Serializer<bool>())
 {
   add_readable_field(docked_f);
   add_readable_field(dock_config_f);
   add_readable_field(is_turning_f);
+}
 
+void DockingController::init() {
   docking_config_cmd_fp = find_writable_field<bool>("docksys.config_cmd", __FILE__, __LINE__);
 }
```
tanishqaggarwal: This state field pointer initialization is broken out into a separate field to allow Docking Controller access to the config_cmd field created by MissionManager.

PR #119, opened 2019-12-26
Faults base class

tanishqaggarwal:
Adds a fault class to the flight software. This fault class has persistence, so an error needs to be signaled multiple times before the fault is set.

I'm still looking for good, non-confusing names for these fault functions. Let me know if you have any good ideas.

It'd be great if everyone can review this before the merge, since this is so core to the infrastructure


src/FCCode/Fault.cpp
```diff
@@ -0,0 +1,45 @@
...
+    override_f(name + ".override", Serializer<bool>())
+{
+  set(default_setting);
+  override_f.set(false);
+  suppress_f.set(false);
+}
+
+bool Fault::add_to_registry(StateFieldRegistry& r) {
+    bool added = false;
+    added |= r.add_readable_field(static_cast<ReadableStateFieldBase*>(this));
```
athena255: True | False == True 
Maybe you meant &= ?

src/FCCode/Fault.hpp
```diff
@@ -0,0 +1,71 @@
...
+    // methods instead.
+    using ReadableStateField<bool>::set;
+    using ReadableStateField<bool>::get;
+
+    unsigned int& cc; // Control cycle count
+    unsigned int last_fault_time = 0; // Last control cycle # that the fault condition
+                                      // occurred
+
+    unsigned int persistence; // Persistence threshold for fault signal
+    unsigned int num_consecutive_faults = 0; // Number of consecutive fault condition
```
athena255: What is the meaning of current moment? Is it number of faults since the ground acknowledge the fault?

test/test_fsw_fault/test_fsw_fault.cpp
```diff
@@ -0,0 +1,106 @@
...
+    // Fault should not be signaled when signal condition happens under the persistence
+    for(int i = 0; i < 4; i++) {
+        control_cycle_count++; fault.signal();
+        TEST_ASSERT_FALSE(fault.is_faulted());
+    }
+
+    // Resetting the fault should not cause the fault to be signaled, but should cause
+    // the num_conseuctive_faults to reset. Resultantly, incrementing the control
+    // cycle count and signaling should not cause the fault to be signaled.
+    fault.unsignal();
```
athena255: Is this test off by one? num_consecutive must be strictly greater than persistence to cause a fault. num_consecutive is only 4 before this line.

src/FCCode/Fault.cpp
```diff
@@ -0,0 +1,45 @@
...
+    added |= r.add_writable_field(static_cast<WritableStateFieldBase*>(&suppress_f));
+    added |= r.add_writable_field(static_cast<WritableStateFieldBase*>(&override_f));
+    return added;
+}
+
+void Fault::signal() {
+    if (cc > static_cast<unsigned int>(last_fault_time) || cc == 0) {
```
athena255: Why do we increment when cc == 0?

src/FCCode/Fault.hpp
```diff
@@ -0,0 +1,71 @@
...
+                                             // occurrences at the current moment.
+
+    /**
+     * @brief State fields that can be set by the ground to suppress
+     * or forcibly signal the fault.
```
athena255: Could you explain the difference between a suppress and an override?

src/FCCode/Fault.cpp
tanishqaggarwal: Yeah I think this should probably return false if any one of the adds fails. I'll change it.

src/FCCode/Fault.cpp
tanishqaggarwal: If cc = 0 then we're on the first control cycle and therefore last_fault_time must also be zero. In that case `cc > last_fault_time` will be false, but we should still signal the condition. Therefore, I check for `cc == 0` separately.

src/FCCode/Fault.hpp
tanishqaggarwal: It's the number of control cycles before and including the current control cycle (i.e. the current moment) that the fault has been consecutively signaled.

src/FCCode/Fault.hpp
tanishqaggarwal: Sure, suppress means that `is_faulted` will always return false, and override means that `is_faulted` will always return true.

Suppress is useful because there are some fault conditions that we've already addressed on the ground and determined not to be a threat to the satellite. In that case, we want to stop the Fault object from causing the satellite to go into a safe state every time it detects the fault.

Override is useful because we might be detecting a problem on the ground that the satellite doesn't detect, or if we want to disable some feature of the satellite that we think is starting to become unsafe. In that case, override provides a mechanism for forcing the satellite into a safe state.

test/test_fsw_fault/test_fsw_fault.cpp
tanishqaggarwal: At this line, we're under the persistence limit and indeed don't expect the fault to be signaled, so I think the behavior is correct.

tanishqaggarwal: If you're referring to line 46's assertion, then I think the behavior there is correct as well because the behavior is signaled 6 times:
- Once at Line 34
- 4 times at line 39
- Once at line 45
Thanks for checking this though; off-by-one errors are obviously super important here.

test/test_fsw_fault/test_fsw_fault.cpp
```diff
@@ -0,0 +1,106 @@
...
+ * @brief Test normal behavior of fault without overrides or supressings.
+ */
+void test_fault_normal_behavior() {
+    unsigned int control_cycle_count = 0;
+
+    // Test constructors
+    Fault fault("fault", 5, control_cycle_count);
+    TEST_ASSERT_FALSE(fault.is_faulted());
+    Fault fault2("fault", 5, control_cycle_count, true);
+    TEST_ASSERT(fault2.is_faulted());
```
shihaocao: I think we should explicitly test that adding a second fault with the same name should return false and that it is not added twice.

tanishqaggarwal: This kind of testing is already captured by existing tests for the state field registry, isn't it?

shihaocao: right

tanishqaggarwal:
> A comment: If we're overriding or suppressing the fault, we can still accumulate num_consecutive_faults which may have unintended consequences when we turn off the suppression or override. Just wanted to make this behavior clear.

Good point. I'll make the following change: when overriding or suppressing, `num_consecutive_faults` is set to zero. This will alleviate any autonomous response that occurs when the ground commands are turned off.

> Also, I feel like there needs to be a way to command an unsiginal() from the ground. If we're not doing any kind of autonomous fault response, I think this is a necessary reset functionality. Perhaps through a writeable state field that is automatically set back to false?

Right now, `signal` and `unsignal` are an interface for autonomous behavior, and `override` and `suppress` are the interface for manual behavior. Do you see circumstances where we would want to call `unsignal` instead of just setting `suppress`?

shihaocao:
> Right now, signal and unsignal are an interface for autonomous behavior, and override and suppress are the interface for manual behavior. Do you see circumstances where we would want to call unsignal instead of just setting suppress?

I'm thinking of the following case: 

1. A fault flag becomes tripped. 
2. On the ground, we deduce what caused it, _and_ also realize a set of commands to upload to prevent it from occurring in the future.
3. We upload those commands, and merely ask the satellite to unsignal(), that way if somehow if it happens again, we can be notified.

Or slightly differently:

1. A fault flag becomes tripped.
2. We deduce what caused it, and realized there's nothing we can do to prevent it from being tripped again (say it'll happen every few days), but there is a set of commands to properly recover from it.
3. We upload the recovery commands, as well as the unsignal() command so that the satellite can continue to function until the fault condition occurs again (and still protect itself from the fault).

flight_and_test_envs.ini
```diff
@@ -60,6 +60,7 @@ lib_archive = false
 lib_compat_mode = off
 test_build_project_src = true
 test_filter = test_fsw_*
+test_ignore = test_fsw_EEPROM
```
tanishqaggarwal: @fatimayousuf

fatimayousuf: The test asserts in the test file are protected with #ifndef Desktop macros, so the test should work on a desktop

PR #120, opened 2019-12-26
Attitude computer

tanishqaggarwal:
This PR implements the pointing strategy control for the spacecraft.


src/FCCode/AttitudeComputer.cpp
```diff
@@ -0,0 +1,100 @@
...
+                {sat2target_r_body_hat(0), sat2target_r_body_hat(1), sat2target_r_body_hat(2)};
+
+            adcs_vec1_current_f.set(r_hat_body_vec);
+            adcs_vec1_desired_f.set({1,0,0});
+            adcs_vec2_current_f.set(sat2target_r_body_hat_vec);
+            adcs_vec2_desired_f.set({0,0,-1});
+        }
+        break;
+        case adcs_state_t::limited: {
+            // ???
```
tanishqaggarwal: This is a TODO

src/FCCode/AttitudeComputer.cpp
```diff
@@ -0,0 +1,100 @@
...
+    add_writable_field(adcs_vec1_desired_f);
+    add_writable_field(adcs_vec2_current_f);
+    add_writable_field(adcs_vec2_desired_f);
+
+    adcs_state_fp = find_writable_field<unsigned char>("adcs.state", __FILE__, __LINE__);
+    q_body_eci_fp = find_readable_field<f_quat_t>("attitude_estimator.q_body_eci", __FILE__, __LINE__);
+    ssa_vec_fp = find_readable_field<f_vector_t>("adcs_monitor.ssa_vec", __FILE__, __LINE__);
+    time_fp = find_readable_field<gps_time_t>("piksi.time", __FILE__, __LINE__);
+    pos_fp = find_readable_field<d_vector_t>("orbit.pos", __FILE__, __LINE__);
+    baseline_pos_fp = find_readable_field<d_vector_t>("orbit.baseline_pos", __FILE__, __LINE__);
```
tanishqaggarwal: These fields are in the ECI frame and are to-be-defined by an orbit estimator task.

tanishqaggarwal:
Closing to simplify an interface.

tanishqaggarwal:
Simplified.

src/FCCode/adcs_state_t.enum
```diff
@@ -1,11 +1,12 @@
...
+#define adcs_state_t_enum_
 
-// adcs.mode
-enum class adcs_mode_t {
+// adcs.state
+enum class adcs_state_t {
     startup,
     limited,
     zero_torque,
+    zero_L,
```
tanishqaggarwal: We planned to use this ADCS mode in post-docking operations to dump angular momentum into the magnetic field. @nhz2 Are we still planning on using this state in sim?

test/test_fsw_attitude_computer/test_fsw_attitude_computer.cpp
```diff
@@ -0,0 +1,109 @@
...
+
+void test_point_docking() {
+    TestFixture tf;
+
+    tf.adcs_state_fp->set(static_cast<unsigned char>(adcs_state_t::point_docking));
+    tf.pos_fp->set({0,2,0});
+    tf.pos_baseline_fp->set({0,0,3});
+    tf.attitude_computer->execute();
+    // TODO Test assertions 1 and 3 are incorrect here because of GNC rotate_frame instability. Fix when the
+    // issue is resolved.
```
tanishqaggarwal: @kkrol27 this is what I DMed you about

src/FCCode/AttitudeComputer.cpp
```diff
@@ -0,0 +1,110 @@
...
+            const f_vector_t ssa_vec = ssa_vec_fp->get();
+
+            if (!posdata_is_set) {
+                // We don't have a GPS reading. Point in the direction of the sun
+                // since that's the only vector we know reliably. It'll charge up
+                // our battery and we'll eventually have GPS coverage.
+                constexpr float nan = std::numeric_limits<float>::quiet_NaN();
+
+                adcs_vec1_current_f.set(ssa_vec);
+                adcs_vec1_desired_f.set({1,0,0});
```
shihaocao: I assume adcs_vec#'s are in body frame, to clarify is {1,0,0} = +x? and {0,1,0} = +y and {0,0,1} = +z?

Also shouldn't our desired be {+-sqrt(2)/2, +-sqrt(2)/2, 0}? Optimum charging is 45 degrees to sun right? @saa243

src/FCCode/AttitudeComputer.cpp
```diff
@@ -0,0 +1,110 @@
...
+                adcs_vec2_current_f.set({nan, nan, nan});
+                adcs_vec2_desired_f.set({nan, nan, nan});
+            }
+            else {
+                // We've got a GPS reading. Point in a direction that
+                // maximizes comms and power.
+                const lin::Vector3f ssa_normalized = {ssa_vec[0], ssa_vec[1], ssa_vec[2]};
+                lin::Vector3f r_cross_ssa = lin::cross(r_hat_body, ssa_normalized);
```
shihaocao: This issue might have already been thought about, but what happens if the ssa_vec aligns with r_hat_body?

What will lin::cross output? and will setting the current_f to that cause?

src/FCCode/AttitudeComputer.cpp
stewartaslan: Yes, we should change it to be `[+-sqrt(2)/2, +-sqrt(2)/2, 0]`. This will get better power generation and we should avoid pointing +X toward the sun because that's the antenna face and sun directly on the antennas is bad for comms (though this should also be tested). These vectors are in the body frame, but the important thing to remember here is that we're doing feedback control to minimize an error quaternion, and in this case the error is between 'adcs_vec1_current_f' and 'adcs_vec1_desired_f', so when the error is zero, these two vectors will be aligned.

shihaocao: Cool, so then in that case, we should choose `{-sqrt(2)/2, +-sqrt(2)/2, 0}` <- one of those two to point the +X face away from sun?

src/FCCode/MissionManager.cpp
```diff
@@ -77,11 +78,11 @@ void MissionManager::dispatch_startup() {
 
 void MissionManager::dispatch_detumble() {
     mission_mode_f.set(static_cast<unsigned int>(mission_mode_t::detumble));
-    adcs_mode_fp->set(static_cast<unsigned int>(adcs_state_t::detumble));
+    adcs_state_f.set(static_cast<unsigned int>(adcs_state_t::detumble));
     if (adcs_ang_rate_fp->get() <= adcs_min_stable_ang_rate_fp->get())
     {
         adcs_cmd_attitude_fp->set({2,2,2,2}); // TODO fix to a good value
```
shihaocao: why do we set a commanded attitude even during detumble?

test/test_fsw_attitude_computer/test_fsw_attitude_computer.cpp
```diff
@@ -0,0 +1,113 @@
...
+        attitude_computer = std::make_unique<AttitudeComputer>(registry, 0);
+
+        // Check that attitude computer creates its expected fields
+        adcs_vec1_current_fp = registry.find_writable_field_t<f_vector_t>("adcs.control.vec1.current");
+        adcs_vec1_desired_fp = registry.find_writable_field_t<f_vector_t>("adcs.control.vec1.desired");
+        adcs_vec2_current_fp = registry.find_writable_field_t<f_vector_t>("adcs.control.vec2.current");
+        adcs_vec2_desired_fp = registry.find_writable_field_t<f_vector_t>("adcs.control.vec2.desired");
+
+        // Set quaternion to non-rotating value
+        q_body_eci_fp->set({0,0,0,1});
```
shihaocao: to clarify, our notation for quaternions have the real component as i, j, k followed by the real part? that's what i'm guessing here?

src/FCCode/MissionManager.cpp
tanishqaggarwal: this is outdated, we don't do this anymore in missionmanager

test/test_fsw_attitude_computer/test_fsw_attitude_computer.cpp
tanishqaggarwal: Yep it's nx sin t/2, ny sin t/2, nz sin t/2, and then cos t/2. We should make a note on this somewhere.

src/FCCode/AttitudeComputer.cpp
tanishqaggarwal: This is a really good point, I had to think about it for a bit. Notice that aligning the +Z face with (r x sun vector) is a secondary pointing goal, so it's achieved on a "best effort" basis. The main goal is to just have the +X face point outwards to get best comms, which is achieved regardless of the value of the cross product. The second vector is there to lock down _some_ orientation if possible.

@nhz2 Can you back me up on this?

src/FCCode/AttitudeComputer.cpp
tanishqaggarwal: This makes sense to me. @nhz2 I'll implement this here, do you want to implement it on psim?

tanishqaggarwal: One thing we should keep in mind is that we want to pick the "closest" long edge to point toward the satellite.

src/FCCode/AttitudeComputer.cpp
nhz2: Yeah, r_hat cross sat2sun will always be perpendicular to the the sun vector. Trying to pointing the z direction perpendicular to the sun vector means a long face will be pointing at the sun.

src/FCCode/AttitudeComputer.cpp
```diff
@@ -0,0 +1,133 @@
...
+                size_t long_edge_choice = 0;
+                float min_rotation = 6.28; // Radians
+                for(size_t i = 0; i < 4; i++) {
+                    lin::Vector3f long_edge = {
+                        long_edges[i][0],
+                        long_edges[i][1],
+                        long_edges[i][2]
+                    };
+                    lin::Vector3f ssa_vector = {ssa_vec[0], ssa_vec[1], ssa_vec[2]};
+                    const float rotation_angle = acos(0.5 / (lin::dot(long_edge, ssa_vector)));
```
nhz2: Keep everything in dot products, don't actually use angles.

tanishqaggarwal: Agreed, the names are super confusing. I'll work out a more consistent naming scheme.

tanishqaggarwal:
I'm using the `_arr` suffix for vectors that aren't lin vectors and your convention for vectors that are lin vectors.

src/FCCode/AttitudeComputer.cpp
```diff
@@ -0,0 +1,134 @@
...
+                };
+                size_t long_edge_choice = 0;
+                float min_projection = -2;
+                for(size_t i = 0; i < 4; i++) {
+                    lin::Vector3f long_edge = {
+                        long_edges_arrs[i][0],
+                        long_edges_arrs[i][1],
+                        long_edges_arrs[i][2]
+                    };
+                    lin::Vector3f ssa_vector = {ssa_vec[0], ssa_vec[1], ssa_vec[2]};
```
shihaocao: `lin::Vector3f ssa_vector = {ssa_vec[0], ssa_vec[1], ssa_vec[2]};`
This can be moved out of the for loop right?

src/FCCode/AttitudeComputer.cpp
```diff
@@ -0,0 +1,134 @@
...
+                    const float projection = 1 / lin::dot(long_edge, ssa_vector);
+                    if (projection < min_projection) {
+                        min_projection = projection;
+                        long_edge_choice = i;
+                    }
+                }
```
shihaocao: I'm a bit confused about the calculations here.

If a certain long_edge matches up with the ssa_vec, dot should return 1, but projection would then also be 1 since 1 = 1/1.
But 1 is not less than -2, and thus that long edge will not be chosen

src/FCCode/AttitudeComputer.cpp
shihaocao: Unresolving because I was more concerned about the following cases:

If the cross product happens to produce a zero, will the code be robust to a 0/0? After thinking about it, I think this probability is pretty low thanks to float precision.

But I was more worried about the following case:
When r_hat_body is close to ssa_body, specifically within a few degrees and also within the sensor noise for the ssa_vector itself, I'm worried about a slightly varying ssa_body to send the secondary pointing objective swinging all about the circle of vectors perpendicular to r_hat. (and wasting battery).

test/test_fsw_attitude_computer/test_fsw_attitude_computer.cpp
```diff
@@ -0,0 +1,113 @@
...
+    PAN_TEST_ASSERT_EQUAL_FLOAT_VEC(f_vector_t({nan,nan,nan}).data(), tf.adcs_vec2_current_fp->get().data(), 1e-10);
+    PAN_TEST_ASSERT_EQUAL_FLOAT_VEC(f_vector_t({nan,nan,nan}).data(), tf.adcs_vec2_desired_fp->get().data(), 1e-10);
+
+    tf.pos_fp->set({0,2,0});
+    tf.attitude_computer->execute();
+    PAN_TEST_ASSERT_EQUAL_FLOAT_VEC(f_vector_t({0,1,0}).data(), tf.adcs_vec1_current_fp->get().data(), 1e-10);
+    PAN_TEST_ASSERT_EQUAL_FLOAT_VEC(f_vector_t({1,0,0}).data(), tf.adcs_vec1_desired_fp->get().data(), 1e-10);
+    PAN_TEST_ASSERT_EQUAL_FLOAT_VEC(f_vector_t({0,0,-1}).data(), tf.adcs_vec2_current_fp->get().data(), 1e-10);
+    PAN_TEST_ASSERT_EQUAL_FLOAT_VEC(f_vector_t({0,0,1}).data(), tf.adcs_vec2_desired_fp->get().data(), 1e-10);
+}
```
shihaocao: could you add tests that test the long edge calculations / choices?

src/FCCode/AttitudeComputer.cpp
tanishqaggarwal: Shit, you're right. The logic should be ">=". Nice catch.

src/FCCode/AttitudeComputer.cpp
tanishqaggarwal: Yeah  this works

test/test_fsw_attitude_computer/test_fsw_attitude_computer.cpp
tanishqaggarwal: Done. Thanks for catching.

src/FCCode/AttitudeComputer.cpp
tanishqaggarwal: Yeah, this is a good point. I'm going to add a threshold for the dot product.

src/FCCode/AttitudeComputer.cpp
```diff
@@ -0,0 +1,140 @@
...
+            baseline_pos_eci = baseline_pos_eci / lin::norm(baseline_pos_eci);
+            lin::Vector3f baseline_pos_body;
+            gnc::utl::rotate_frame(q_body_eci, baseline_pos_eci, baseline_pos_body);
+            const f_vector_t baseline_pos_body_arr =
+                {baseline_pos_body(0), baseline_pos_body(1), baseline_pos_body(2)};
+
+            adcs_vec1_current_f.set(r_hat_body_arr);
+            adcs_vec1_desired_f.set({1,0,0});
+            adcs_vec2_current_f.set(baseline_pos_body_arr);
+            adcs_vec2_desired_f.set({0,0,-1});
```
shihaocao: The baseline vector is the vector running from the base station piksi to the rover piksi.
Documentation from Swift Manual for `MSG_BASELINE_ECEF`

>This message reports the baseline solution in Earth Centered Earth Fixed (ECEF) coordinates. This baseline is the relative vector distance from the base station to the rover receiver.

So `basepine_pos_body_arr` runs from the other satellite to the current satellite. Since the docking face _is_ the -Z face, we should intend to align the +Z face with the baseline vector, and thus line 124 should be:

`adcs_vec2_desired_f.set({0,0,1});`

src/FCCode/AttitudeComputer.cpp
```diff
@@ -0,0 +1,140 @@
...
+                static_cast<float>(baseline_pos_eci_arr[2])
+            };
+            baseline_pos_eci = baseline_pos_eci / lin::norm(baseline_pos_eci);
+            lin::Vector3f baseline_pos_body;
+            gnc::utl::rotate_frame(q_body_eci, baseline_pos_eci, baseline_pos_body);
+            const f_vector_t baseline_pos_body_arr =
+                {baseline_pos_body(0), baseline_pos_body(1), baseline_pos_body(2)};
+
+            adcs_vec1_current_f.set(r_hat_body_arr);
+            adcs_vec1_desired_f.set({1,0,0});
```
shihaocao: I'm worried this might be physically impossible, or at least lead to issues:

Since both satellites are physically identical, it's impossible for both Satellites to be docked and have both +X faces facing Zenith.

This needs to be changed so that for example perhaps they both face horizon (but opposite one another), or we accept that they will approach not optimally set for docking, and then do something drastic once we get close.

Additional thoughts? @saa243 @kkrol27 @tanishqaggarwal @nhz2

stewartaslan: Yeah this is definitely a good catch. We can decide something now but I feel like it would be nice to be able to feel out what the comms situation is once we get to orbit before making a final decision (i.e. have this attitude be modifiable from the ground)

src/FCCode/AttitudeComputer.cpp
nhz2: Nice catch, I'll add the threshold to psim as well.

src/FCCode/AttitudeComputer.cpp
```diff
@@ -0,0 +1,140 @@
...
+                adcs_vec2_desired_f.set({nan, nan, nan});
+            }
+            else {
+                // We've got a GPS reading. Point in a direction that
+                // maximizes comms and power.
+                lin::Vector3f r_cross_ssa_body = lin::cross(r_hat_body, ssa_vec);
+                r_cross_ssa_body = r_cross_ssa_body / lin::norm(r_cross_ssa_body);
+
+                f_vector_t r_cross_ssa_body_arr;
+                if (lin::norm(r_cross_ssa_body) > alignment_threshold) {
```
nhz2: Move the `r_cross_ssa_body` normalization on line 90 to inside the if statement on line 93, otherwise `r_cross_ssa_body` is already normalized.

src/FCCode/AttitudeComputer.hpp
```diff
@@ -0,0 +1,57 @@
...
+    const ReadableStateField<f_vector_t>* ssa_vec_fp;
+
+    /**
+     * @brief Inputs required from orbit estimator for pointing.
+     */
+    const ReadableStateField<d_vector_t>* pos_fp;
+    const ReadableStateField<d_vector_t>* baseline_pos_fp;
+
+    /**
+     * @brief Fields used in control of attitude.
```
nhz2: add a comment that these are unit vectors in the body frame, or NAN

src/FCCode/AttitudeComputer.cpp
tanishqaggarwal: @nhz2 This makes sense, could you confirm?

src/FCCode/AttitudeComputer.cpp
tanishqaggarwal: Good point, thanks

src/FCCode/AttitudeComputer.cpp
tanishqaggarwal: I don't understand why the satellites can't have +X face outwards from Earth and be docked at the same time. Could you elaborate?

shihaocao: Explained over PR, moving discussion to issue #133

src/FCCode/AttitudeComputer.cpp
nhz2: yeah, that makes sense.

PR #121, opened 2019-12-26
Radio testing 2

athena255:
Changes to QLocate - SBDIX buffer was not being cleared before.


lib/Drivers/QLocate.cpp
```diff
@@ -49,27 +47,18 @@ int QLocate::query_is_functional_1()
 
 int QLocate::get_is_functional()
 {
-    int status = consume(F("0\r"));
-    if (status != OK)
-        return status;
-    return OK;
+    return consume(F("0\r"));
```
tanishqaggarwal: Lol makes sense

PR #124, opened 2019-12-27
implement momentum to speed changes

shihaocao:
This is a short PR that implements the ADCS momentum to speed changes in CommonSoftware.


shihaocao:
yeeted

PR #126, opened 2019-12-27
Mission manager (without faults)

tanishqaggarwal:
This implements the mission manager without hardware fault handling. I'll start addressing hardware faults once the fault PR is merged in.

Useful references (but outdated) in Startup and Master here:
https://cornellprod-my.sharepoint.com/personal/saa243_cornell_edu/_layouts/15/onedrive.aspx?id=%2Fpersonal%2Fsaa243%5Fcornell%5Fedu%2FDocuments%2FOAAN%20Team%20Folder%2FSubsystems%2FSoftware%2FDiagrams%2FOld

Verbose description that's slightly more up to date:
https://pan-software.readthedocs.io/en/latest/flight_software/mission_manager.html

From you guys, I'm particularly looking for test gaps that I should cover. Please watch out for this.


src/FCCode/MissionManager.cpp
```diff
@@ -1,128 +1,309 @@
...
+
+    // Mission has ended, so remove "follower" and "leader" designations.
+    set(sat_designation_t::undecided);
 }
 
 void MissionManager::dispatch_paired() {
-    // TODO
+    set(mission_state_t::paired);
+    set(sat_designation_t::undecided);
+    set(adcs_state_t::set_paired_gains);
```
tanishqaggarwal: @shihaocao I think the ADCS Box Controller should handle this ADCS state by setting the appropriate ADCS controller gains and then immediately moving to the `point_standby` state.

tanishqaggarwal: This "set paired gains" state is only meant to be a transient state to flag the ADCS controllers into modifying control gains.

src/FCCode/MissionManager.hpp
```diff
@@ -2,72 +2,135 @@
...
+#include "sat_designation_t.enum"
 
 class MissionManager : public TimedControlTask<void> {
    public:
     MissionManager(StateFieldRegistry& registry, unsigned int offset);
     void execute() override;
 
+    // Constants that drive state transitions.
+
+    static constexpr double close_approach_trigger_dist = 200; // in meters
```
nhz2: Why does this have to be so low, could it be really high, so the mode is always in close approach?

src/FCCode/MissionManager.hpp
```diff
@@ -2,72 +2,135 @@
...
     /**
-     * @brief Current angular rate of ADCS system in the body frame.
+     * @brief Current angular velocity of ADCS system in the body frame.
      **/
-    ReadableStateField<float>* adcs_ang_rate_fp;
+    ReadableStateField<f_vector_t>* adcs_ang_vel_fp;
     /**
      * @brief Minimum angular rate of ADCS system that can be considered "stable".
      **/
     WritableStateField<float>* adcs_min_stable_ang_rate_fp;
```
nhz2: For pointing, there is a maximum angular momentum that the wheels can hold, instead of a minimum rate body.

src/FCCode/adcs_state_t.enum
```diff
@@ -10,7 +10,9 @@ enum class adcs_state_t {
     detumble,
```
nhz2: What is `zero_L` mode? L usually stands for torque, but there is already a `zero_torque` mode.

src/FCCode/MissionManager.hpp
tanishqaggarwal: Aren't the pointing strategies different between far-field and near-field rendezvous? That's the goal of creating the "close approach" distinction

src/FCCode/MissionManager.hpp
tanishqaggarwal: This represents the minimum angular rate that can be considered "detumbled". Should i be using some other metric for determining if the satellite is detumbled?

src/FCCode/adcs_state_t.enum
tanishqaggarwal: From our notes from last semester, we wanted to have a mode that slowly dumps all of the stored wheel momentum of one of the satellites into the Earth's magnetic field, once we're docked. Are we still doing this?

src/FCCode/MissionManager.hpp
nhz2: Yes, `norm(estimator.angular_momentum_body)<const.MAXWHEELRATE*const.JWHEEL*0.2` is the condition I'm using is psim. You could replace `const.MAXWHEELRATE*const.JWHEEL*0.2` with something like `adcs_min_tumbling_angular_momentum_fp` if you want.

tanishqaggarwal: Why do we use the maximum wheel rate? Shouldn't our condition be much smaller than that? 

I guess 1/5th of maximum rate is pretty stable

src/FCCode/adcs_state_t.enum
nhz2: Okay, that makes sense.

src/FCCode/MissionManager.hpp
nhz2: It's the max wheel rate * wheel moment of inertia, so if the total angular momentum of the space craft is below this, the wheels can store all of the angular momentum of the satellite, and stop the satellite tumbling. The 1/5th is a safety factor and helps save power, because the wheels only need to spin 1/5 max speed.

tanishqaggarwal: Got it, so the goal is that the "pointing" mode takes over the rest of the detumbling action with wheels?

nhz2: Yeah.

tanishqaggarwal: Fixed in latest commit.

nhz2: Compare angular momentum, not angular rate.

tanishqaggarwal: The estimator returns `w_body`, not `L_body`. Shouldn't we compare against w then?

tanishqaggarwal: Fixed.

src/FCCode/AttitudeEstimator.cpp
```diff
@@ -24,6 +25,7 @@ AttitudeEstimator::AttitudeEstimator(StateFieldRegistry &registry,
         //Add outputs
         add_readable_field(q_body_eci_f);
         add_readable_field(w_body_f);
+        add_internal_field(l_body_f);
```
kkrol27: Just to be clear, is this the angular momentum of the spacecraft as a whole? Do we want to define this as a separate state field as it can just be calculated from the `w` of the body and wheels?

src/FCCode/MissionManager.cpp
```diff
@@ -1,128 +1,330 @@
...
+        set(mission_state_t::leader_close_approach);
+    }
 }
 
 void MissionManager::dispatch_leader_close_approach() {
-    // TODO
+    set(sat_designation_t::leader);
+    set(mission_state_t::leader_close_approach);
+    set(adcs_state_t::point_docking);
+    set(prop_mode_t::active);
```
kkrol27: Why is prop active during close approach for he leader and not during standard leader operations? Shouldn't it be off on the leader spacecraft essentially the entire time unless we have over-pressurization happening?

src/FCCode/MissionManager.cpp
kkrol27: Do we currently have a way of "unpairing" or are we just forgetting about that atm?

src/FCCode/MissionManager.hpp
kkrol27: I agree with you @tanishqaggarwal although we (like a lot of other constants floating around) may want to edit this and make it larger. I'd hope to be able to acquire CDGPS at farther than 200 m.
**EDIT**: Not at all important for this PR thought!

src/FCCode/MissionManager.cpp
tanishqaggarwal: You're right. This is a mistake. Thanks!

tanishqaggarwal:
> Do we currently have a way of "unpairing" or are we just forgetting about that atm?

For some reason I can't directly reply to your comment, but I think we can just unpair via a ground command to the mission state.

src/FCCode/AttitudeEstimator.cpp
tanishqaggarwal: I think it can just be computed. See the definition of execute() a little further below.

kkrol27: Ahhh okay so it's the angular momentum without the momentum stored in the wheels. Is there a task currently taking that as an input right now?

tanishqaggarwal: The mission manager uses the norm of L to determine whether it's finished detumbling. I imagine there might be other uses for the vector which is why I made it a state field

kkrol27: Ah okay gotcha. I'm wondering if it's just easier for us to work with `w_body` instead to keep state fields to a minimum. We can cut it out later thought if that happens to be the case.

kkrol27: Ah I see you and @nhz2 talked about this above too

src/FCCode/AttitudeEstimator.hpp
```diff
@@ -61,7 +62,10 @@ class AttitudeEstimator : public TimedControlTask<void> {
     ReadableStateField<f_quat_t> q_body_eci_f;
     Serializer<f_vector_t> w_body_sr;
     ReadableStateField<f_vector_t> w_body_f;
+    InternalStateField<lin::Vector3f> l_body_f;
```
nhz2: Add some comments about what this is, and units of N m s. Also fyi, L is usually torque, H is usually angular momentum, and J is usually moment of inertia in ADCS speak.

tanishqaggarwal: Right...I'm too used to the physics convention lol

src/FCCode/prop_mode_t.enum
```diff
@@ -0,0 +1,10 @@
+#ifndef prop_mode_t_enum
```
stewartaslan: It seems confusing and unnecessary to have both prop modes and prop states. Can we combine them into one list? It could be something like:

idle,
awaiting_pressurization,
pressurizing,
disabled

src/FCCode/MissionManager.cpp
```diff
@@ -1,128 +1,334 @@
...
+
+    // Detumble until satellite angular rate is below an allowable threshold
+    const float momentum = lin::norm(adcs_ang_momentum_fp->get());
+    const float threshold = rwa::max_speed_read * rwa::moment_of_inertia * detumble_safety_factor;
+    if (momentum <= threshold)
     {
-        adcs_cmd_attitude_fp->set({2,2,2,2}); // TODO fix to a good value
-        adcs_mode_fp->set(static_cast<unsigned int>(adcs_state_t::point_standby));
-        mission_mode_f.set(static_cast<unsigned int>(mission_mode_t::standby));
+        set(adcs_state_t::point_standby);
```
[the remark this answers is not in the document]
tanishqaggarwal: The goal of standby mode was to create a state where the satellite is fully commandable. If we reassert the pointing mode to "standby" on each control cycle that the mission mode is "standby", then we won't be able to manually command the spacecraft's attitude. Instead, the point_standby is set autonomously by the MissionManager only when the spacecraft is _transitioning_ into standby autonomously.

src/FCCode/MissionManager.cpp
```diff
@@ -1,128 +1,293 @@
...
+    set(prop_mode_t::active); // Active in order to enable ground propulsion commands.
+    set(radio_mode_t::active);
+
+    sat_designation_t const sat_designation =
+        static_cast<sat_designation_t>(sat_designation_f.get());
+
+    if (sat_designation == sat_designation_t::follower) {
+        adcs_paired_fp->set(false);
+        set(adcs_state_t::point_standby);
+        set(mission_state_t::follower);
```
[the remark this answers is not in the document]
tanishqaggarwal: The spacecraft's designation is "undecided" upon startup, so the spacecraft will stay in standby until the ground commands the spacecraft's designation to be either leader or follower.

src/FCCode/MissionManager.cpp
```diff
@@ -1,128 +1,293 @@
...
+    }
+    else if (sat_designation == sat_designation_t::leader) {
+        adcs_paired_fp->set(false);
+        set(adcs_state_t::point_standby);
+        set(mission_state_t::leader);
+    }
+    else {
+        // The mission hasn't started yet. Let the satellite subsystems do their thing.
```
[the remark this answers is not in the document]
tanishqaggarwal: When I add fault scenarios that cause transitions to standby, I'll be sure to make the transition change the pointing mode to standby. As I mentioned above, I don't think it's a good idea to assert the point_standby ADCS state on every control cycle that the satellite is in standby.

src/FCCode/MissionManager.cpp
```diff
@@ -1,128 +1,293 @@
...
+void MissionManager::dispatch_leader() {
+    set(sat_designation_t::leader);
+    set(mission_state_t::leader);
+    set(adcs_state_t::point_docking);
+    set(prop_mode_t::disabled);
+    set(radio_mode_t::active);
+
+    if (too_long_since_last_comms()) {
+        set(sat_designation_t::undecided);
+        set(adcs_state_t::point_standby);
```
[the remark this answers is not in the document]
tanishqaggarwal: Addressed above

src/FCCode/MissionManager.cpp
```diff
@@ -1,128 +1,293 @@
...
 
 void MissionManager::dispatch_docked() {
-    // TODO
+    set(mission_state_t::docked);
+    set(adcs_state_t::zero_torque);
+    set(prop_mode_t::disabled);
+    set(radio_mode_t::active);
+
+    // Mission has ended, so remove "follower" and "leader" designations.
+    set(sat_designation_t::undecided);
```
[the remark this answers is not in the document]
tanishqaggarwal: The transitions after docked are either docked -> paired -> standby or docked -> spacejunk -> standby. Upon entering standby we don't to trigger the logic that causes a transition to leader/follower, so at some point we need to reset the satellite designation. I thought this  would be a convenient place to do it.

src/FCCode/MissionManager.cpp
```diff
@@ -1,128 +1,293 @@
...
+    set(mission_state_t::standby);
 }
 
 void MissionManager::dispatch_spacejunk() {
-    // TODO
+    set(mission_state_t::spacejunk);
+    set(adcs_state_t::zero_L);
+    set(prop_mode_t::disabled);
+    set(radio_mode_t::active);
+    set(sat_designation_t::undecided);
```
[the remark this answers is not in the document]
tanishqaggarwal: Answered

src/FCCode/MissionManager.cpp
```diff
@@ -1,128 +1,293 @@
...
+    set(radio_mode_t::active);
+
+    // Mission has ended, so remove "follower" and "leader" designations.
+    set(sat_designation_t::undecided);
 }
 
 void MissionManager::dispatch_paired() {
-    // TODO
+    set(mission_state_t::paired);
+    set(sat_designation_t::undecided);
```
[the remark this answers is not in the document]
tanishqaggarwal: Answered

src/FCCode/MissionManager.hpp
```diff
@@ -2,72 +2,136 @@
...
 class MissionManager : public TimedControlTask<void> {
    public:
     MissionManager(StateFieldRegistry& registry, unsigned int offset);
     void execute() override;
 
+    // Constants that drive state transitions.
+    static constexpr double detumble_safety_factor = 0.2;
+
+    static constexpr double close_approach_trigger_dist = 200; // in meters
+    static constexpr double docking_trigger_dist = 1; // in meters
```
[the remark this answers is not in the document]
tanishqaggarwal: Sure, I'll change it

src/FCCode/MissionManager.hpp
```diff
@@ -2,72 +2,136 @@
...
+        static constexpr unsigned int deployment_wait = 15000; // ~30 mins
+    #else
+        static constexpr unsigned int deployment_wait = 10;
+    #endif
+    /**
+     * @brief Number of control cycles to wait before declaring "too long since comms".
+     */
+    #ifdef FLIGHT
+        static constexpr unsigned int max_radio_silence_duration = 
+            24 * 60 * 60 * 1000 / PAN::control_cycle_time_ms; // 24 hours
```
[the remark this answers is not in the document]
tanishqaggarwal: Not right now, but I noted on the Recurring Constants Review sheet that this is something we should definitely make ground-modifiable:

https://cornellprod-my.sharepoint.com/personal/saa243_cornell_edu/_layouts/15/Doc.aspx?sourcedoc=%7B10E398A9-3D68-44F8-BE1B-E4FA5DBECBC8%7D&file=Recurring%20Constants%20Review.xlsx&action=default&mobileredirect=true&cid=379ece68-9a69-46ee-9047-c15fa28f6c43

src/FCCode/MissionManager.hpp
```diff
@@ -2,72 +2,136 @@
...
-     * @brief Minimum angular rate of ADCS system that can be considered "stable".
-     **/
-    WritableStateField<float>* adcs_min_stable_ang_rate_fp;
+     * @brief Field created by AttitudeEstimator that tracks if the satellites
+     * should be treated as paired when computing attitude-related variables.
+     */
+    WritableStateField<bool>* adcs_paired_fp;
+
+    // Fields provided by Piksi and orbital estimation subsystems
+    const ReadableStateField<unsigned char>* piksi_mode_fp; // Piksi reading mode
```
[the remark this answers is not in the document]
tanishqaggarwal: Yep, that's what it's denoting. I'll make the comment better.

src/FCCode/prop_mode_t.enum
tanishqaggarwal: We can't. The difference between modes and states is that the mode decides whether a state machine is on/off, while the states are the actual stepping stones within the state machine. This is needed because in a mission state where we want propulsion to be active, we can't set the propulsion state--otherwise, the state machine will never progress to other states.

src/FCCode/MissionManager.cpp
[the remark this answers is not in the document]
stewartaslan: What I'm saying here is that if the spacecraft is in Follower mode, then gets sent to Standby (by a fault or something), the logic on line 179 will immediately set mission_state_t to "follower" which should only be done if the spacecraft is in Follower mode, not Standby, right? Maybe I'm not understanding what mission_state_t is?

src/FCCode/AttitudeEstimator.cpp
```diff
@@ -24,6 +24,11 @@ AttitudeEstimator::AttitudeEstimator(StateFieldRegistry &registry,
         //Add outputs
         add_readable_field(q_body_eci_f);
         add_readable_field(w_body_f);
+        add_internal_field(h_body_f);
+        add_writable_field(adcs_paired_f);
+
+        // Initialize flags
+        adcs_paired_f.set(false);
```
shihaocao: How is this adcs_paired_f state field different from the docked_f state field? Can we use docked_f instead?

shihaocao: Self resolved: docked_f notifies mission manager, and adcs_paired_f should be updated from mission manager

src/FCCode/MissionManager.cpp
```diff
@@ -1,128 +1,293 @@
...
+    set(mission_state_t::docked);
+    set(adcs_state_t::zero_torque);
+    set(prop_mode_t::disabled);
+    set(radio_mode_t::active);
+
+    // Mission has ended, so remove "follower" and "leader" designations.
+    set(sat_designation_t::undecided);
 }
 
 void MissionManager::dispatch_paired() {
```
shihaocao: what is the difference between paired and docked?

further, what are the entry conditions for paired?

src/FCCode/AttitudeEstimator.cpp
shihaocao: Actually lol, I have no idea what this statefield is supposed to represent, can you clarify when it should be true or false?

src/FCCode/MissionManager.cpp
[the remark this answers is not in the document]
shihaocao: If I'm correct, based on looking at the code, every entry condition into point standby is pre-empted with a 
`set(sat_designation_t::undecided);`, atleast within mission manager. We should take care to make sure we clear the designations when we enter standby mode from faults as well.

A bit of a future follow up question: I suppose that means we should have a FaultManager section of MissionManager? or even another TCT? or was that already planned

src/FCCode/MissionManager.cpp
```diff
@@ -1,128 +1,293 @@
...
+            set(mission_state_t::safehold);
             break;
     }
 }
 
 void MissionManager::dispatch_startup() {
-    // For now, do absolutely nothing. Wait for the startup controller to
-    // bring us out of this mode.
+    set(mission_state_t::startup);
+    set(adcs_state_t::startup);
```
shihaocao: what does will adcs do in startup? just be disabled?

src/FCCode/MissionManager.cpp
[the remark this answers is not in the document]
shihaocao: Another quick side thought because I can't think of where else to put it:

We should be careful to make sure that we dont have any faults that might cause us to leave startup mode and accidentally enter a standby mode early:
a) make sure faults don't happen in startup
or perhaps
b) make sure that throwing a fault sends us into startup and not standby if our deployment timer hasn't fully elapsed

src/FCCode/MissionManager.cpp
```diff
@@ -1,128 +1,293 @@
...
+}
+
 void MissionManager::execute() {
-    mission_mode_t mode = static_cast<mission_mode_t>(mission_mode_f.get());
+    mission_state_t mode = static_cast<mission_state_t>(mission_state_f.get());
+
+    if (mode != mission_state_t::startup) {
+        bool faulted = check_hardware_faults();
+        if (faulted) set(mission_state_t::safehold);
+    }
```
shihaocao: If I'm correct, this chunk of code can send us into safehold mode if we have a hardware fault before the deployment timer elapses, which will enable the radio before we're "allowed" to.

Fixes?:
Perhaps make sure faults wont happen in startup
Or alternatively, force check that we've deployed before checking faults?

I personally think we should implement the second option, but also try to make sure faults don't happen in startup

tanishqaggarwal: Doesn't the "if (mode != startup)" on line 53 take care of this?

src/FCCode/MissionManager.cpp
tanishqaggarwal: Yep. We cannot actuate _anything_ during the deployment period in startup

src/FCCode/MissionManager.cpp
tanishqaggarwal: Paired/spacejunk are sister states that branch off of docked. The "paired" satellite is responsible for steering the attitude of the docked pair of satellites, while the "spacejunk" satellite does nothing.

The ground is responsible for the transition from docked to paired/spacejunk.

src/FCCode/AttitudeEstimator.cpp
tanishqaggarwal: `docked_f` is set by the DockingController when the docking switch is pressed. It is a sufficient but not a necessary condition for docking to have actually taken place.

`adcs_paired_f` is the ADCS state that a satellite in the "paired" state operates in. See my answer below your question about the difference between the docked and paired states.

src/FCCode/MissionManager.cpp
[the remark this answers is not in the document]
tanishqaggarwal: Yep, Shihao's right. When I'm implementing fault logic I'll be careful to ensure that any transitions into standby due to faults will reset the satellite designation. Currently, though, the only possible transition into standby mode from follower -> standby resets the satellite designation.

Of course, we shouldn't rely on my carefulness; I'll write mission-level tests to ensure we can't get into a state loop.

src/FCCode/MissionManager.cpp
shihaocao: Yup. You’re right. I can’t believe I missed that.

tanishqaggarwal: Haha it happens

src/FCCode/MissionManager.cpp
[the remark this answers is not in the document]
shihaocao: I was thinking about this some more, why should the adcs mode be the only mode that is only set during transition?

If we instead had a method like `transition_to_mode()` that asserts all the necessary modes at the end of every dispatch decision tree, then we could command every single mode of the CubeSat.

Essentially, why not have every subsystem mode by commanded during transition only?
At least, at the bare minimum, I think all subsystem mode switching should be standardized and commanded at the same time, not half a control cycle ahead for just adcs.

src/FCCode/MissionManager.cpp
```diff
@@ -1,128 +1,293 @@
...
+
+    // Step 3. If no hardware faults exist, go into detumble
+    set(mission_state_t::detumble);
 }
 
 void MissionManager::dispatch_detumble() {
-    mission_mode_f.set(static_cast<unsigned int>(mission_mode_t::detumble));
-    adcs_mode_fp->set(static_cast<unsigned int>(adcs_state_t::detumble));
-    if (adcs_ang_rate_fp->get() <= adcs_min_stable_ang_rate_fp->get())
+    set(mission_state_t::detumble);
```
athena255: Shouldn't mission_state_t be in detumble when we enter this function? If so, why set it and not assert it?

src/FCCode/MissionManager.cpp
```diff
@@ -1,128 +1,293 @@
...
-        adcs_mode_fp->set(static_cast<unsigned int>(adcs_state_t::point_standby));
-        mission_mode_f.set(static_cast<unsigned int>(mission_mode_t::standby));
+        set(adcs_state_t::point_standby);
+        set(mission_state_t::standby);
     }
 }
 
 void MissionManager::dispatch_initialization_hold() {
-    // TODO
+    set(mission_state_t::initialization_hold);
```
athena255: Why set mission_state_t to initialization_hold when mission_state_t should already be in initialization_hold upon entering this function.

src/FCCode/MissionManager.cpp
```diff
@@ -1,128 +1,293 @@
...
+    // Stay in this state until ground commands fix the satellite and
+    // command the satellite out of this state.
 }
 
 void MissionManager::dispatch_follower() {
-    
-    // TODO
-}
+    set(sat_designation_t::follower);
+    set(mission_state_t::follower);
```
athena255: Why set mission_state_t to follower?

src/FCCode/MissionManager.cpp
```diff
@@ -1,128 +1,293 @@
...
+    }
+
+    if (distance_to_other_sat() < docking_trigger_dist) {
+        set(mission_state_t::docking);
+    }
 }
 
 void MissionManager::dispatch_standby() {
-    // TODO
+    set(mission_state_t::standby);
```
athena255: why set mission_state_t to standby?

src/FCCode/MissionManager.cpp
```diff
@@ -1,128 +1,293 @@
...
+    else {
+        // The mission hasn't started yet. Let the satellite subsystems do their thing.
+    }
 }
 
-void MissionManager::dispatch_leader_close_approach() {
-    // TODO
+void MissionManager::dispatch_leader() {
+    set(sat_designation_t::leader);
+    set(mission_state_t::leader);
```
athena255: Why set mission_state_t to leader?

src/FCCode/MissionManager.cpp
```diff
@@ -1,128 +1,293 @@
...
+    set(prop_mode_t::disabled);
+    set(radio_mode_t::active);
+
+    if (too_long_since_last_comms()) {
+        set(sat_designation_t::undecided);
+        set(adcs_state_t::point_standby);
+        set(mission_state_t::standby);
+    }
+
+    if (distance_to_other_sat() < docking_trigger_dist) {
```
athena255: Should this be an else if here? Or is it okay to switch to docking even if we have no comms for a while?

src/FCCode/MissionManager.cpp
```diff
@@ -1,128 +1,293 @@
...
+
+    docking_config_cmd_f.set(true);
+    if (docked_fp->get()) {
+        set(mission_state_t::docked);
+    }
 }
 
 void MissionManager::dispatch_docked() {
-    // TODO
+    set(mission_state_t::docked);
```
athena255: Why set mission_state_t to docked instead of checking that mission_state_t is in docked?

src/FCCode/MissionManager.cpp
```diff
@@ -1,128 +1,293 @@
...
+    set(mission_state_t::paired);
+    set(sat_designation_t::undecided);
+    adcs_paired_fp->set(true);
+    set(adcs_state_t::point_standby);
+    set(mission_state_t::standby);
 }
 
 void MissionManager::dispatch_spacejunk() {
-    // TODO
+    set(mission_state_t::spacejunk);
```
athena255: Same as prior comments, why set mission_state_t to the state that it is suppose to be in?

src/FCCode/MissionManager.cpp
```diff
@@ -1,128 +1,293 @@
...
+
+void MissionManager::dispatch_manual() {
+    // Do nothing.
+}
+
+
+double MissionManager::distance_to_other_sat() const {
+    const d_vector_t dr = propagated_baseline_pos_fp->get();
+    if (std::isnan(dr[0])) return dr[0];
+    lin::Vector3d dr_vec = {dr[0], dr[1], dr[2]};
```
athena255: dr can never be null?

src/FCCode/MissionManager.hpp
```diff
@@ -2,72 +2,136 @@
...
     InternalStateField<unsigned char>* radio_mode_fp;
+    InternalStateField<unsigned int>* last_checkin_cycle_fp;
 
     // Fields that control overall mission state.
     /**
      * @brief Current mission mode (see mission_mode_t.enum)
      */
-    Serializer<unsigned char> mission_mode_sr;
-    WritableStateField<unsigned char> mission_mode_f;
+    WritableStateField<unsigned char> mission_state_f;
```
athena255: Sort of off-topic, but why do we not just template on the enum. That way, we don't need to keep static casting?

src/FCCode/MissionManager.hpp
```diff
@@ -2,72 +2,136 @@
...
     /**
      * @brief True if the satellite has exited the deployment timing phase.
      */
-    Serializer<bool> is_deployed_sr;
     ReadableStateField<bool> is_deployed_f;
```
athena255: Why make this a ReadableStateField? From what I understand, if we have not deployed yet, then we will not be able to use the radio anyway. If we are able to use the radio, then it follows that we have deployed.

src/FCCode/mission_state_t.enum
```diff
@@ -1,20 +1,21 @@
-#ifndef mission_mode_t_enum_
-#define mission_mode_t_enum_
+#ifndef mission_state_t_enum_
+#define mission_state_t_enum_
 
-// pan.mode
-enum class mission_mode_t {
+// pan.state
+enum class mission_state_t {
```
athena255: It might be helpful to put comments explaining what each of these states mean and the assumptions/specifications associated with each state.

test/test_fsw_attitude_estimator/test_attitude_estimator.cpp
```diff
@@ -33,6 +35,8 @@ class TestFixture {
                 // initialize pointers to statefields
                 q_body_eci_fp = registry.find_readable_field_t<f_quat_t>("attitude_estimator.q_body_eci");
                 w_body_fp = registry.find_readable_field_t<f_vector_t>("attitude_estimator.w_body");
+                h_body_fp = registry.find_internal_field_t<lin::Vector3f>("attitude_estimator.h_body");
```
athena255: This test does not seem to test anything.

test/test_fsw_mission_manager/test_fixture.cpp
```diff
@@ -0,0 +1,141 @@
...
+        TEST_ASSERT_NOT_EQUAL(static_cast<unsigned int>(state_it), adcs_state_fp->get());
+    }
+}
+
+void TestFixture::assert_ground_uncommandability(prop_mode_t exception_mode) {
+    for(prop_mode_t mode_it : prop_modes) {
+        if (mode_it == exception_mode) continue;
+        set(mode_it);
+        step();
+        TEST_ASSERT_NOT_EQUAL(static_cast<unsigned int>(mode_it), prop_mode_fp->get());
```
athena255: I don't understand what is being tested here. How does this ensures that no state except the given state can be achieved?

test/test_fsw_mission_manager/test_fixture.cpp
```diff
@@ -0,0 +1,141 @@
...
+    // TODO
+}
+
+// Set the distance between the two satellites.
+void TestFixture::set_sat_distance(double dist) { propagated_baseline_pos_fp->set({dist, 0, 0}); }
+
+// Set the # of control cycles that comms has not been established
+// with the ground.
+void TestFixture::set_comms_blackout_period(int ccno) {
+    last_checkin_cycle_fp->set(mission_manager->control_cycle_count - ccno);
```
athena255: If mission_manager happens to have a lower value than ccno, then this field will be negative/very large.

test/test_fsw_mission_manager/test_mission_manager.cpp
```diff
@@ -1,76 +1,266 @@
...
+    tf.step();
+    tf.check(sat_designation_t::undecided);
+    tf.check(mission_state_t::detumble);
+    tf.check(adcs_state_t::detumble);
+    tf.check(prop_mode_t::disabled);
+    tf.check(radio_mode_t::active);
+
+    // If satellite is no longer tumbling, spacecraft exits detumble mode
+    // and starts pointing in the expected direction.
+    tf.set_ang_rate(threshold - delta);
```
athena255: Does this mean that we expect angular rate to decrease by 10% every control cycle?

test/test_fsw_mission_manager/test_mission_manager.cpp
```diff
@@ -1,76 +1,266 @@
...
-    std::unique_ptr<MissionManager> mission_manager;
-    // Output state fields from mission manager
-    WritableStateField<unsigned char>* mission_mode_fp;
-    WritableStateField<unsigned char>* sat_designation_fp;
+    // Standby -> follower transition test
+    {
+        TestFixture tf(mission_state_t::standby);
+        tf.set(sat_designation_t::follower);
+        tf.step();
+        TEST_ASSERT_FALSE(tf.adcs_paired_fp->get());
```
athena255: Maybe ASSERT_TRUE tf.adcs_paired_fp->get() before the tf.step() so that we know the value definitely changed during the step().

src/FCCode/MissionManager.cpp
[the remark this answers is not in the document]
athena255: I agree with Shihao that it would be cleaner to separate out the transitions. Setting a state can be considered a transition since it is a state change, and it does not make sense to me that the first thing to do upon entering a dispatch function is to set states of various subsystems. The states should have been set prior to entering the dispatch function.

src/FCCode/AttitudeEstimator.cpp
```diff
@@ -46,16 +51,20 @@ void AttitudeEstimator::set_data(){
...
     q_body_eci_f.set(q_temp);
 
-    f_vector_t w_temp;
-    w_temp[0] = estimate.w_body(0);
-    w_temp[1] = estimate.w_body(1);
-    w_temp[2] = estimate.w_body(2);  
+    f_vector_t w_temp = { estimate.w_body(0), estimate.w_body(1), estimate.w_body(2) };
     w_body_f.set(w_temp);
+
+    lin::Vector3f wvec = {w_temp[0], w_temp[1], w_temp[2]};
```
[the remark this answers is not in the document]
tanishqaggarwal: f_vector_t is just a renaming for std::array<float, 3>. Vector3f is a special vector math class designed by Kyle that makes math with vectors a little easier

src/FCCode/MissionManager.cpp
```diff
@@ -1,128 +1,293 @@
...
+    is_deployed_f("pan.deployed", Serializer<bool>()),
+    deployment_wait_elapsed_f("pan.deployment.elapsed", Serializer<unsigned int>(0, 15000, 32)),
+    sat_designation_f("pan.sat_designation", Serializer<unsigned char>(2))
 {
-    add_writable_field(mission_mode_f);
+    add_writable_field(adcs_state_f);
+    add_writable_field(docking_config_cmd_f);
+    add_writable_field(mission_state_f);
     add_readable_field(is_deployed_f);
+    add_readable_field(deployment_wait_elapsed_f);
```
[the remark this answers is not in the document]
tanishqaggarwal: The EEPROM controller, which needs to save the the deployment counter in case the satellite powers down during the deployment phase, currently only accepts ReadableStateFields as inputs

src/FCCode/MissionManager.cpp
```diff
@@ -1,128 +1,293 @@
...
+    // Step 1. Wait for the deployment timer length.
+    if (deployment_wait_elapsed_f.get() < deployment_wait) {
+        deployment_wait_elapsed_f.set(deployment_wait_elapsed_f.get() + 1);
+        return;
+    }
+
+    // Step 2. Check for hardware faults that would necessitate
+    // going into an initialization hold
+    if (check_hardware_faults()) {
+        set(mission_state_t::initialization_hold);
```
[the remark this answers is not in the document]
tanishqaggarwal: Shoot yeah you're right. The unit test didn't catch this since I haven't implemented fault testing yet

src/FCCode/MissionManager.cpp
tanishqaggarwal: Mostly just as a protection against radiation bitflips

src/FCCode/MissionManager.cpp
tanishqaggarwal: Answered above for detumble

src/FCCode/MissionManager.cpp
tanishqaggarwal: Answered above for detumble

src/FCCode/MissionManager.cpp
tanishqaggarwal: Answered above for detumble

src/FCCode/MissionManager.cpp
tanishqaggarwal: Answered above for detumble

src/FCCode/MissionManager.cpp
tanishqaggarwal: I think it's actually desirable to switch to docking. We don't want the satellites suddenly changing their attitude if they're less than a meter apart

src/FCCode/MissionManager.cpp
tanishqaggarwal: Answered above for detumble

src/FCCode/MissionManager.cpp
tanishqaggarwal: Answered above for detumble

src/FCCode/MissionManager.cpp
tanishqaggarwal: dr isn't a pointer, it's an `std::array<double, 3>`. So instead of null we use NaN as an "invalid value" signal

src/FCCode/MissionManager.hpp
tanishqaggarwal: That's an interesting idea. How does that work? I'm not familiar.

src/FCCode/MissionManager.hpp
tanishqaggarwal: See comment above about deployment_wait_elapsed_f

src/FCCode/mission_state_t.enum
tanishqaggarwal: I explained this inside the RTM documentation, but yeah maybe I can add a few sentences inside code.

test/test_fsw_attitude_estimator/test_attitude_estimator.cpp
tanishqaggarwal: Since it's just a wrapper around a GNC function, I don't think there's anything to test here. The actual GNC functionality is already unit-tested (or is planned to be unit-tested) within psim.

test/test_fsw_mission_manager/test_mission_manager.cpp
tanishqaggarwal: Nah. "delta" is just a small amount to make the conditions change. I'm actually not sure how much we expect our rate to decrease by on every control cycle

src/FCCode/AttitudeEstimator.cpp
```diff
@@ -24,6 +24,11 @@ AttitudeEstimator::AttitudeEstimator(StateFieldRegistry &registry,
...
         add_readable_field(q_body_eci_f);
         add_readable_field(w_body_f);
+        add_internal_field(h_body_f);
+        add_writable_field(adcs_paired_f);
+
+        // Initialize flags
+        adcs_paired_f.set(false);
     }
 
 void AttitudeEstimator::execute(){
```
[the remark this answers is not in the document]
tanishqaggarwal: Yikes! Thanks for catching, I'll fix this

test/test_fsw_mission_manager/test_fixture.cpp
tanishqaggarwal: If we try to set a subsystem state, then on the next control cycle it's overridden with the default behavior of the mission state.

test/test_fsw_mission_manager/test_fixture.cpp
tanishqaggarwal: For most of the unit test `control_cycle_count = 0` but this test still works with nonzero values of `ccno`. I think it's because the signage of differences is handled as we'd expect.

What would you suggest I do to make the behavior a little cleaner? I'm not a fan of changing the control cycle count within this function since that might be a property of a testcase that we want to preserve

test/test_fsw_mission_manager/test_mission_manager.cpp
tanishqaggarwal: Yeah this is clearer, thanks for the suggestion

tanishqaggarwal: Actually, in this place it's not always true that the ADCS state will be "paired". So we make an assertion before step().

src/FCCode/MissionManager.cpp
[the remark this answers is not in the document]
tanishqaggarwal: Yeah, I agree. I still want to maintain subsystem assertions on certain mission states (this helps protect against radiation bitflips of the state) but I agree the transition logic can be a little cleaner. I'll make some changes

test/test_fsw_mission_manager/test_fixture.cpp
athena255: You can check that control cycle count > ccno and assert(fail) if not.

src/FCCode/mission_state_t.enum
athena255: A comment to see the RTM documentation or a link to it would suffice.

src/FCCode/FieldCreatorTask.hpp
```diff
@@ -11,41 +11,23 @@
...
       FieldCreatorTask(StateFieldRegistry& r) : 
         ControlTask<void>(r),
-        adcs_cmd_attitude_f("adcs.cmd_attitude", Serializer<f_quat_t>()),
-        adcs_ang_rate_f("adcs.ang_rate", Serializer<float>(0, 10, 4)),
-        adcs_min_stable_ang_rate_f("adcs.min_stable_ang_rate", Serializer<float>(0, 10, 4)),
-        pos_f("orbit.pos", Serializer<d_vector_t>(0,100000,100)),
-        pos_baseline_f("orbit.baseline_pos", Serializer<d_vector_t>(0,100000,100)),
-        docking_config_cmd_f("docksys.config_cmd", Serializer<bool>()),
+        pos_f("orbit.pos", Serializer<d_vector_t>(0, 100000, 100)),
+        pos_baseline_f("orbit.baseline_pos", Serializer<d_vector_t>(0, 100000, 100)),
```
shihaocao: To clarify orbit pos and orbit baseline are based off of COM right?

src/FCCode/AttitudeEstimator.cpp
shihaocao: Another follow-up question, as it stands is the plan to wait for a ground command to set `adcs_paried_f` to `true`?

src/FCCode/MissionManager.cpp
```diff
@@ -1,129 +1,244 @@
...
+
+    if (mode == mission_state_t::startup) {
+        set(radio_mode_t::disabled);
+    }
+    else {
+        set(radio_mode_t::active);
+        bool faulted = check_hardware_faults();
+        if (faulted) {
+            transition_to_state(mission_state_t::safehold,
+                adcs_state_t::startup,
```
shihaocao: Can you put in quick descriptions in the adcs_state_t.enum file for each mode so I can have a better picture of whats happening for modes that haven't been defined in AttitudeComputer?

src/FCCode/MissionManager.cpp
```diff
@@ -1,129 +1,244 @@
...
+        case mission_state_t::standby:             dispatch_standby();             break;
+        case mission_state_t::leader:              dispatch_leader();              break;
+        case mission_state_t::docking:             dispatch_docking();             break;
+        case mission_state_t::docked:              dispatch_docked();              break;
+        case mission_state_t::safehold:            dispatch_safehold();            break;
+        case mission_state_t::manual:              dispatch_manual();              break;
         default:
             printf(debug_severity::error, "Master state not defined: %d\n", static_cast<unsigned int>(mode));
-            mission_mode_f.set(static_cast<unsigned int>(mission_mode_t::safehold));
+            set(mission_state_t::safehold);
```
shihaocao: This should be a full 
`transition_to_state(mission_state_t::safehold,
                adcs_state_t::startup,
                prop_mode_t::disabled);`
block right?

src/FCCode/MissionManager.cpp
```diff
@@ -1,129 +1,244 @@
...
+    const float momentum = lin::norm(adcs_ang_momentum_fp->get());
+    const float threshold = rwa::max_speed_read * rwa::moment_of_inertia * detumble_safety_factor;
+    if (momentum <= threshold)
     {
-        adcs_cmd_attitude_fp->set({2,2,2,2}); // TODO fix to a good value
-        adcs_state_f.set(static_cast<unsigned int>(adcs_state_t::point_standby));
-        mission_mode_f.set(static_cast<unsigned int>(mission_mode_t::standby));
+        transition_to_state(mission_state_t::standby,
+            adcs_state_t::point_standby,
+            prop_mode_t::active);
```
shihaocao: why is prop_mode active in standby?

src/FCCode/MissionManager.cpp
```diff
@@ -1,129 +1,244 @@
...
+    // Stay in this state until ground commands fix the satellite and
+    // command the satellite out of this state.
 }
 
 void MissionManager::dispatch_follower() {
-    
-    // TODO
-}
+    if (too_long_since_last_comms()) {
+        transition_to_state(mission_state_t::standby,
```
shihaocao: what happened to auto clearing sat_designations? Won't standby auto exit if we don't clear designations?

(same with the `dispatch_leader()` block)

src/FCCode/FieldCreatorTask.hpp
tanishqaggarwal: I think so.

src/FCCode/MissionManager.cpp
tanishqaggarwal: Yep thanks

src/FCCode/MissionManager.cpp
tanishqaggarwal: It allows for ground operators to manually command the satellite's propulsion systems

src/FCCode/MissionManager.cpp
tanishqaggarwal: Shit thanks for catching

src/FCCode/MissionManager.cpp
tanishqaggarwal: The better place for this is RTD. I'll update that when I have more time.

src/FCCode/MissionManager.cpp
```diff
@@ -1,129 +1,246 @@
...
 void MissionManager::dispatch_docking() {
-    // TODO
+    // Docking command is applied iff the docking configuration is
+    // not "docked".
+    if (!dock_config_fp->get()) {
+        docking_config_cmd_f.set(true);
+    }
+    else {
+        docking_config_cmd_f.set(false);
+    }
```
shihaocao: Issue:
If `docking_config_fp->get()` is true, this code will request a `cmd_f` of `false`. The DockingController will move the undocking config. This is not desirable right?

Fix:
remove the else block

tanishqaggarwal: Thanks for catching this. I think a better fix would be including the cmd flag in the if condition

src/FCCode/MissionManager.cpp
```diff
@@ -1,129 +1,246 @@
...
+    adcs_paired_fp = find_writable_field<bool>("adcs.paired", __FILE__, __LINE__);
+    adcs_ang_momentum_fp = find_internal_field<lin::Vector3f>("attitude_estimator.h_body", __FILE__, __LINE__);
+
     radio_mode_fp = find_internal_field<unsigned char>("radio.mode", __FILE__, __LINE__);
+    last_checkin_cycle_fp = find_internal_field<unsigned int>("radio.last_comms_ccno", __FILE__, __LINE__);
+
+    prop_mode_fp = find_readable_field<unsigned char>("prop.mode", __FILE__, __LINE__);
+
+    piksi_mode_fp = find_readable_field<unsigned char>("piksi.state", __FILE__, __LINE__);
+    propagated_baseline_pos_fp = find_readable_field<d_vector_t>("orbit.baseline_pos", __FILE__, __LINE__);
```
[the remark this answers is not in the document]
tanishqaggarwal: Given that this is already so widespread, I'll defer this to another issue

src/FCCode/AttitudeEstimator.cpp
```diff
@@ -46,16 +51,20 @@ void AttitudeEstimator::set_data(){
...
-    w_temp[0] = estimate.w_body(0);
-    w_temp[1] = estimate.w_body(1);
-    w_temp[2] = estimate.w_body(2);  
+    f_vector_t w_temp = { estimate.w_body(0), estimate.w_body(1), estimate.w_body(2) };
     w_body_f.set(w_temp);
+
+    lin::Vector3f wvec = {w_temp[0], w_temp[1], w_temp[2]};
+    lin::Vector3f result;
+    if (adcs_paired_f.get()) result = (gnc::constant::JB_docked_sats * wvec).eval();
+    else result = (gnc::constant::JB_single_sat * wvec).eval();
```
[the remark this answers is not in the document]
tanishqaggarwal: Thanks this is clean

src/FCCode/MissionManager.cpp
tanishqaggarwal: Nevermind, you're right

tanishqaggarwal: I just completely removed the if statement since the DockingController already handles this logic for us

src/FCCode/MissionManager.cpp
```diff
@@ -1,129 +1,238 @@
...
-    mission_mode_f.set(static_cast<unsigned int>(mission_mode_t::startup));
+    docked_fp = find_readable_field<bool>("docksys.docked", __FILE__, __LINE__);
+
+    // Initialize a bunch of variables
+    transition_to_state(mission_state_t::startup,
+        adcs_state_t::startup,
+        prop_mode_t::disabled); // "Starting" transition
     is_deployed_f.set(false);
+    deployment_wait_elapsed_f.set(0);
     sat_designation_f.set(0);
```
fatimayousuf: dock_config_f isn't initialized in the docking control task. shihao brought up that the satellite starts in the docking config. should dock_config_f be initialized to true here?

src/FCCode/MissionManager.cpp
```diff
@@ -1,129 +1,238 @@
...
 }
```
fatimayousuf: I agree that the if/else block wasn't needed. However, shouldn't the docking_config_cmd be initialized to true at startup? Otherwise there would be a null pointer exception when docking controller is first executed?

src/FCCode/MissionManager.cpp
shihaocao: Question/Problem:
If `too_long_since_last_comms()` is true, and we're < docking_trigger_dist, both blocks are called. However, line 160: `set(sat_designation_t::undecided);` is still executed even if we want to continue docking. I don't think this is intended.

Suggested fix:
Problably break out both blocks into an if-else just for clarity with the `if (distance_to_other_sat() < docking_trigger_dist) {` block first?

src/FCCode/MissionManager.cpp
tanishqaggarwal: Nice catch. There won't be a null pointer exception, but there's no default initialization so the value of the field will be undefined. Thanks!

src/FCCode/MissionManager.cpp
tanishqaggarwal: I think a better place for subsystem initializations is within the subsystem controller, so I'll add it there.

src/FCCode/MissionManager.cpp
tanishqaggarwal: Good point, we want to have clear, defined behavior. I'll fix this

src/FCCode/MissionManager.cpp
```diff
@@ -1,129 +1,237 @@
...
+void MissionManager::set(adcs_state_t state) {
+    adcs_state_f.set(static_cast<unsigned int>(state));
+}
+
+void MissionManager::set(prop_mode_t mode) {
+    prop_mode_fp->set(static_cast<unsigned int>(mode));
+}
+
+void MissionManager::set(radio_mode_t mode) {
+    radio_mode_fp->set(static_cast<unsigned int>(mode));
```
athena255: radio_mode_fp should be an unsigned char

src/FCCode/MissionManager.cpp
```diff
@@ -1,129 +1,237 @@
...
+void MissionManager::set(mission_state_t state) {
+    mission_state_f.set(static_cast<unsigned int>(state));
+}
+
+void MissionManager::set(adcs_state_t state) {
+    adcs_state_f.set(static_cast<unsigned int>(state));
+}
+
+void MissionManager::set(prop_mode_t mode) {
+    prop_mode_fp->set(static_cast<unsigned int>(mode));
```
athena255: prop_mode_fp should be unsigned char.  I'm not sure if it matters in a static_cast, but it might be safe to check that each of these set functions sets the state to type that they were declared.

tanishqaggarwal: I think doing that requires RTTI, which takes up a whole lot of runtime memory + code size. It doesn't matter a whole lot if it ends up being unsigned char or int; thanks for catching this though.

test/test_fsw_quake_manager/test_quake_manager.cpp
```diff
@@ -509,37 +506,7 @@ void test_mt_ready_set_after_sbdrb_success()
     TEST_ASSERT_EQUAL(tf.quake_manager->dbg_get_qct().get_MT_length(), tf.radio_mt_len_fp->get());
 }
 
-// Test mode-based behavior of Quake Manager
-void test_modes() {
```
athena255: Why do we delete this test?

test/test_fsw_quake_manager/test_quake_manager.cpp
```diff
@@ -568,20 +535,19 @@ int test_mission_manager() {
     RUN_TEST(test_update_mo_load_new_snap);
     RUN_TEST(test_valid_initialization); 
     RUN_TEST(test_mt_ready_set_after_sbdrb_success);
-    RUN_TEST(test_modes);
     return UNITY_END();
 }
 
 #ifdef DESKTOP
 int main() {
-    return test_mission_manager();
```
athena255: Since this is a unit test, shouldn't the tests be able to run independently?

test/test_fsw_quake_manager/test_quake_manager.cpp
tanishqaggarwal: I removed modes from QuakeManager. It only operates on states.

test/test_fsw_quake_manager/test_quake_manager.cpp
tanishqaggarwal: I don't think this affects the independent running of tests. I changed this line because the file is testing QuakeManager, not MIssionManager.

tanishqaggarwal: I don't think this affects the independent running of tests. I changed this line because the file is testing QuakeManager, not MIssionManager.

Issue #36, opened 2019-10-12
Naming Convention


tanishqaggarwal:
Not an issue but a documentation note.

We want to standardize naming conventions of the many different objects that'll be floating around. Here's what I've been using so far and what I want to propose we use from this point onward:
- The `_sr` suffix indicates an object is a serializer
- The `_f` suffix indicates that an object is a `StateField`
- The `_fp` suffix indicates that an object is a `shared_ptr` to a `StateField`

We should keep updating this page as we develop more conventions. @athena255

tanishqaggarwal:
Some class naming conventions (adapted from another company): the three kinds of `ControlTask`s we should have are `Manager`s, `Monitor`s, and `Control`s. Monitors just passively read data (either from sensors or from other State Fields) and dump the results of calculations into other state fields. `Control`s read state fields and cause changes to other state fields or actuators. `Manager`s might have a more overarching purpose of both monitoring and controlling.

These distinctions aren't super tidy, but are meant to be guidelines.

tanishqaggarwal:
Within a state machine: use `dispatch_<state name>` as the naming convention for running an event sequence associated with a particular state of the machine.

Enums should be branched off into their own separate files, with file ending `.enum`, so that they can be easily read by Python scripts for processing. At the top of the enum should be a comment indicating for which state field the enum encodes data. See `mission_mode_t.enum` as an example.

tanishqaggarwal:
Enums should be `enum class`es, so that the compiler can enforce some type-checking for values.

tanishqaggarwal:
If we need to create multiple state fields of a similar kind, use the `std::vector` class and use the `_fm` prefix to denote "multiple fields". Example:

`std::vector<WritableStateField<float>> sun_sensor_voltages_fm;`

tanishqaggarwal:
Fault state fields should be marked with the suffix `_fault_f` rather than simply `_f`.

tanishqaggarwal:
I'm putting this here: https://github.com/pathfinder-for-autonomous-navigation/FlightSoftware/wiki/Naming-Convention

PR #127, opened 2019-12-27
Prop controller

athena255:
Draft


src/FCCode/PropController.cpp
```diff
@@ -0,0 +1,227 @@
...
+bool PropController::done_filling()
+{
+    prop_time_t delta = now() - state_vars.inner_start_time;
+    return delta >= convert_sec(prop_constants.inner_open_duration);
+}
+
+void PropController::stop_filling()
+{
+    close_inner(0);
+    close_inner(1);
```
tanishqaggarwal: The open and close functions are provided in the driver

src/FCCode/PropController.cpp
```diff
@@ -0,0 +1,227 @@
...
+            printf(debug_severity::error, "Prop state not defined: %d", prop_state_f.get());
+    }
+}
+
+
+void PropController::dispatch_idle()
+{
+    prop_time_t delta = abs(now() - fire_time_f.get());
+    // Return if there are more than 5 minutes between now and when we have to fire
+    if ( delta > convert_min(prop_constants.minutes_until_pressurize) )
```
tanishqaggarwal: I think there's a timing problem here if you're immediately firing after pressurizing. Suppose the firing is 4 mins 59 seconds away, and pressurization ends up taking only 2 minutes. Then we'll fire ~3 minutes before we're supposed to.

src/FCCode/PropController.cpp
```diff
@@ -0,0 +1,227 @@
...
+    else // If inner valves are closed
+    {
+        if ( done_pressurizing() )
+        {
+            prop_state_f.set(prop_state_t::firing);
+            // TODO: enable interrupts
+        }
+        else if ( have_fills_left() ) 
+        {
+            start_filling();
```
tanishqaggarwal: Is this logic supposed to include waiting between fills yet? If it does, I couldn't find it

src/FCCode/PropController.cpp
```diff
@@ -0,0 +1,227 @@
...
+void PropController::start_filling()
+{
+    prop_time_t delta = now() - state_vars.inner_start_time;
+    // Return if we have fired outer valves within the last 10 seconds
+    if (delta < convert_sec(prop_constants.wait_time_between_fills))
+        return;
+
+    ++state_vars.num_fills;
+    state_vars.inner_start_time = open_inner(0);
+    open_inner (1);
```
tanishqaggarwal: I think we should only open one intertank valve at a time, not both. This allows one valve to serve as a backup. The choice of valve should be settable via a state field.

src/FCCode/PropController.h
```diff
@@ -0,0 +1,221 @@
...
+        // start pressurizing when we are within 5 min of fire_time
+        prop_time_t minutes_until_pressurize = 5;
+        // do not open inner valves more than 20 times during pressurizing state
+        prop_time_t max_num_fills = 20;
+        // wait 10 seconds between opening inner valves
+        prop_time_t wait_time_between_fills = 10;
+        // open inner valves for 1s
+        prop_time_t inner_open_duration = 1; 
+        // open inner valves for 2s if one of the inner valves stuck close
+        prop_time_t inner_open_duration_bak = 2;
```
tanishqaggarwal: This second constant is a great idea, for simplicity though I think we can just have one constant. I don't think the opening time matters too much because 20 fills is already a huge number of fills. Also, the power constraints only allow us to hold a valve open for ~1 s I think. @stewartaslan could you confirm?

src/FCCode/PropController.h
```diff
@@ -0,0 +1,221 @@
...
+    /**
+     * Constants
+     */
+    static const struct _prop_constants {
+        // start pressurizing when we are within 5 min of fire_time
+        prop_time_t minutes_until_pressurize = 5;
+        // do not open inner valves more than 20 times during pressurizing state
+        prop_time_t max_num_fills = 20;
+        // wait 10 seconds between opening inner valves
+        prop_time_t wait_time_between_fills = 10;
```
tanishqaggarwal: This should probably be ~30 s. @stewartaslan can you number check

src/FCCode/PropController.h
```diff
@@ -0,0 +1,221 @@
...
+    struct _state_vars{
+        bool firing = 0;
+        bool done_firing = 0;
+        bool outers_done[4] = {0, 0, 0, 0};
+        prop_time_t outer_start_times[4] = {0, 0, 0, 0};
+        prop_time_t inner_start_time = 0;
+        bool inner_open = 0;
+        size_t num_fills = 0;
+
+    } state_vars;
```
tanishqaggarwal: where do you instantiate the state_vars? I couldn't find it

src/FCCode/PropController.cpp
```diff
@@ -0,0 +1,227 @@
...
+{ 
+    add_internal_field(prop_state_f);
+    add_internal_field(prop_mode_f);
+    add_internal_field(prop_fault_f);
+    add_internal_field(fire_time_f);
+    add_internal_field(delta_thrust_times_f);
+
+    prop_state_f.set(prop_state_t::idle);
+    prop_mode_f.set(prop_mode_t::disabled);
+    prop_fault_f.set(prop_fault_t::nofaults);
```
tanishqaggarwal: can you name this `prop_fault_cause_f` to distinguish from a fault bit (see Naming Conventions)

src/FCCode/PropController.cpp
```diff
@@ -0,0 +1,227 @@
+#include "PropController.h"
+
+PropController::PropController(StateFieldRegistry &registry, unsigned int offset) : 
+    TimedControlTask<void>(registry, "prop", offset),
+    prop_state_f("prop.state"),
+    prop_mode_f("prop.mode"),
+    prop_fault_f("prop.fault"),
+    fire_time_f("prop.fire_time"),
+    delta_thrust_times_f("prop.thrust_vector")
```
tanishqaggarwal: Since these will come in from an external source, consider making them state field pointers instead of state fields. Unless these are the snapshotted copy?

src/FCCode/PropController.cpp
```diff
@@ -0,0 +1,227 @@
...
+        if ( !state_vars.outers_done[i] )
+        {
+            if ( abs(now() - state_vars.outer_start_times[i]) >= thrust_vector[i] )
+            {
+                close_outer(i);
+                state_vars.outers_done[i] = true;
+            }
+        }
+    }
+}
```
tanishqaggarwal: Looks good!

src/FCCode/PropController.cpp
tanishqaggarwal: #36

src/FCCode/PropController.cpp
```diff
@@ -0,0 +1,227 @@
...
+        else 
+        {
+            prop_fault_f.set(prop_fault_t::cannot_refill);
+        }
+    }
+}
+
+bool PropController::done_filling()
+{
+    prop_time_t delta = now() - state_vars.inner_start_time;
```
tanishqaggarwal: I think done_filling should also check for the actual tank pressure exceeding the "fill threshold"

tanishqaggarwal:
Closing due to review.

PR #130, opened 2019-12-30
Dev/gomspace controller

fatimayousuf:


src/FCCode/GomspaceController.cpp
```diff
@@ -60,7 +60,38 @@ GomspaceController::GomspaceController(StateFieldRegistry &registry, unsigned in
...
+    counter_reset_cmd_f("gomspace.counter_reset_cmd", counter_reset_cmd_sr),
+
+    wdt_reset_cmd_sr(),
+    wdt_reset_cmd_f("gomspace.wdt_reset_cmd", wdt_reset_cmd_sr),
+
+    gs_reset_cmd_sr(),
+    gs_reset_cmd_f("gomspace.gs_reset_cmd", gs_reset_cmd_sr),
+
+    gs_reboot_cmd_sr(),
+    gs_reboot_cmd_f("gomspace.gs_reboot_cmd", gs_reboot_cmd_sr)
```
tanishqaggarwal: What's the difference between the  reset and reboot commands? I forget

src/FCCode/GomspaceController.cpp
```diff
@@ -60,7 +60,38 @@ GomspaceController::GomspaceController(StateFieldRegistry &registry, unsigned in
...
+    ppt_mode_cmd_sr(1,2),
+    ppt_mode_cmd_f("gomspace.pptmode_cmd", ppt_mode_cmd_sr),
+
+    heater_cmd_sr(),
+    heater_cmd_f("gomspace.heater_cmd", heater_cmd_sr),
+
+    counter_reset_cmd_sr(),
+    counter_reset_cmd_f("gomspace.counter_reset_cmd", counter_reset_cmd_sr),
+
+    wdt_reset_cmd_sr(),
```
tanishqaggarwal: We don't need this control because we automatically reset the WDT timer every time we talk to the Gomspace.

src/FCCode/GomspaceController.cpp
```diff
@@ -60,7 +60,38 @@ GomspaceController::GomspaceController(StateFieldRegistry &registry, unsigned in
...
+    output1_cmd_f("gomspace.output1_cmd", output_cmd_sr),
+    output2_cmd_f("gomspace.output2_cmd", output_cmd_sr),
+    output3_cmd_f("gomspace.output3_cmd", output_cmd_sr),
+    output4_cmd_f("gomspace.output4_cmd", output_cmd_sr),
+    output5_cmd_f("gomspace.output5_cmd", output_cmd_sr),
+
+    pv_output_cmd_sr(0,4000,9),
+    pv1_output_cmd_f("gomspace.pv1_cmd", pv_output_cmd_sr),
+    pv2_output_cmd_f("gomspace.pv2_cmd", pv_output_cmd_sr),
+    pv3_output_cmd_f("gomspace.pv3_cmd", pv_output_cmd_sr),
```
tanishqaggarwal: This is for setting the boost converter voltages, right?

test/test_fsw_gomspace/test_fsw_gomspace.cpp
```diff
@@ -180,10 +225,36 @@ void test_task_initialization() {
...
+
+    TEST_ASSERT_NOT_NULL(tf.heater_cmd_fp);
+
+    TEST_ASSERT_NOT_NULL(tf.counter_reset_cmd_fp);
+
+    TEST_ASSERT_NOT_NULL(tf.wdt_reset_cmd_fp);
+
+    TEST_ASSERT_NOT_NULL(tf.gs_reset_cmd_fp);
+
+    TEST_ASSERT_NOT_NULL(tf.gs_reboot_cmd_fp);
```
tanishqaggarwal: These assertions are actually automatically done now by the find_readable_field_t functions of the StateFieldRegistryMock.

src/FCCode/GomspaceController.cpp
```diff
@@ -106,11 +137,39 @@ GomspaceController::GomspaceController(StateFieldRegistry &registry, unsigned in
...
+
+        add_writable_field(gs_reboot_cmd_f);
      }
 
 void GomspaceController::execute() {
     //get hk data from struct in driver
     assert(gs.get_hk());
+
+    // Set the gomspace outputs to the values of the statefield commands every 30 seconds
+    if (control_cycle_count%300==0){
```
tanishqaggarwal: Can you break out 300 to be a named constant inside the header file?

src/FCCode/GomspaceController.cpp
```diff
@@ -157,3 +216,56 @@ void GomspaceController::execute() {
...
+    }
+    if (output3_f.get()!=output3_cmd_f.get()){
+        gs.set_single_output(3, output3_cmd_f.get());
+    }
+    if (output4_f.get()!=output4_cmd_f.get()){
+        gs.set_single_output(4, output4_cmd_f.get());
+    }
+    if (output5_f.get()!=output5_cmd_f.get()){
+        gs.set_single_output(5, output5_cmd_f.get());
+    }
```
tanishqaggarwal: This kind of logic makes a power cycle a very manual process for a controlling subsystem (since it has to request an "off" and an "on" manually.) Can you perhaps implement some higher-level logic that automatically handles the power cycling within the Gomspace?

Also, when an output turns off, there's a low-pass filter applied so that it very slowly transitions from going off to on. This means that the output feedback from the gs.get_hk() is not an accurate representation of the digital state of the output. I don't think this affects your logic here, but I wanted you to be aware that this behavior will exist when you're hardware-testing the code.

src/FCCode/GomspaceController.cpp
tanishqaggarwal: I think you should avoid setting outputs on the first control cycle, because the values of the HK values are undefined and the values of the commands are uninitialized. Relatedly, on the first control cycle we should probably _set_ the values of the commands to be equal to the actual HK values, in order to prevent unwanted, weird writes from occurring.

test/test_fsw_gomspace/test_fsw_gomspace.cpp
```diff
@@ -231,6 +302,111 @@ void test_task_execute() {
...
+    tf.ppt_mode_cmd_fp->set(1); // 0 is hardware default; 1 is maximum power point tracking
+    tf.heater_cmd_fp->set(true);
+
+    tf.counter_reset_cmd_fp->set(false);
+    tf.wdt_reset_cmd_fp->set(false);
+    tf.gs_reset_cmd_fp->set(false);
+    tf.gs_reboot_cmd_fp->set(false);
+
+    // Let 30 seconds or 300 control cycles pass
+    TimedControlTaskBase::control_cycle_count=3000;
```
tanishqaggarwal: I think you meant to say 300 here

test/test_fsw_gomspace/test_fsw_gomspace.cpp
```diff
@@ -231,6 +302,111 @@ void test_task_execute() {
...
+    tf.output2_cmd_fp->set(true);
+    tf.output3_cmd_fp->set(true);
+    tf.output4_cmd_fp->set(true);
+    tf.output5_cmd_fp->set(true);
+
+    tf.pv1_output_cmd_fp->set(1000);
+    tf.pv2_output_cmd_fp->set(2000);
+    tf.pv3_output_cmd_fp->set(3000);
+
+    tf.ppt_mode_cmd_fp->set(1); // 0 is hardware default; 1 is maximum power point tracking
```
tanishqaggarwal: Fyi, I think our hardware default is MPPT. This doesn't affect your test though

test/test_fsw_gomspace/test_fsw_gomspace.cpp
```diff
@@ -231,6 +302,111 @@ void test_task_execute() {
...
+
+    tf.counter_reset_cmd_fp->set(false);
+    tf.wdt_reset_cmd_fp->set(false);
+    tf.gs_reset_cmd_fp->set(false);
+    tf.gs_reboot_cmd_fp->set(false);
+
+    // Let 30 seconds or 300 control cycles pass
+    TimedControlTaskBase::control_cycle_count=3000;
+
+    #ifndef DESKTOP
```
tanishqaggarwal: Why do you need this include guard?

test/test_fsw_gomspace/test_fsw_gomspace.cpp
```diff
@@ -231,6 +302,111 @@ void test_task_execute() {
...
+    tf.counter_reset_cmd_fp->set(true);
+    tf.gs_controller->execute();
+
+    // The boot and WDT counters should reset
+    TEST_ASSERT_EQUAL(boot_count+1, tf.counter_boot_fp->get());
+    TEST_ASSERT_EQUAL(wdt_i2c_count+1, tf.counter_wdt_i2c_fp->get());
+    TEST_ASSERT_EQUAL(wdt_gnd_count+1, tf.counter_wdt_gnd_fp->get());
+    TEST_ASSERT_EQUAL(wdt_csp_count1+1, tf.counter_wdt_csp1_fp->get());
+    TEST_ASSERT_EQUAL(wdt_csp_count2+1, tf.counter_wdt_csp2_fp->get());
+    TEST_ASSERT_EQUAL(false, tf.counter_reset_cmd_fp->get());
```
tanishqaggarwal: You say the counters should reset, but aren't they still incrementing here? I'm confused.

src/FCCode/GomspaceController.cpp
fatimayousuf: yes

src/FCCode/GomspaceController.cpp
fatimayousuf: I believe a reboot turns the power off and then on. A reset would restore the gomspace to the factory default settings (i.e the PPT mode would be MPPT)

test/test_fsw_gomspace/test_fsw_gomspace.cpp
fatimayousuf: sorry i must have misread the gomspace manual

test/test_fsw_gomspace/test_fsw_gomspace.cpp
fatimayousuf: The test file wouldn't pass the desktop test given the existing gomspace driver. For example, in Gomspace.cpp, get_heater() always returns 0 on the desktop test. should i add to the gomspace driver or leave the include guard?

src/FCCode/GomspaceController.cpp
tanishqaggarwal: Interesting, thanks.

test/test_fsw_gomspace/test_fsw_gomspace.cpp
tanishqaggarwal: I think it'd be better to still have all the data wiring get tested in desktop contexts. We want to push platform-specific complexity down towards lower levels (i.e. drivers) as much as possible.

PR #132, opened 2020-01-02
Fixes to release and Docker scripts

tanishqaggarwal:


tanishqaggarwal:
The docker script is only for local development use; the flight and ground unit tests are the only tests that could succeed on our local OS but might fail remotely.

PR #137, opened 2020-01-16
HAVT Update Pio Unit Test

shihaocao:
The main addition is the PIO Unit Test for the HAVT Update sister PR.

Sister PR:
https://github.com/pathfinder-for-autonomous-navigation/ADCSSoftware/pull/26


test/test_adcs_havt/test_adcs_havt.cpp
```diff
@@ -0,0 +1,110 @@
...
+#include <string>
+#include <unity.h>
+
+// tests have passed where this is 10, try increment higher if not passing
+constexpr static unsigned int wait_for_ADCSC = 10;
+
+// personal sanity check
+void test_trivial(){
+    TEST_ASSERT_TRUE(true);
+}
```
tanishqaggarwal: Don't need this anymore

test/test_adcs_havt/test_adcs_havt.cpp
```diff
@@ -0,0 +1,110 @@
...
+    adcs.get_who_am_i(&temp);
+    
+    TEST_ASSERT_EQUAL(Devices::ADCS::WHO_AM_I_EXPECTED, temp);
+}
+
+// initial test of read capability, will fail if not a clean boot
+// ex if there was a previous set_havt command that failed
+void test_read_table(){
+    std::bitset<havt::max_devices> read_table_read(0);
+    std::bitset<havt::max_devices> rt_expected("00000000000000000000001110111000");
```
tanishqaggarwal: Why is this the expected bootup string?

shihaocao: Counting from RIGHT to LEFT, this string indicates that all MTR's and WHEEL's are functional(). That's because the only way they become disabled is through a ground command. All other devices auto-disable since nothing is connected to the ADCSC in just the ADCSC + FC HITL setup.

MTRs are indices 3,4,5
WHEELs are indices 7,8,9

PR #141, opened 2020-01-17
ADCS Controller

shihaocao:
This PR contains the ADCSBoxController Control Task. Its main purpose is to use the ADCS driver and execute low-level hardware calls. Its inputs mostly come from the yet to be created ADCSCommander.

It also applies the havt_cmd table, and restarts SSA vector calculation.

### DIFF
--- a/src/FCCode/ADCSBoxController.hpp
+++ b/src/FCCode/ADCSBoxController.hpp
@@ -79,7 +79,7 @@ protected:
      * @brief HAVT command table, a vector of pointers to bool state fields
      * 
      */
-    std::vector<WritableStateField<bool>*> havt_cmd_table_vector_fp;
+    std::vector<const WritableStateField<bool>*> havt_cmd_table_vector_fp;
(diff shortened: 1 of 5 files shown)


src/FCCode/ADCSBoxController.hpp
```diff
@@ -0,0 +1,86 @@
...
+    const WritableStateField<float>* imu_gyr_temp_kp_fp;
+    const WritableStateField<float>* imu_gyr_temp_ki_fp;
+    const WritableStateField<float>* imu_gyr_temp_kd_fp;
+    const WritableStateField<float>* imu_gyr_temp_desired_fp;
+
+    /**
+     * @brief HAVT command table, a vector of pointers to bool state fields
+     * 
+     */
+    std::vector<WritableStateField<bool>*> havt_cmd_table_vector_fp;
```
tanishqaggarwal: The items can be const pointers

PR #155, opened 2020-01-19
Event storage

tanishqaggarwal:
This implements an "event log", i.e. multiple fields for storing the same kind of event. The idea behind this is that such a storage mechanism automatically stores a new event in the next available "slot" for storing an event, wrapping around stale data if necessary.


src/fsw/FCCode/Event.cpp
```diff
@@ -19,6 +19,12 @@ Event::Event(const std::string& name,
     }
 }
 
+Event::Event(Event&& other) :
+    StateField<bool>(other.name(), true, false),
```
kkrol27: I'm assuming here that `other.name()` is a `std::string const &`? If so, everything looks good.

tanishqaggarwal: Yep

src/fsw/FCCode/UplinkCommon.cpp
```diff
@@ -59,7 +59,7 @@ bool Uplink::_validate_packet(bitstream& bs)
     // if padding was not 0
     return (u8 != 0 || bits_consumed > 7) ? false : true;
 }
-#include <iostream>
```
tanishqaggarwal: We should be extremely careful about leaving iostream as an include since it bloats up the code size by about 160 KB (15% of total memory.)

src/fsw/FCCode/EventStorage.cpp
```diff
@@ -0,0 +1,44 @@
...
+}
+
+void EventStorage::add_event_to_registry(StateFieldRegistry& registry) {
+    for(Event& e : sub_events) {
+        registry.add_readable_field(static_cast<ReadableStateFieldBase*>(&e));
+    }
+}
+
+size_t EventStorage::bitsize() const {
+    return sub_events[event_ptr].bitsize();
```
athena255: Is sub_events[event_ptr] always guaranteed to have a bit_array?

test/test_fsw_event/test_fsw_event.cpp
```diff
@@ -32,18 +33,17 @@ struct TestFixture {
...
 
     // Set values for test
     tf.control_cycle_count = 20;
     tf.data1_f.set(true);
     tf.data2_f.set(false);
-    
+
     // Verify that upon serialization, the values are written into the event's bitset in the way
     // that we would expect
     tf.event.signal();
```
athena255: Why do we copy ba to ccno instead of using the bit_array to ulong function?

src/fsw/FCCode/EventStorage.hpp
```diff
@@ -0,0 +1,30 @@
+#ifndef EVENT_STORAGE_HPP_
+#define EVENT_STORAGE_HPP_
+
```
athena255: What is EventStorage suppose to do?

src/fsw/FCCode/Event.cpp
```diff
@@ -19,6 +19,12 @@ Event::Event(const std::string& name,
...
+    data_fields(other.data_fields),
+    field_data(std::move(other.field_data)),
+    ccno(other.ccno) {}
+
```
athena255: in serialize(), is field_data guaranteed to have more capacity than 32 + field->bitsize()?

test/test_fsw_event/test_fsw_event.cpp
```diff
@@ -65,10 +65,46 @@ void test_event() {
...
+    test_single_event(tf);
+    test_single_event(tf);
+    test_single_event(tf);
+}
+
 #ifdef DESKTOP
 int main() {
     UNITY_BEGIN();
     RUN_TEST(test_event);
+    RUN_TEST(test_event_storage);
```
athena255: I think it might be useful to test boundary cases. For example, test that upper limits of event storage fail elegantly when they should, and/or test with adding more than one event to registry.

src/fsw/FCCode/EventStorage.cpp
tanishqaggarwal: I think so, since they're all actual event objects.

src/fsw/FCCode/EventStorage.cpp
```diff
@@ -0,0 +1,44 @@
...
+                 const unsigned int& _ccno)
+{
+    assert(storage_size < 100); // So that the suffixed event count doesn't have more than 2 digits
+    sub_events.reserve(storage_size);
+    for(unsigned char i = 1; i <= storage_size; i++) {
+        char x[4];
+        memset(x, 0, 4);
+        x[0] = '.';
+        sprintf(x, ".%d", i);
+        sub_events.emplace_back(name + std::string(x),
```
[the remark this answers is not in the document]
tanishqaggarwal: Why should we do that?

src/fsw/FCCode/EventStorage.cpp
```diff
@@ -0,0 +1,44 @@
...
+EventStorage::EventStorage(const std::string& name,
+                 const unsigned int storage_size,
+                 std::vector<ReadableStateFieldBase*>& _data_fields,
+                 const char* (*_print_fn)(const unsigned int, std::vector<ReadableStateFieldBase*>&),
+                 const unsigned int& _ccno)
+{
+    assert(storage_size < 100); // So that the suffixed event count doesn't have more than 2 digits
+    sub_events.reserve(storage_size);
+    for(unsigned char i = 1; i <= storage_size; i++) {
+        char x[4];
```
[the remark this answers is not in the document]
tanishqaggarwal: I think it's ok, it's a very small amount of memory. It's all stack-allocated anyway.

src/fsw/FCCode/EventStorage.cpp
```diff
@@ -0,0 +1,44 @@
...
+
+EventStorage::EventStorage(const std::string& name,
+                 const unsigned int storage_size,
+                 std::vector<ReadableStateFieldBase*>& _data_fields,
+                 const char* (*_print_fn)(const unsigned int, std::vector<ReadableStateFieldBase*>&),
+                 const unsigned int& _ccno)
+{
+    assert(storage_size < 100); // So that the suffixed event count doesn't have more than 2 digits
+    sub_events.reserve(storage_size);
+    for(unsigned char i = 1; i <= storage_size; i++) {
```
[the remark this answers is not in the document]
tanishqaggarwal: Why? We only need `storage_size` spaces for the events.

src/fsw/FCCode/EventStorage.cpp
```diff
@@ -0,0 +1,44 @@
...
+        memset(x, 0, 4);
+        x[0] = '.';
+        sprintf(x, ".%d", i);
+        sub_events.emplace_back(name + std::string(x),
+            _data_fields, _print_fn, _ccno);
+    }
+}
+
+void EventStorage::add_event_to_registry(StateFieldRegistry& registry) {
+    for(Event& e : sub_events) {
```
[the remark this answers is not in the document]
tanishqaggarwal: We could do that.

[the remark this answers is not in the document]
tanishqaggarwal: The pointers added to the registry are not const so we can't use const references.

test/test_fsw_event/test_fsw_event.cpp
tanishqaggarwal: I could do that. Thanks!

src/fsw/FCCode/EventStorage.hpp
tanishqaggarwal: It's essentially a storage for an event log. It points to the next "free" event to store an event in.

src/fsw/FCCode/Event.cpp
tanishqaggarwal: Yep because the constructor of Event() guarantees this

tanishqaggarwal:
Closing while I make significant changes.

PR #157, opened 2020-01-20
Send telemetry over USB

tanishqaggarwal:
This adds the ability for sending telemetry over USB.


src/fsw/FCCode/QuakeManager.h
```diff
@@ -91,7 +91,12 @@ class QuakeManager : public TimedControlTask<bool> {
     * The last cycle for which we had comms
     */
    InternalStateField<unsigned int> last_checkin_cycle_f;
-  
+
+   /**
+    * @brief This flag can be used by the sim to dump telemetry over the USB line.
+    */
+   WritableStateField<bool> dump_telemetry_f;
```
kkrol27: Do we want to wrap this in an `#ifdef FUNCTIONAL_TEST` as well?

tanishqaggarwal: Nah I think it’s more trouble than it’s worth to conditionally compile the existence of a state field. It would be better to just conditionally compile the _behavior_.

PR #158, opened 2020-01-21
Fully Integrated ADCS Software

kkrol27:
All of the ADCS build targets are now fully integrated into the FSW source tree and prefixed FSW targets with `fsw_`. Everything with the `*.ini` files went smoothly except for following two points:

 * Multiple inheritance (with the extends keyword) seems to work weird with default arguments so I suggest we try and stay away from it.
 * For some reason the `lib_ldf_mode` option didn't seem to propegate between files even though everything else did... I'm very skeptical if this is really what was going on but figured I'd get these changes in ASAP so the reorganization can ramp down sharply.

@tanishqaggarwal Everything should be set to rework some of the FSW `*.ini` files whenever!

@shihaocao I didn't implement a HAVT testing target atm because I wanted to ask you about your configuration really quick.


tanishqaggarwal:
Can the ADCSSoftware folder be deleted?

kkrol27:
> Can the ADCSSoftware folder be deleted?

It should have been removed by this PR - at least I think I deleted it.

Issue #154, opened 2020-01-19
DCDC Controller


tanishqaggarwal:
We need some code to manage DCDC enable pins

tanishqaggarwal:
Being implemented in #201

PR #159, opened 2020-01-21
Prop controller

athena255:


lib/fsw/Drivers/PropulsionSystem.cpp
```diff
@@ -11,77 +14,94 @@
...
-#ifdef DESKTOP
-// TODO
-#else
-    thrust_valve_loop_timer.begin(thrust_valve_loop, thrust_valve_loop_interval_us);
-#endif
+bool Tank::is_valve_open(size_t valve_idx)
+{
+    if (valve_idx >= num_valves)
+        return false;
+    return digitalRead(valve_pins[valve_idx]);
```
stewartaslan: Maybe I'm missing something here, but can we do a digital read on the valve pins since they're already set to be analog outputs?

lib/fsw/Drivers/PropulsionSystem.cpp
```diff
@@ -93,77 +113,172 @@ float PropulsionSystem::get_pressure() {
...
+    tank2.close_all_valves();
+    is_enabled = false;
+
+    // Release tank2 lock because tank2 did not get to fire
+    if (tank2.valve_lock.get_end_time() > micros())
+        tank2.valve_lock.end_time = 0;
+}
+
+bool PropulsionSystem::is_functional() { 
     return digitalRead(DCDC::dcdc_sph_enable_pin);
```
stewartaslan: This is just checking if the DCDC enable pin is high, which doesn't tell us whether prop is functional or whether DCDC is functional, it just tells us if we have tried to enable the DCDC. I don't think we really have a way of telling if the prop system is functional other than looking at fault flags that could be thrown because of pressure sensor, temp sensors, or valves.

lib/fsw/Drivers/PropulsionSystem.cpp
```diff
@@ -11,77 +14,94 @@
...
+{
+    for (size_t i = 0; i < num_valves; ++i)
         pinMode(valve_pins[i], OUTPUT);
-    }
 
-    pinMode(pressure_sensor_low_pin, INPUT);
-    pinMode(pressure_sensor_high_pin, INPUT);
-    pinMode(temp_sensor_inner_pin, INPUT);
-    pinMode(temp_sensor_outer_pin, INPUT);
+    pinMode(temp_sensor_pin, INPUT);
```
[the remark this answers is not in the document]
athena255: From what I understand, tank1 has a 2 pins for the two valves and a temperature sensor pin. Tank2 has 4 mins for the four valves, a temperature sensor pin, and a pressure sensor pin.

lib/fsw/Drivers/PropulsionSystem.cpp
```diff
@@ -11,77 +14,94 @@
...
-#else
-    for (unsigned char i = 0; i < 4; i++) thrust_valve_schedule[i] = 0;
-    for (unsigned char i = 0; i < 6; i++) digitalWrite(valve_pins[i], LOW);
-    thrust_valve_loop_timer.end();
-#endif
+/* Tank implementation */
+
+int Tank::get_temp()
+{
+    return analogRead(temp_sensor_pin);
```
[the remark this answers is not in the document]
athena255: The temperature sensor driver seems like a good idea. I don't know the conversion for the temperature. If we plan to use the temperature driver, should I just make this method empty for now?

lib/fsw/Drivers/PropulsionSystem.cpp
```diff
@@ -93,77 +113,172 @@ float PropulsionSystem::get_pressure() {
     return pressure;
```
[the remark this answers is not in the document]
athena255: We could use the pressure sensor driver instead. The original code in this document had the equation there, so I just left it since I don't know how to convert pressure. Should I leave this method blank?

lib/fsw/Drivers/PropulsionSystem.cpp
```diff
@@ -93,77 +113,172 @@ float PropulsionSystem::get_pressure() {
...
+    // Unlock 2.9 ms from start time so as to not be one interrupt-cycle (3 ms) late
+    if (tank2.valve_lock.procure(tank2.start_time - micros() - 100))
+    {
+        tank2.thrust_valve_loop_timer.begin(thrust_valve_loop, tank2.thrust_valve_loop_interval_us);
+        is_enabled = true;
+    }
+    return is_enabled;
+}
+
+bool PropulsionSystem::is_start_time_ok(uint32_t start_time_us)
```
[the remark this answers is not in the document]
athena255: start_time is the time at which Tank2 should fire. Would it be better to call it fire_time?

lib/fsw/Drivers/PropulsionSystem.cpp
```diff
@@ -93,77 +113,172 @@ float PropulsionSystem::get_pressure() {
...
-            digitalWrite(valve_pins[i - 2], LOW);
-            #endif
-
-            thrust_valve_schedule[i - 2] = 0;
-            is_valve_opened[i] = false;
+    noInterrupts(); 
+    // Serial.printf("\nCurrent Time: %u\n",micros());
+    for (size_t i = 0; i < tank2.num_valves; ++i) {
+        // If a valve is scheduled to open for less than 3 ms, then ignore it
+        if (tank2.schedule[i] < tank2.thrust_valve_loop_interval_ms) {
```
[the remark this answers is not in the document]
athena255: Are you saying that if the valve is scheduled to open for less than 10 ms, then ignore it? If so, should i make that variable a WritableStateField?

lib/fsw/Drivers/PropulsionSystem.cpp
athena255: I thought that the valve pins were digital since they are only either open or closed.

lib/fsw/Drivers/PropulsionSystem.cpp
athena255: The original code defined is_functional like this. I'm not very sure what the dcdc_sph_enable_pin does. I can change this to look at all the fault flags though, but I have not written fault handling in the driver.

lib/fsw/Drivers/PropulsionSystem.cpp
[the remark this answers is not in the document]
stewartaslan: Agreed, so tank1 has the `temp_sensor_inner_pin` and tank2 has the `temp_sensor_outer` pin. What is `temp_sensor_pin`?

lib/fsw/Drivers/PropulsionSystem.cpp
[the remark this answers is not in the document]
stewartaslan: Yes I think so. The temperature sensor driver will essentially just read the analog voltage and then convert that to a temperature by a linear equation which we'll get from [this](https://cornellprod-my.sharepoint.com/:x:/r/personal/saa243_cornell_edu/_layouts/15/Doc.aspx?sourcedoc=%7B03403C8C-0D4E-4C55-9D01-5805E6D845CC%7D&file=SP102F1J%20REV%20A%20(R-T%20Table).xls&action=default&mobileredirect=true) table.

lib/fsw/Drivers/PropulsionSystem.cpp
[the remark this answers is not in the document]
stewartaslan: Yes, we have a pressure sensor driver somewhat written, so I think the conversion should continue to live in the pressure sensor driver rather than the prop driver.

lib/fsw/Drivers/PropulsionSystem.cpp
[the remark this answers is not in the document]
stewartaslan: In that case, shouldn't start_time be more than 10 minutes in the future since it could take up to 10 minutes to pressurize tank2?

lib/fsw/Drivers/PropulsionSystem.cpp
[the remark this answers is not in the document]
stewartaslan: Yes, valve opening times less than 10ms should be ignored. This doesn't need to be a WriteableStateField though, we will set this on the ground and we don't need to adjust it during the mission.

lib/fsw/Drivers/PropulsionSystem.cpp
stewartaslan: Ah ok yes, they are digital outputs. But in order to do a digitalRead, don't we first have to set the pinMode to INPUT? And to open the valves the pins need a digitalWrite which requires pinMode to be OUTPUT. So if the valve is on and then we switch to INPUT and do a digitalRead, the valve will turn off because the pin is no longer a digital output. So I don't think it will work to read a pin like this? I might just be not understanding how GPIO pins work though.

lib/fsw/Drivers/PropulsionSystem.cpp
stewartaslan: dcdc_sph_enable_pin just sets the DCDC enable pin high which supplies 24V to the valves. I think we should take this out and just have is_functional always return true for now since we don't have faults yet.

test/test_prop/prop_common.h
```diff
@@ -0,0 +1,21 @@
+
+#ifndef PROP_COMMON_H_
+#define PROP_COMMON_H_
+
+#include <unity.h>
+#include "../../lib/Drivers/PropulsionSystem.hpp"
```
kkrol27: Same include things here. We should just. be able to do `#include <PropulsionSystem.hpp>`.

lib/fsw/Drivers/PropulsionSystem.cpp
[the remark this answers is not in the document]
athena255: factored out common logic into a Tank class and then derived Tank 1 and Tank 2 from Tank. So since both Tank1 and Tank2 have a temp_sensor, it is represented as just temp_sensor in Tank.

lib/fsw/Drivers/PropulsionSystem.cpp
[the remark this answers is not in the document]
athena255: So my thinking was that 10 minutes is enforcable by the Prop controller since it is much greater than 140 ms. In this case, I only care that tank2 can fire in the time requested (assuming that it is already pressurized, but Prop controller should have made sure of that). 

Essentially, the driver does not know or care if tank2 is pressurized. It only opens and closes valves and enforces timing for opening and closing valves.

lib/fsw/Drivers/PropulsionSystem.cpp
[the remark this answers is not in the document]
athena255: Changing this to 10ms will also cut off the valve firing up to 10ms in advance. Is this the intended behavior? So if we asked it to fire for 25 ms, it will only be guaranteed to fire for 15 ms.

lib/fsw/Drivers/PropulsionSystem.cpp
athena255: I thought we could digitalRead on a pin set to OUTPUT because it's reading its own signal? I'm not sure either.

lib/fsw/Drivers/PropulsionSystem.hpp
```diff
@@ -1,115 +1,384 @@
...
- * sph.setup();
- * **/
+ * 
+ * It consists of a static Tank1 (inner tank) and Tank2 (thrust tank) data objects. 
+ * 
+ * Specifications:
+ * Tank1 valves are at numbered 0, 1
+ * Tank2 valves are at numbered 0, 1, 2, 3
+ * mandatory_wait_time is the duration of time after opening a valve before 
+ * we can open another valve. Tank1 is 3ms. Tank2 is 10*1000 ms
```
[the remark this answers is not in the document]
athena255: That is a typo. Tank1 wait time is 10 s and Tank2 wait time is 3ms

lib/fsw/Drivers/PropulsionSystem.cpp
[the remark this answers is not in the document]
stewartaslan: After talking to Tanishq, we'll leave the temperature and pressure methods in and we'll use these instead of having separate temperature and pressure sensor drivers

lib/fsw/Drivers/PropulsionSystem.cpp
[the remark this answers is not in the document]
stewartaslan: Same as the comment above, leave the methods in.

lib/fsw/Drivers/PropulsionSystem.cpp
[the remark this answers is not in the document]
stewartaslan: Ok, that makes sense.

lib/fsw/Drivers/PropulsionSystem.cpp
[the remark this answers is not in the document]
stewartaslan: No that is not the intended behavior. I think we need to define a new variable that is minimum firing time. The `thrust_valve_loop_interval_ms` variable should remain 3ms

lib/fsw/Drivers/PropulsionSystem.cpp
stewartaslan: Sounds like we can do that, so it should all be good.

configs/hardware_test_envs.ini
```diff
@@ -29,16 +29,30 @@ test_filter = test_docking
...
 board = teensy36
-extends = fsw_hardware_common
+extends = hardware_common
+build_flags = ${hardware_common.build_flags} -llibc -lc
 test_filter = test_prop
 
+[env:teensy35_test_propulsion_with_sensors]
+board = teensy35
+extends = hardware_common
+build_flags = ${hardware_common.build_flags} -llibc -lc
```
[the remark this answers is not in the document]
athena255: we don't need it

lib/fsw/Drivers/PropulsionSystem.cpp
```diff
@@ -93,77 +113,172 @@ float PropulsionSystem::get_pressure() {
...
 }
 
-bool PropulsionSystem::is_functional() { 
-#ifdef DESKTOP
+bool PropulsionSystem::enable()
+{
+    if (is_enabled)
+    {
+        disable(); // If this happens, then there's a big problem
+        return false;
```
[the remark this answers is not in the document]
athena255: Calling enable() when it is already enabled implies that we don't know that it is already enabled.

lib/fsw/Drivers/PropulsionSystem.cpp
```diff
@@ -93,77 +113,172 @@ float PropulsionSystem::get_pressure() {
...
+    return true;
+}
+
+bool PropulsionSystem::open_valve(Tank& tank, size_t valve_idx)
+{
+    if (valve_idx >= tank.num_valves || !tank.valve_lock.is_free()) 
+        return false;
+    noInterrupts();
+    {
+        if (tank.valve_lock.procure(tank.mandatory_wait_time*1000 - 100))
```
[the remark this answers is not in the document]
athena255: i will rename it to mandatory_wait_time_ms

lib/fsw/Drivers/PropulsionSystem.hpp
```diff
@@ -1,115 +1,384 @@
...
+ *  - tank1 10s mandatory wait time
+ * IntervalTimer (thrust_valve_loop_timer) enforces
+ *  - tank2 3ms mandatory wait time
+ *  - tank2 firing schedule
+ * 
+ * Only tank2 has a schedule since only tank2 uses the IntervalTimer to fire.
+ * 
+ * No mandatory wait time is enforced when opening valves from different tanks. For example,
+ * suppose we open a valve on tank1. Although we cannot open the other valve for another 10s,
+ * we can immediately open any valve on tank2.
```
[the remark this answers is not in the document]
athena255: It will be the prop controller's responsibility to not do something like that.

lib/fsw/Drivers/PropulsionSystem.hpp
```diff
@@ -1,115 +1,384 @@
...
- * execute_schedule().
- * Example:
+ * PropulsionSystem class defines the interface for the propulsion system, which 
+ * consists of an inner tank, Tank1, and the (outer) thrust tank, Tank2. 
  *
- * SpikeAndHold sph;
- * sph.setup();
- * **/
+ * 
+ * It consists of a static Tank1 (inner tank) and Tank2 (thrust tank) data objects.
```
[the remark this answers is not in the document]
athena255: because interrupt function must be static

lib/fsw/Drivers/PropulsionSystem.cpp
```diff
@@ -93,77 +113,172 @@ float PropulsionSystem::get_pressure() {
...
+
+void PropulsionSystem::disable() {
+    noInterrupts();
+    tank2.thrust_valve_loop_timer.end();
+    interrupts();
+
+    tank2.close_all_valves();
+    is_enabled = false;
+
+    // Release tank2 lock because tank2 did not get to fire
```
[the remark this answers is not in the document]
athena255: yes, this comment is correct, but I will reword it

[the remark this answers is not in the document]
athena255: It is correct now

configs/hardware_test_envs.ini
```diff
@@ -1,130 +1,115 @@
...
+[env:fsw_teensy35_test_propulsion]
+board = teensy35
+extends = fsw_hardware_common
+test_filter = test_prop
+
+[env:fsw_teensy36_test_propulsion]
+board = teensy36
+extends = fsw_hardware_common
+test_filter = test_prop
+
```
tanishqaggarwal: Why did you get rid of with_sensors?

src/common/propulsion_fault_t.enum
```diff
@@ -3,9 +3,9 @@
 // prop.fault
```
tanishqaggarwal: I couldn't comment on the psim update but for some reason the submodule revision is different from master. I would reset this.

src/fsw/FCCode/Drivers/PropulsionSystem.hpp
```diff
@@ -16,13 +16,18 @@
  *  - SpikeAndHold must be enabled in order to use this system
```
tanishqaggarwal: Can you explain why you're having to conditionally compile so many things in this driver?

src/fsw/FCCode/Drivers/PropulsionSystem.cpp
```diff
@@ -165,19 +183,22 @@ bool PropulsionSystem::is_start_time_ok(uint32_t start_time_us)
...
-    if (tank2.valve_lock.get_end_time() > micros())
+    // Release tank2 lock if tank2 had not started firing
+    if (tank2.start_time > micros())
         tank2.valve_lock.end_time = 0;
 }
 
 bool PropulsionSystem::is_functional() { 
-    return digitalRead(DCDC::dcdc_sph_enable_pin);
+    // TODO: change this later maybe
+    return true;
```
tanishqaggarwal: #154

configs/hardware_test_envs.ini
athena255: because it was broken

src/fsw/FCCode/Drivers/PropulsionSystem.hpp
athena255: because all the Arduino things cannot be compiled by a Desktop

src/common/propulsion_fault_t.enum
athena255: this is considered up to date with master

src/fsw/FCCode/Drivers/PropulsionSystem.cpp
```diff
@@ -1,87 +1,117 @@
...
 
-void PropulsionSystem::reset() {
-    disable();
+void Tank2::setup()
+{
+    Tank::setup();
 #ifndef DESKTOP
-    delay(10);
+    pinMode(pressure_sensor_high_pin, INPUT);
+    pinMode(pressure_sensor_high_pin, INPUT);
```
stewartaslan: Should one of these be `pressure_sensor_low_pin`?

src/fsw/FCCode/Drivers/PropulsionSystem.cpp
```diff
@@ -1,87 +1,117 @@
...
 #endif
-    enable();
 }
 
-float PropulsionSystem::get_pressure() {
+/* Tank implementation */
+
+int Tank::get_temp() const
+{
+    return analogRead(temp_sensor_pin);
```
stewartaslan: Add a TODO about converting the analogRead value to a temperature

src/fsw/FCCode/Drivers/PropulsionSystem.cpp
```diff
@@ -1,87 +1,117 @@
...
-const std::array<unsigned char, 6>
-PropulsionSystem::valve_pins = {27, 28, 3, 4, 5, 6};
+#ifdef DESKTOP
+static void interrupts(){}
+void noInterrupts(){}
+uint8_t analogRead(uint8_t pin){return pin%2;}
+uint8_t digitalRead(uint8_t pin){return pin%2;}
+void digitalWrite(uint8_t pin, uint8_t val){}
+#define LOW 0
+#define HIGH 1
```
stewartaslan: Why these definitions? `LOW` and `HIGH` are used elsewhere as-is

src/fsw/FCCode/Drivers/PropulsionSystem.cpp
```diff
@@ -1,87 +1,117 @@
...
+    for (size_t i = 0; i < num_valves; ++i)
+    {
+        digitalWrite(valve_pins[i], LOW);
+        is_valve_opened[i] = 0;
+    }
+}
+
+/* Tank2 implementation */
+
+float Tank2::get_pressure() const {
```
stewartaslan: Add TODO to get all the right values for these constants

src/fsw/FCCode/Drivers/PropulsionSystem.cpp
```diff
@@ -93,77 +123,177 @@ float PropulsionSystem::get_pressure() {
     return pressure;
 }
 
-signed int PropulsionSystem::get_temp_inner() {
```
stewartaslan: Change to `get_temp_tank1` to be consistent with the tanks naming

src/fsw/FCCode/Drivers/PropulsionSystem.cpp
```diff
@@ -93,77 +123,177 @@ float PropulsionSystem::get_pressure() {
...
-    return 0; // TODO replace with analogRead
-#endif
+unsigned int Tank2::get_schedule_at(size_t valve_num) const
+{
+    if (valve_num >= num_valves)
+        return 0;
+    return schedule[valve_num];
 }
 
-signed int PropulsionSystem::get_temp_outer() {
```
stewartaslan: Change to `get_temp_tank2` to be consistent with tank naming

src/fsw/FCCode/Drivers/PropulsionSystem.cpp
```diff
@@ -93,77 +123,177 @@ float PropulsionSystem::get_pressure() {
...
+
+bool PropulsionSystem::enable()
+{
+    if (is_enabled)
+    {
+        disable(); // If this happens, then there's a big problem
+        return false;
+    }
+
+    if ( !is_start_time_ok(tank2.start_time) )
```
stewartaslan: Add a comment about what start_time is

src/fsw/FCCode/Drivers/PropulsionSystem.cpp
```diff
@@ -93,77 +123,177 @@ float PropulsionSystem::get_pressure() {
...
-    enable();
+bool PropulsionSystem::is_functional() { 
+    // TODO: change this later maybe
+    return true;
+}
+
+bool PropulsionSystem::set_schedule(uint32_t a, uint32_t b, uint32_t c, uint32_t d, uint32_t start_time_us)
+{
+
+    // Do not set the schedule when enabled
```
stewartaslan: What does it mean to set the schedule?

src/fsw/FCCode/Drivers/PropulsionSystem.cpp
```diff
@@ -93,77 +123,177 @@ float PropulsionSystem::get_pressure() {
...
+bool PropulsionSystem::clear_schedule()
+{
+    if (is_enabled)
+        return false;
+    for (size_t i = 0; i < 4; ++i)
+        tank2.schedule[i] = 0;
+    return true;
+}
+
+bool PropulsionSystem::is_done_firing() const
```
stewartaslan: What is the purpose of `is_done_firing`?

src/common/propulsion_fault_t.enum
```diff
@@ -0,0 +1,11 @@
+#pragma once
+
+// prop.fault
+enum class prop_fault_t {
```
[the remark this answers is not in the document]
athena255: I have not done anything with faults yet. These were default values. I planned to handle faults in the controller and not the driver.

src/fsw/FCCode/Drivers/PropulsionSystem.cpp
athena255: yes, they should be.

src/fsw/FCCode/Drivers/PropulsionSystem.cpp
athena255: LOW and HIGH are defined in Arduino, which cannot be included in the Desktop environment. Rather than wrap ifndef DESKTOP everywhere, I just defined them.

src/fsw/FCCode/Drivers/PropulsionSystem.cpp
athena255: This line is no longer in the code. The red highlight indicates that it has been deleted.

src/fsw/FCCode/Drivers/PropulsionSystem.cpp
athena255: This line is no longer in the code. The red highlight indicates that it has been deleted.

src/fsw/FCCode/Drivers/PropulsionSystem.cpp
athena255: I documented this in PropulsionSystem.hpp, but I can copy that over if that is more clear. 
"scheduled (to fire)" means that tank2 currently has a scheduled start time in the future

src/fsw/FCCode/Drivers/PropulsionSystem.cpp
athena255: is_done_firing tells us that tank2 has finished firing. It checks that the schedule is set to 0. I have a typo in the header on line 61. It should be is_done_firing and not is_firing. 

I usually write API comments in the header as oppose to the implementation file, which I think should contain implementation comments.

src/fsw/FCCode/Drivers/PropulsionSystem.cpp
athena255: > Just a few changes requested in the comments, although I don't think I fully understand how TimedLock is functioning here. Overall I think in the future there could be more comments, but most of the functionality looks good.

Have you read line 207 in PropulsionSystem.hpp regarding TimedLock? A TimedLock is like a countdown timer on a bomb or an oven. It has no kill switch, and will only unlock when there is no more time. It is not implemented as a countdown timer, but it is functionally the same.

src/fsw/FCCode/Drivers/PropulsionSystem.cpp
tanishqaggarwal: This is a clever idea for unit testing. Although, I think it would be good to add a comment block explaining why these functions exist

src/fsw/FCCode/Drivers/PropulsionSystem.cpp
tanishqaggarwal: I feel like the better place for such a TODO is Recurring Constants Review

src/fsw/FCCode/Drivers/PropulsionSystem.cpp
stewartaslan: Oops, sorry about that

src/fsw/FCCode/Drivers/PropulsionSystem.cpp
stewartaslan: Yeah the documentation in PropulsionSystem.hpp is good, but I still am not sure what is meant by "set the schedule." Does that mean retrieve a new schedule from where it is stored by GNC software? Or does that mean change an existing schedule? Or something else?

src/fsw/FCCode/Drivers/PropulsionSystem.cpp
stewartaslan: Yes, PropulsionSystem.hpp answers my question, I didn't read that carefully enough.

PR #160, opened 2020-01-22
Fix FSW build structure

tanishqaggarwal:
This significantly changes environment names for Flight Software and consolidates the previous build changes that were made. All FSW environments are now prefixed with `fsw_` and end in `leader` or `follower` to indicate which of the two hardware satellites they're compiled for.


configs/flight_and_test_envs.ini
```diff
@@ -1,17 +1,18 @@
+[fsw_common]
+lib_extra_dirs = lib/fsw
+lib_deps =
+  CommonSoftware/
+  https://github.com/pathfinder-for-autonomous-navigation/psim
```
kkrol27: You can remove psim here. It's in the `lib/common` as a submodule!

platformio.ini
```diff
@@ -33,6 +33,7 @@ build_flags =
   -DDESKTOP
   -std=c++14
   -Wall
+  -lpthread
```
kkrol27: What is this used for exactly?

configs/flight_and_test_envs.ini
kkrol27: You may need to run `git submodule --init --update`.

platformio.ini
tanishqaggarwal: When flight software runs in desktop mode it runs the debugging stuff in a separate thread so that the control cycle doesn't get stuck due to std::cin.

kkrol27: Gotcha - we want that across all builds or just fsw internally? Doesn't matter a ton cuz adcs doesn't even use native builds.

PR #161, opened 2020-01-23
Fix Serializers

shihaocao:
- Changed unit tests to actually create a separate deserializer and test it
- Fixed vector and quaternion serializers


CommonSoftware/src/SerializerTypes.inl
```diff
@@ -380,16 +391,22 @@ class VectorSerializer : public SerializerBase<std::array<T, N>> {
...
+    bit_array sign_of_max_comp;
+
     /**
      * @brief Bit arrays that store the scaled-down representations of each vector component.
      */
     std::array<bit_array, N - 1> component_scaled_values;
 
     constexpr static size_t quat_magnitude_sz = (quat_sz - 3 * quat_component_sz);
-    constexpr static size_t vec_min_sz = 2 + 2 * vec_component_sz;
+    constexpr static size_t vec_min_sz = 2 + 2 * vec_component_sz + 1; // + 1 for the sign serializer
```
shihaocao: @tanishqaggarwal, I never checked any of the math of these # of bit constants, but this should be the only value I modified. Further the current test campaign passes with only this constant value changed.

CommonSoftware/test/serializers/test_serializers.cpp
```diff
@@ -447,47 +593,53 @@ void test_quat_serializer() {
                          f_quat_t,
                          d_quat_t>::type;
 
-    // TODO write serialization initializations for edge cases.
+    // TODO write serialization initializations for edge cases. // ?
```
shihaocao: Does this still need to be completed?

src/common/SerializerTypes.inl
```diff
@@ -495,27 +559,88 @@ class VectorSerializer : public SerializerBase<std::array<T, N>> {
...
         serialize(*dest);
         return true;
     }
 
     void deserialize(std::array<T, N>* dest) const override {
-        T magnitude = 0.0f;
-        magnitude_serializer->deserialize(&magnitude);
-        (*dest)[max_component_idx] = 1;
+        // a counter to keep track of the current serialized val index to process
+        unsigned int poor_mans_pointer = 0;
```
tanishqaggarwal: Can you rename this to something nicer

src/common/SerializerTypes.inl
```diff
@@ -495,27 +559,88 @@ class VectorSerializer : public SerializerBase<std::array<T, N>> {
...
+
+            // loop through each bit belonging to the serializer
+            for(size_t j = 0; j < vector_element_serializers[i]->bitsize(); j++){       
+                bit_vec.push_back(static_cast<bool>(this->serialized_val[poor_mans_pointer]));
+                poor_mans_pointer++;
+            }
+
+            bit_array new_bitarr(bit_vec);
+            vector_element_serializers[i]->set_bit_array(new_bitarr);
+        }
```
tanishqaggarwal: We can't use std::vector inside serialize!!! std::vector's construction causes allocation on the heap.
If you need the bit array from an underlying serializer, use get_bit_array(). To deal with the constness, either change the interface of get_bit_array() so that it's no longer const, or use const_cast to cast away the constness.

src/common/SerializerTypes.inl
```diff
@@ -495,27 +559,88 @@ class VectorSerializer : public SerializerBase<std::array<T, N>> {
...
+        for(unsigned int i = 0; i<(N-1); i++){
+            std::vector<bool> bit_vec;
+
+            // loop through each bit belonging to the serializer
+            for(size_t j = 0; j < vector_element_serializers[i]->bitsize(); j++){       
+                bit_vec.push_back(static_cast<bool>(this->serialized_val[poor_mans_pointer]));
+                poor_mans_pointer++;
+            }
+
+            bit_array new_bitarr(bit_vec);
```
tanishqaggarwal: This is especially bad because you're creating a copy of a heap object on the heap.

test/test_common_serializers/test_serializers.cpp
```diff
@@ -2,12 +2,85 @@
...
+ * @brief Returns the magnitude of a vector
+ * 
+ */
+template <typename T, size_t N>
+T magnitude_of(std::array<T, N>& src){
+    T sum = 0.0;
+    for(size_t i = 0; i<N; i++){
+        sum += src[i]*src[i];
+    }
+    return std::sqrt(sum);
```
tanishqaggarwal: You can use lin::norm for this as well. You could also use this function as a subroutine in the normalize() function above.

test/test_common_serializers/test_serializers.cpp
```diff
@@ -56,11 +134,13 @@ void test_value_float_or_double(std::shared_ptr<Serializer<T>>& s, const T val,
...
 void test_bool_serializer() {
     auto serializer = std::make_shared<Serializer<bool>>();
+    auto dl_deserializer = std::make_shared<Serializer<bool>>();
+
     bool dest_ptr;
 
     // Normal serialize and deserialize
-    test_value(serializer, true, true);
-    test_value(serializer, false, false);
+    test_value(serializer, dl_deserializer, true, true);
```
tanishqaggarwal: What does dl_ stand for?

test/test_common_serializers/test_serializers.cpp
```diff
@@ -274,45 +389,53 @@ void test_float_or_double_serializer() {
                   "To use this function, the value being tested must either be a float or a double.");
 
     std::shared_ptr<Serializer<T>> serializer;
+    std::shared_ptr<Serializer<T>> dl_deserializer;
     T threshold;
 
     // Test edge-case initializations
     // TODO
```
tanishqaggarwal: Seems like you finished this TODO? Thanks, feel free to remove the comment

test/test_common_serializers/test_serializers.cpp
```diff
@@ -357,46 +481,69 @@ void test_vec_serializer() {
 
     // TODO write serialization initializations for edge cases.
 
-    size_t csz = 0;
-    if (std::is_same<T, float>::value) csz = SerializerConstants::fvcsz;
-    else csz = SerializerConstants::dvcsz;
+    // size_t csz = 0;
+    // if (std::is_same<T, float>::value) csz = SerializerConstants::fvcsz;
+    // else csz = SerializerConstants::dvcsz;
+
```
tanishqaggarwal: You can remove this code

test/test_common_serializers/test_serializers.cpp
```diff
@@ -357,46 +481,69 @@ void test_vec_serializer() {
...
-    const size_t magnitude_bitsize = vec_bitsize - 2 - 2 * csz; // Used for error threshold.
-    T magnitude_err = 2.0 / powf(2, magnitude_bitsize);
-    magnitude_err = 2; // This is a bandaid to allow the unit test to pass temporarily.
-                       // We definitely need tighter precision than this.
 
     // (Deterministically) generate random vectors of magnitude 2, and see if they work 
     // with the serializer.
-    // Criterion for functionality: the vector that's reported has a displacement from the
-    // input vector of magnitude at most magnitude_err.
+    // Criterion for functionality: vector is within ??? degree, and within 1% of magnitude
```
tanishqaggarwal: Seems like you can replace ??? degree with like 1-2 degrees?

tanishqaggarwal: magnitude is also 0.1% right now.

CommonSoftware/test/serializers/test_serializers.cpp
tanishqaggarwal: I don't mind if it's not

src/common/SerializerTypes.inl
shihaocao: Oops, I forgot to change that

test/test_common_serializers/test_serializers.cpp
shihaocao: dl_ is short for downlink, wanted to keep variable names shorter.
Added a comment at all declaration to clarify things.

test/test_common_serializers/test_serializers.cpp
shihaocao: I didn't add any more test_value_float_or_double calls actually, so I wanted to ask if we had actually tested the full scope of serializer initializations or not

test/test_common_serializers/test_serializers.cpp
shihaocao: Fixed

src/common/SerializerTypes.inl
```diff
@@ -578,27 +578,25 @@ class VectorSerializer : public SerializerBase<std::array<T, N>> {
         
         // if there's a magnitude serializer, save data into the member variable
         if(N == 3){
-            std::vector<bool> bit_vec;
+            // use const_cast to circumvent get_bit_array returning a constant reference
```
tanishqaggarwal: You can remove this comment because the use of const_cast is understood to be doing this. Comments are better used for segments of code where logic is difficult to understand.

src/common/SerializerTypes.inl
```diff
@@ -578,27 +578,25 @@ class VectorSerializer : public SerializerBase<std::array<T, N>> {
...
+            bit_array& local_arr_ref = const_cast <bit_array&>(magnitude_serializer->get_bit_array());
             for(size_t i = 0; i < magnitude_serializer->bitsize(); i++){
-                bit_vec.push_back(this->serialized_val[poor_mans_pointer]);
+                local_arr_ref[i] = this->serialized_val[poor_mans_pointer];
```
tanishqaggarwal: Pls rename this variable

shihaocao: Oops right

PR #162, opened 2020-01-24
Migrated ADCS Device Drivers into the Source Folder

kkrol27:
I've migrated all the ADCS drivers into the source folder and refactored the ADCS code accordingly. I've also duplicated the ADCS files that were present in common software.

@shihaocao If you could pay close attention to the HAVT changes (shouldn't be any huge changes) that be great. Also, once this is merged, it'd be awesome if we could migreate FSW ADCS code to use the headers in `src/adcs/**` and not common software (we can remove ADCS code from common software then)!

Most of the changes are updates to file comments, include guards, and removing the `extern` keyword from some of the function interfaces.

It may be easier to check the branch out locally and poke around.

### DIFF
--- a/src/adcs/dev/I2CDevice.inl
+++ b/src/adcs/dev/I2CDevice.inl
@@ -1,6 +1,6 @@
 //
-// lib/Devices/I2CDevice.inl
-// ADCS
+// src/adcs/dev/I2CDevice.inl
+// FlightSoftware
 //
 // Contributors:
 //   Kyle Krol         kpk63@cornell.edu
@@ -12,6 +12,7 @@
 
 #include "I2CDevice.hpp"
 
+namespace adcs {
 namespace dev {
 
 inline bool I2CDevice::reset() {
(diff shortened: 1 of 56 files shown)


src/adcs/dev/I2CDevice.inl
```diff
@@ -37,7 +38,7 @@ inline bool I2CDevice::i2c_peek_errors() {
 inline bool I2CDevice::i2c_pop_errors() {
   if (this->i2c_peek_errors())
     if (++this->error_count >= DEV_I2C_ERROR_COUNT) 
-      this->dev::Device::disable(); //don't attempt further I2C comms
+      this->Device::disable(); //don't attempt further I2C comms
```
shihaocao: Why do we drop the dev::?

shihaocao: Self resolved: I think/assume it's to call the Device specific `disable()`

Followup: does that mean our previous fix for the infinite loop was wrong?

kkrol27: Nope, I just took all of the device drivers and put them in `dev::` as well.

PR #163, opened 2020-01-25
Merge common software into src/

tanishqaggarwal:
The files that are particularly important are the .ini files and the files in tools.


configs/adcs.ini
```diff
@@ -31,7 +29,7 @@ extends = adcs_teensy35_flight
 build_flags =
   ${adcs_teensy35_flight.build_flags}
   ${leader.build_flags}
-
+  -D FLIGHT
```
tanishqaggarwal: This compiler flag was required in order to prevent multiple definition errors due to main() conflicting between psim and ADCS

kkrol27: Ah okay interesting. Maybe throw a comment in there about it?

PR #164, opened 2020-01-26
fixed calibration of mag2

nhz2:
I fixed the magnetometer 2 calibration function in the driver, the calibration is also done on reset now. I also added a median average of 3 offset calculations into the calibration function because I found that rarely, the offset would be very wrong.


src/adcs/dev/MMC34160PJ.cpp
```diff
@@ -17,20 +17,12 @@ namespace dev {
...
 
 bool MMC34160PJ::reset() {
   this->I2CDevice::reset();
   while (this->is_functional()) {
-    this->i2c_begin_transmission();
-    this->i2c_write(MMC34160PJ::REG::INTERNAL_CONTROL_0);
-    this->i2c_write(this->sample_rate | 0b10); // Sets the SR and cont. mode
-    this->i2c_end_transmission(I2C_NOSTOP);
-    if (!this->i2c_pop_errors()) return true;
+    if(this->calibrate()) return true;
```
kkrol27: I didn't include this here originally because I was trying to keep the time overhead of reset calls low. I think we may want to have the flight computer explicitly call the calibration function each time?

I have the calibration functions called on startup regardless so the sensors will start calibrated.

src/adcs/dev/MMC34160PJ.cpp
```diff
@@ -43,26 +35,56 @@ void MMC34160PJ::disable() {
...
+        }
+      }
+  }
+  //reset to continous mode
+  this->i2c_begin_transmission();
+  this->i2c_write(MMC34160PJ::REG::INTERNAL_CONTROL_0);
+  this->i2c_write(this->sample_rate | 0b10); // Sets the SR and cont. mode
+  this->i2c_end_transmission(I2C_NOSTOP);
+  if (this->i2c_pop_errors()) return false;
+  return true;
```
kkrol27: And I think we should put this in the reset function...? Or no?
It'd be more consistent with other ADCS devices if it was there at least.

src/adcs/dev/MMC34160PJ.cpp
nhz2: Ok, I separated out the calibration.

src/adcs/dev/MMC34160PJ.hpp
```diff
@@ -67,11 +69,11 @@ class MMC34160PJ : public I2CDevice {
   /** Attempts to disable continuous measurement mode. */
   virtual void disable() override;
   /** Calibrates the magnetometer and updates the offset parameter. This process
-   *  will take about 11 ms and should be done periodically to ensure accuracy.
+   *  will take about 300 ms and should be done periodically to ensure accuracy.
    *  The device will be marked as disabled and the \c reset function must be
    *  called to resume readings.
    *  @returns True if the calibration was succesful and false otherwise. */
-  void calibrate();
+  bool calibrate();
```
tanishqaggarwal: If the calibration takes so long, ADCS won't be able to respond to I2C requests from master while this happens, right? How do we prevent this?

tanishqaggarwal: Also, what is the flight software-facing interface for calling calibrate? @kkrol27

nhz2: I guess it depends on how we want calibrate to be used. Right now its just done on startup. When we want to do calibration more often, the driver probably should be restructured to avoid busy waiting.

tanishqaggarwal: Can we do this now? I thought we were intending to calibrate every time we leave detumble.

nhz2: Fixing this would require some pretty big changes to how calibration works on a higher level, this PR is just so mag2 can output at-least somewhat usable data, so I can continue with characterization.

nhz2:
When I ran `pio run -e adcs_teensy35_script_MMC34160PJ_test -t upload` on my face 5 magnetometer mockup, This is what I got. So it seem to be working, at least in isolation from the rest of ADCS.

```
Starting
-352,-197,525
-351,-195,522
-352,-194,525
-355,-195,523
-353,-197,525
-349,-196,525
-353,-197,526
-353,-196,522
-354,-194,525
-353,-194,524
calibration
-352,-193,525
-350,-197,524
-353,-194,526
-353,-194,527
-353,-196,526
-352,-194,525
-353,-195,525
-355,-196,524
-352,-196,525
-355,-195,526
calibration
-351,-196,528
-351,-198,527
-353,-194,526
```

src/adcs/dev/MMC34160PJ.cpp
```diff
@@ -120,10 +147,12 @@ bool MMC34160PJ::single_read(uint16_t *array) {
   this->i2c_write(MMC34160PJ::REG::INTERNAL_CONTROL_0);
   this->i2c_write(0x01);
   this->i2c_end_transmission();
+  delay(10);
   while (!this->is_ready())
     if (!this->is_functional()) return false;
+    delay(5);
   if (!this->read()) return false;
-  for (unsigned int i = 0; i < 3; i++) array[i] = this->b_vec[i];
+  for (int i = 0; i < 3; i++) array[i] = this->b_vec[i];
```
tanishqaggarwal: These are pretty substantial delays for the is_ready function. Doesn't this function get called on every ADCS control cycle? Are you sure they're required?

nhz2: this is in the private function `bool MMC34160PJ::single_read(uint16_t *array) {`

src/adcs/dev/MMC34160PJ.cpp
```diff
@@ -120,10 +147,12 @@ bool MMC34160PJ::single_read(uint16_t *array) {
   this->i2c_write(MMC34160PJ::REG::INTERNAL_CONTROL_0);
   this->i2c_write(0x01);
   this->i2c_end_transmission();
+  delay(10);
   while (!this->is_ready())
     if (!this->is_functional()) return false;
+    delay(5);
```
tanishqaggarwal: The delay here is not part of the loop right? It would be good to unindent it one level to make this clearer.

nhz2: Nice catch, I forgot the brackets on the while loop.

PR #165, opened 2020-01-26
Two quick fixes for MCL

shihaocao:
- Remove assert, leave gs.get_hk()
- Change a 0 to 1 in GomspaceController

### DIFF
--- a/src/fsw/FCCode/Drivers/ADCS.cpp
+++ b/src/fsw/FCCode/Drivers/ADCS.cpp
@@ -14,7 +14,7 @@ using namespace Devices;
 
 #ifndef DESKTOP
 ADCS::ADCS(i2c_t3 &i2c_wire, unsigned char address)
-    : I2CDevice("adcs", i2c_wire, address, 1000000) {}
+    : I2CDevice("adcs", i2c_wire, address, 1000) {}
 #else
 ADCS::ADCS()
     : I2CDevice("adcs", 0) {}
--- a/src/fsw/FCCode/Drivers/ADCS.cpp
+++ b/src/fsw/FCCode/Drivers/ADCS.cpp
@@ -14,7 +14,7 @@ using namespace Devices;
 
 #ifndef DESKTOP
 ADCS::ADCS(i2c_t3 &i2c_wire, unsigned char address)
-    : I2CDevice("adcs", i2c_wire, address) {}
+    : I2CDevice("adcs", i2c_wire, address, 1000000) {}
 #else
 ADCS::ADCS()
     : I2CDevice("adcs", 0) {}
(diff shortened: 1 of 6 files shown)


src/fsw/FCCode/MissionManager.hpp
```diff
@@ -28,7 +28,7 @@ class MissionManager : public TimedControlTask<void> {
     #ifdef FLIGHT
         static constexpr unsigned int deployment_wait = 15000; // ~30 mins
     #else
-        static constexpr unsigned int deployment_wait = 10;
+        static constexpr unsigned int deployment_wait = 100;
```
fatimayousuf: I approved this because you made the right changes to the gomspace controller. Because you changed the deployment wait to 100, I think that you would have to change the text fixture in test_mission_manager.cpp and let 90 more steps occur (instead of 1) before checking that the mission state is in detumble.

src/fsw/FCCode/Drivers/ADCS.cpp
```diff
@@ -14,7 +14,7 @@ using namespace Devices;
 
 #ifndef DESKTOP
 ADCS::ADCS(i2c_t3 &i2c_wire, unsigned char address)
-    : I2CDevice("adcs", i2c_wire, address) {}
+    : I2CDevice("adcs", i2c_wire, address, 1000000) {}
```
tanishqaggarwal: Why the enormous timeout?

src/fsw/FCCode/MissionManager.hpp
tanishqaggarwal: Great catch @fatimayousuf .

src/fsw/FCCode/GomspaceController.cpp
```diff
@@ -150,11 +150,11 @@ GomspaceController::GomspaceController(StateFieldRegistry &registry, unsigned in
 
 void GomspaceController::execute() {
     //get hk data from struct in driver
-    assert(gs.get_hk());
+    gs.get_hk();
```
tanishqaggarwal: @fatimayousuf Can you open a separate PR later to turn the returned value of get_hk into a state field?

src/fsw/FCCode/Drivers/ADCS.cpp
shihaocao: It was the value that kyle and I allowed during testing, but clearly as of his post in software it's probably not necessary to be that high. I'm thinking 1 ms is ok

tanishqaggarwal: Be sure to add this to Recurring Constants Review

PR #168, opened 2020-01-26
added get_hk statefield

fatimayousuf:


src/fsw/FCCode/GomspaceController.hpp
```diff
@@ -31,6 +31,9 @@ class GomspaceController : public TimedControlTask<void> {
 
    protected:
     Devices::Gomspace &gs;
+
+    Serializer<bool> get_hk_sr;
+    ReadableStateField<bool> get_hk_f;
```
tanishqaggarwal: Sorry that I didn't mention this before. Could you try using the fault mechanism for this flag, instead of a simple boolean state field? It'll make it easy to act on loss of information.

PR #170, opened 2020-01-28
Updated ADCS Logging Capabilities and Removed Duplicate Headers

kkrol27:
I've updated the ADCS debug logging capabilities to include logging levels now. This makes it far easier to adjust how version you would like the output to be and is super helpful during testing.

To make the division between ADCS and FSW I've also wrapped up all things in ADCS into the namespace `adcs::`. This required some refactoring on the FSW side of things but everything is building again. @shihaocao if you could specifically take a look at those changes that'd be great.

There was a small update to the PSim submodule so it may be necessary to run `git submodule update --recursive`.

Lastly, there seem to be lots of lingering issues with the build system that need to get figured out. Specifically with some of the hardware tests it seems like. The `src_filter` parameter needs to be updated because the drivers aren't built by default anymore. I've done this for the ADCS test in this PR already.


src/adcs/ssa.cpp
```diff
@@ -30,23 +30,40 @@ dev::ADS1015 adcs[5];
 lin::Matrix<float, 5, 4> voltages = lin::zeros<float, 5, 4>();
 
 void setup() {
+  LOG_INFO_header
```
tanishqaggarwal: Why not just make "print the header" an argument in the actual print macro? That way a user doesn't have to remember to put it in before most prints.

kkrol27: Because you may want to make multiple calls to print a single line

kkrol27: At least I do that sometimes

src/adcs/rwa.cpp
```diff
@@ -30,35 +33,63 @@ dev::ADS1015 adcs[3];
...
 
 void setup() {
+  LOG_INFO_header
+  LOG_INFO_printlnF("Initializing the RWA module")
+
   // Setup and reset the potentiometer
   potentiometer.setup(pot_wire, pot_addr, pot_timeout);
   potentiometer.reset();
+
+#if LOG_LEVEL >= LOG_LEVEL_ERROR
```
tanishqaggarwal: This #if statement could probably be replaced with a macro to improve readability.

src/adcs/mtr.cpp
```diff
@@ -14,69 +14,93 @@
...
+      LOG_TRACE_printF(" disabled")
+      continue;
     }
-  } else {
-    // Magnetic torque rods are disabled
-    for (auto &mtr : mtrs) mtr.actuate(0);
+
+    // Clamp the command
+    float cmd = (mtr_cmd(i) > mtr_lim ? mtr_lim : mtr_cmd(i));
+    cmd = (mtr_cmd(i) < -mtr_lim ? -mtr_lim : mtr_cmd(i));
```
tanishqaggarwal: mtr_lim has been added to Recurring Constants Review, right?

src/adcs/rwa.cpp
kkrol27: Can macros expand to other conditional macros? I wasn't sure about that.

src/adcs/mtr.cpp
kkrol27: And should be. It's been in there for a while.

src/adcs/rwa.cpp
kkrol27: I will look into this in the future to aid with readability.

src/adcs/ssa.cpp
tanishqaggarwal: Why would you want multiple calls? Just use printf?

src/adcs/rwa.cpp
tanishqaggarwal: I think you could do something like https://stackoverflow.com/questions/14353104/is-it-possible-to-pass-a-block-of-code-as-a-macro-argument

kkrol27:
I'm gonna wait for @shihaocao on this one unless you want this in ASAP @tanishqaggarwal.

src/adcs/imu.cpp
```diff
@@ -40,54 +40,58 @@ static unsigned char update_mag(unsigned char mode, float mag_flt) {
...
-  DEBUG_print(String(mag_rd(0)) + "," + String(mag_rd(1)) + "," +
-      String(mag_rd(2)) + "," + String(data(0)) + "," + String(data(1)) + "," +
-      String(data(2)))
+
+  LOG_TRACE_header
+  LOG_TRACE_println("Updated magnetometer reading " + String(data(0)) + " "
+      + String(data(1)) + " " + String(data(2)))
+
+  LOG_TRACE_header
+  LOG_TRACE_println("Updated, filtered magnetometer reading "
```
shihaocao: `Updated, filtered`->`Updated filtered`?
`Updated, filtered`->`Updated, and filtered`?

Just marking this incase you intended something else

shihaocao: nvm, i see the convention you're going for now

PR #178, opened 2020-01-30
Low batt fault

fatimayousuf:


src/fsw/FCCode/GomspaceController.cpp
```diff
@@ -171,6 +173,14 @@ void GomspaceController::execute() {
         get_hk_fault.unsignal();
     }
 
+    // Check that the battery voltage is above 7300 mV
+    if (gs.hk->vbatt<7300){
```
tanishqaggarwal: We want this threshold to be configurable.

fatimayousuf: Do you mean it should be a statefield?

tanishqaggarwal: Yep.

PR #181, opened 2020-01-31
Eeprom statefield periods

fatimayousuf:


src/fsw/FCCode/EEPROMController.cpp
```diff
@@ -3,10 +3,11 @@
 #include <EEPROM.h>
 #endif
 
-EEPROMController::EEPROMController(StateFieldRegistry &registry, unsigned int offset, std::vector<std::string>& statefields)
+EEPROMController::EEPROMController(StateFieldRegistry &registry, unsigned int offset, std::vector<std::string>& statefields, 
+std::vector<unsigned int>& periods)
```
tanishqaggarwal: periods can be a const reference.

test/test_fsw_EEPROM/test_fsw_EEPROM.cpp
```diff
@@ -0,0 +1,198 @@
+#include "../StateFieldRegistryMock.hpp"
```
tanishqaggarwal: Looks like a solid test. Nice job.

src/fsw/FCCode/EEPROMController.cpp
```diff
@@ -3,10 +3,11 @@
 #include <EEPROM.h>
 #endif
 
-EEPROMController::EEPROMController(StateFieldRegistry &registry, unsigned int offset, std::vector<std::string>& statefields)
+EEPROMController::EEPROMController(StateFieldRegistry &registry, unsigned int offset, std::vector<std::string>& statefields,
```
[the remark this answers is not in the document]
fatimayousuf: Yes, I use it in the init() function. It gives the names of the statefields that we want to store in the EEPROM. There's also a statefield vector in the test file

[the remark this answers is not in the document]
tanishqaggarwal: It doesn't need to be an argument to the constructor if you're only using it inside the init() function though.

src/fsw/FCCode/MainControlLoop.hpp
```diff
@@ -51,6 +51,7 @@ class MainControlLoop : public ControlTask<void> {
     UplinkConsumer uplink_consumer; // Needs uplink packet from Quake Manager
 
     std::vector<std::string>statefields;
+    std::vector<unsigned int>periods;
```
tanishqaggarwal: Instead of this, could you create a file for the statefields and the periods, similar to how flow_data.cpp is set up?

src/fsw/FCCode/EEPROMController.cpp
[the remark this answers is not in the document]
fatimayousuf: ah gotcha

src/fsw/FCCode/MainControlLoop.hpp
fatimayousuf: not quite sure if I did this correctly, could you check this?

PR #185, opened 2020-02-01
Fault Update

shihaocao:
This upgrades the fault class and adds more naunced behavior.

There are now two more additional bool statefields. `signal_f` and `unsignal_f`. If these are set to true from the ground, the fault will be faulted or unfaulted for one cycle only, regardless of whether or not a `signal()` or `unsignal()` condition was met.

Further if the `override_f` or `suppress_f` statefields are set to true (from false), the Fault will reset its `num_consecutive_signals` to `0`.

### DIFF
--- a/src/fsw/FCCode/Fault.cpp
+++ b/src/fsw/FCCode/Fault.cpp
@@ -10,7 +10,7 @@ Fault::Fault(const std::string& name,
     override_f(name + ".override", fault_bool_sr),
     unsignal_f(name + ".unsignal", fault_bool_sr),
     // 65536 = 2^16 -1
-    persist_sr(0, 65535, 16),
+    persist_sr(65535),
     persistence_f(name + ".persistence", persist_sr)
 {
   set(default_setting);
--- a/src/fsw/FCCode/Fault.cpp
+++ b/src/fsw/FCCode/Fault.cpp
@@ -25,7 +25,6 @@ bool Fault::add_to_registry(StateFieldRegistry& r) {
     if (!r.add_writable_field(static_cast<WritableStateFieldBase*>(&suppress_f))) return false;
     if (!r.add_writable_field(static_cast<WritableStateFieldBase*>(&override_f))) return false;
     if (!r.add_writable_field(static_cast<WritableStateFieldBase*>(&unsignal_f))) return false;
-    if (!r.add_writable_field(static_cast<WritableStateFieldBase*>(&signal_f))) return false;
     return true;
 }
 
--- a/src/fsw/FCCode/Fault.cpp
+++ b/src/fsw/FCCode/Fault.cpp
@@ -24,6 +24,7 @@ bool Fault::add_to_registry(StateFieldRegistry& r) {
     if (!r.add_writable_field(static_cast<WritableStateFieldBase*>(&suppress_f))) return false;
     if (!r.add_writable_field(static_cast<WritableStateFieldBase*>(&override_f))) return false;
     if (!r.add_writable_field(static_cast<WritableStateFieldBase*>(&unsignal_f))) return false;
+    if (!r.add_writable_field(static_cast<WritableStateFieldBase*>(&persistence_f))) return false;
     return true;
 }
 
--- a/src/fsw/FCCode/Fault.cpp
+++ b/src/fsw/FCCode/Fault.cpp
@@ -22,7 +22,7 @@ bool Fault::add_to_registry(StateFieldRegistry& r) {
 }
 
 void Fault::signal() {
-    consider_pulls();
+    process_commands();
 
     if (cc > static_cast<unsigned int>(last_fault_time) || cc == 0) {
         num_consecutive_faults++;
@@ -35,7 +35,7 @@ void Fault::signal() {
 }
 
 void Fault::unsignal() {
-    consider_pulls();
+    process_commands();
     set(false);
     num_consecutive_faults = 0;
 }
@@ -52,8 +52,7 @@ bool Fault::is_faulted() const {
     else return get();
 }
 
-void Fault::consider_pulls(){
-    // suppress_f was set to true from ground
+void Fault::process_commands(){
     if(prev_override == false && override_f.get()){
         num_consecutive_faults = 0;
     }
@@ -62,4 +61,14 @@ void Fault::consider_pulls(){
     }
     prev_override = override_f.get();
     prev_suppress = suppress_f.get();
+
+    if(signal_f.get()){
+        signal_f.set(false);
+        // TODO: should this just be set(true)?
+        signal();
+    }
+    if(unsignal_f.get()){
+        unsignal_f.set(false);
+        unsignal();
+    }
 }
--- a/src/fsw/FCCode/Fault.cpp
+++ b/src/fsw/FCCode/Fault.cpp
@@ -21,35 +21,35 @@ bool Fault::add_to_registry(StateFieldRegistry& r) {
     if (!r.add_writable_field(static_cast<WritableStateFieldBase*>(this))) return false;
     if (!r.add_writable_field(static_cast<WritableStateFieldBase*>(&suppress_f))) return false;
     if (!r.add_writable_field(static_cast<WritableStateFieldBase*>(&override_f))) return false;
+    if (!r.add_writable_field(static_cast<WritableStateFieldBase*>(&unsignal_f))) return false;
+    if (!r.add_writable_field(static_cast<WritableStateFieldBase*>(&signal_f))) return false;
     return true;
 }
 
 void Fault::signal() {
-    process_commands();
-
     if (cc > static_cast<unsigned int>(last_fault_time) || cc == 0) {
-        num_consecutive_faults++;
+        num_consecutive_signals++;
         last_fault_time = cc;
     }
 
-    if (num_consecutive_faults > persistence) {
+    if (num_consecutive_signals > persistence) {
         set(true);
     }
+
+    process_commands();
 }
 
 void Fault::unsignal() {
-    process_commands();
     set(false);
-    num_consecutive_faults = 0;
+    num_consecutive_signals = 0;
+    process_commands();
 }
 
 bool Fault::is_faulted() const {
     if (override_f.get()) {
-        num_consecutive_faults = 0;
         return true;
     }
     else if (suppress_f.get()) {
-        num_consecutive_faults = 0;
         return false;
     }
     else return get();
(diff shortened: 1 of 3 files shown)


src/fsw/FCCode/Fault.cpp
```diff
@@ -18,33 +21,63 @@ bool Fault::add_to_registry(StateFieldRegistry& r) {
...
+    if(prev_suppress == false && suppress_f.get()){
+        num_consecutive_signals = 0;
+    }
+    prev_override = override_f.get();
+    prev_suppress = suppress_f.get();
+
+    if(signal_f.get()){
+        signal_f.set(false);
+        // TODO: can also just be signal()
+        set(true);
```
tanishqaggarwal: Let’s make this signal()

src/fsw/FCCode/Fault.hpp
```diff
@@ -57,15 +65,28 @@ class Fault : public WritableStateField<bool> {
                                       // occurred
 
     unsigned int persistence; // Persistence threshold for fault signal
```
tanishqaggarwal: Part of the ticket was also to make the persistence ground-configurable (I added this as a comment.)

test/test_fsw_fault/test_fsw_fault.cpp
```diff
@@ -11,86 +11,213 @@ void test_fault_normal_behavior() {
...
     TEST_ASSERT(fault.add_to_registry(r));
     TEST_ASSERT(r.find_readable_field("fault"));
     TEST_ASSERT(r.find_writable_field("fault.override"));
     TEST_ASSERT(r.find_writable_field("fault.suppress"));
+    TEST_ASSERT(r.find_writable_field("fault.signal"));
+    TEST_ASSERT(r.find_writable_field("fault.unsignal"));
+
+    TEST_ASSERT(fault2.add_to_registry(r));
+    TEST_ASSERT(r.find_readable_field("fault2"));
+    TEST_ASSERT(r.find_writable_field("fault2.override"));
```
tanishqaggarwal: We don’t need to test that the flags were created for both faults.

test/test_fsw_fault/test_fsw_fault.cpp
```diff
@@ -11,86 +11,213 @@ void test_fault_normal_behavior() {
...
+    WritableStateField<bool>* unsignal_fp = r.find_writable_field_t<bool>("fault.unsignal");
+
+    control_cycle_count++; fault.signal();
+    control_cycle_count++; fault.signal();
+    // num_consecutive_faults == 2 < persistence == 5
+    TEST_ASSERT_EQUAL(2, fault_fp->get_num_consecutive_signals());
+    TEST_ASSERT_FALSE(fault_fp->is_faulted());
+
+    control_cycle_count++; fault.signal(); override_fp->set(true);
+    // num_consecutive_faults == 3 < persistence == 5,
```
tanishqaggarwal: Try to keep your test comments declarative and have a minimum of additional information. It detracts from the understandability of the test.

test/test_fsw_fault/test_fsw_fault.cpp
```diff
@@ -11,86 +11,213 @@ void test_fault_normal_behavior() {
...
+    // apply the unsignal, even though we are in the signal condition
+    control_cycle_count++; fault.signal();
+    TEST_ASSERT_EQUAL(0, fault_fp->get_num_consecutive_signals());
+    TEST_ASSERT_FALSE(fault_fp->is_faulted());
+
+    // begin accumulating signals as normal
+    control_cycle_count++; fault.signal();
+    TEST_ASSERT_EQUAL(1, fault_fp->get_num_consecutive_signals());
+    TEST_ASSERT_FALSE(fault_fp->is_faulted());
+}
```
tanishqaggarwal: Test otherwise looks pretty complete in scope.

src/fsw/FCCode/Fault.hpp
shihaocao: Oops, right

src/fsw/FCCode/Fault.cpp
shihaocao: resolution: signal_f removed

src/fsw/FCCode/Fault.cpp
```diff
@@ -5,24 +5,26 @@ Fault::Fault(const std::string& name,
...
     WritableStateField<bool>(name, Serializer<bool>()),
     cc(control_cycle_count),
-    persistence(_persistence),
     fault_bool_sr(),
     suppress_f(name + ".suppress", fault_bool_sr),
     override_f(name + ".override", fault_bool_sr),
     unsignal_f(name + ".unsignal", fault_bool_sr),
-    signal_f(name + ".signal", fault_bool_sr)
+    // 65536 = 2^16 -1
+    persist_sr(0, 65535, 16),
```
tanishqaggarwal: You can just say persist_sr(65535). The serializer constructor will take care of the rest.

test/test_fsw_fault/test_fsw_fault.cpp
```diff
@@ -110,24 +104,21 @@ void test_process_commands(){
...
-    // num_consecutive_faults == 2 < persistence == 5
     TEST_ASSERT_EQUAL(2, fault_fp->get_num_consecutive_signals());
     TEST_ASSERT_FALSE(fault_fp->is_faulted());
 
     control_cycle_count++; fault.signal(); override_fp->set(true);
-    // num_consecutive_faults == 3 < persistence == 5,
-    // is faulted should be true since override_fp modifies the return of is fauled, 
+    // is_faulted should be true since override_fp modifies the return of is_faulted, 
     // but num_consecutive is not yet 0
     // remember, commands processed during signal() or unsignal()
```
tanishqaggarwal: I feel like these two lines just don't add any information about the test.

test/test_fsw_fault/test_fsw_fault.cpp
```diff
@@ -212,6 +198,65 @@ void test_process_commands(){
...
+    control_cycle_count++; fault.unsignal();
+    TEST_ASSERT_EQUAL(0, fault_fp->get_num_consecutive_signals());
+    TEST_ASSERT_FALSE(fault_fp->is_faulted());
+
+    // nominal satellite operation, and a command to change persistence to 3
+    control_cycle_count++; fault.unsignal(); persistence_fp->set(1);
+    TEST_ASSERT_EQUAL(0, fault_fp->get_num_consecutive_signals());
+    TEST_ASSERT_FALSE(fault_fp->is_faulted());
+
+    // fault now triggers after 2 signals()
```
tanishqaggarwal: Why does it trip after 2? Shouldn't it trip after 4? (Since 4 = 3 + 1)

test/test_fsw_fault/test_fsw_fault.cpp
```diff
@@ -212,6 +198,65 @@ void test_process_commands(){
...
+    control_cycle_count++; fault.signal();
+    TEST_ASSERT_EQUAL(2, fault_fp->get_num_consecutive_signals());
+    TEST_ASSERT_TRUE(fault_fp->is_faulted());
+
+    control_cycle_count++; fault.signal(); 
+    control_cycle_count++; fault.signal();
+    control_cycle_count++; fault.signal();
+    control_cycle_count++; fault.signal(); persistence_fp->set(10);
+    TEST_ASSERT_EQUAL(6, fault_fp->get_num_consecutive_signals());
+    TEST_ASSERT_TRUE(fault_fp->is_faulted());
```
tanishqaggarwal: You should add a comment saying you're resetting the persistence back to 10 in the middle of a bunch of signals.

test/test_fsw_fault/test_fsw_fault.cpp
```diff
@@ -212,6 +198,65 @@ void test_process_commands(){
...
+
+    // if signals <= persistence, fault is released
+    control_cycle_count++; fault.signal();
+    TEST_ASSERT_EQUAL(7, fault_fp->get_num_consecutive_signals());
+    TEST_ASSERT_FALSE(fault_fp->is_faulted());
```
tanishqaggarwal: This test block doesn't test anything new from the previous block.

test/test_fsw_fault/test_fsw_fault.cpp
```diff
@@ -212,6 +198,65 @@ void test_process_commands(){
...
+
+    // edit persistence to below current num_consecutive_signals
+    control_cycle_count++; fault.signal(); persistence_fp->set(5);
+    TEST_ASSERT_EQUAL(8, fault_fp->get_num_consecutive_signals());
+    TEST_ASSERT_FALSE(fault_fp->is_faulted());
+
+    // the next call of signal() triggers fault due to lower persistence
+    control_cycle_count++; fault.signal(); persistence_fp->set(5);
```
tanishqaggarwal: You don't need to set the persistence again

src/fsw/FCCode/Fault.cpp
```diff
@@ -32,9 +34,11 @@ void Fault::signal() {
         last_fault_time = cc;
     }
 
-    if (num_consecutive_signals > persistence) {
+    if (num_consecutive_signals > persistence_f.get()) {
         set(true);
     }
+    else
+        set(false);
```
tanishqaggarwal: Maybe a better place for this is process_commands.

test/test_fsw_fault/test_fsw_fault.cpp
```diff
@@ -212,6 +198,65 @@ void test_process_commands(){
...
+
+    // if signals <= persistence, fault is released
+    control_cycle_count++; fault.signal();
+    TEST_ASSERT_EQUAL(7, fault_fp->get_num_consecutive_signals());
+    TEST_ASSERT_FALSE(fault_fp->is_faulted());
+
+    // edit persistence to below current num_consecutive_signals
+    control_cycle_count++; fault.signal(); persistence_fp->set(5);
+    TEST_ASSERT_EQUAL(8, fault_fp->get_num_consecutive_signals());
+    TEST_ASSERT_FALSE(fault_fp->is_faulted());
```
tanishqaggarwal: If you move the persistence checking logic into process_commands, you could just end the test here instead of having the TEST_ASSERT_EQUAL(9...)

test/test_fsw_fault/test_fsw_fault.cpp
shihaocao: Oops my bad, my comment didn't match the actual command. I did `->set(1);`

src/fsw/FCCode/Fault.cpp
shihaocao: Moved everything to process_commands.

The price: Can no longer do one-time triggering, must use override_f

test/test_fsw_fault/test_fsw_fault.cpp
shihaocao: I think this a slightly different case than the original `->set(1)`

`->set(1)` was with a clean slate right after an `unsignal()`

Whereas this test final test case is for immediately signaling due to a drop in persistence.

test/test_fsw_fault/test_fsw_fault.cpp
shihaocao: This test makes sure that we don't get an unexpected behavior from changing persistence when we were originally already beyond the threshold.

PR #194, opened 2020-02-05
Added simple hardware fault handling to mission manager

tanishqaggarwal:
This adds power, pressurization failure, and ADCS hardware fault handling to the mission manager. Future PRs will handle errors due to Piksi failures and radio failures.


tanishqaggarwal:
Some more work is required.

tanishqaggarwal:
Added more robust logic and testing.

src/fsw/FCCode/MissionManager.cpp
```diff
@@ -55,26 +70,47 @@ MissionManager::MissionManager(StateFieldRegistry& registry, unsigned int offset
     set(sat_designation_t::undecided);
 }
 
-bool MissionManager::check_hardware_faults() {
-    return false;
-    // TODO
+bool MissionManager::check_adcs_hardware_faults() const {
+    return wheel1_adc_fault_fp->get() || wheel2_adc_fault_fp->get() || wheel3_adc_fault_fp->get()
```
shihaocao: all ->get()'s need to be ->is_faulted(); resolved and understood over PM

src/fsw/FCCode/MissionManager.hpp
```diff
@@ -34,12 +38,31 @@ class MissionManager : public TimedControlTask<void> {
...
+    static constexpr std::array<mission_state_t, 5> fault_responsive_states = {
+        mission_state_t::follower,
+        mission_state_t::leader,
+        mission_state_t::follower_close_approach,
+        mission_state_t::leader_close_approach,
+        mission_state_t::standby
+    };
+    // These states do not respond to fault conditions.
+    static constexpr std::array<mission_state_t, 6> fault_nonresponsive_states = {
+        mission_state_t::detumble,
```
shihaocao: Do we plan to use wheels for detumbling? Asking because currently, detumble will not do anything special for wheel faults.

src/fsw/FCCode/MissionManager.cpp
```diff
@@ -55,26 +76,48 @@ MissionManager::MissionManager(StateFieldRegistry& registry, unsigned int offset
...
+        const bool power_faulted = low_batt_fault_fp->is_faulted();
+        const bool adcs_faulted = check_adcs_hardware_faults();
+
+        if (prop_depressurized) {
+            transition_to_state(mission_state_t::standby,
+                adcs_state_t::point_standby,
+                prop_state_t::idle);
+            return;
+        }        
+        else if (power_faulted || adcs_faulted) {
```
shihaocao: I think this should be reordered, with the more critical fault conditions coming first. Otherwise, if we have adcs_faults and a prop_depressurized fault, we'd stay in point_standby.

tanishqaggarwal: Thanks for catching this; I'm going to reorder the logic and add a test to ensure that critical fault conditions trigger first.

src/fsw/FCCode/MissionManager.hpp
tanishqaggarwal: Detumble only uses the wheels.

src/fsw/FCCode/MissionManager.hpp
```diff
@@ -34,12 +38,31 @@ class MissionManager : public TimedControlTask<void> {
...
+    // These states respond to fault conditions.
+    static constexpr std::array<mission_state_t, 5> fault_responsive_states = {
+        mission_state_t::follower,
+        mission_state_t::leader,
+        mission_state_t::follower_close_approach,
+        mission_state_t::leader_close_approach,
+        mission_state_t::standby
+    };
+    // These states do not respond to fault conditions.
+    static constexpr std::array<mission_state_t, 6> fault_nonresponsive_states = {
```
shihaocao: Do these states also not respond to radio faults or GPS faults (that are not yet implemented)? If so, then perhaps a different structure for listing out whether or not a state responds to a fault might be cleaner. Like a map which has states as keys, and a set of kinds of faults it responds to as a value?

Once again, if these two sets of arrays work for all kinds of faults, then this should be fine.

shihaocao: As a followup question, are all fault responses going to be contained within MissionManager?

Example question: Is a prop over pressure fault handled within MissionManager, and then a state change to vent passed to PropController? or will PropController "catch" the event?

shihaocao: I think `safehold` is missing from this list, not a bug, but it should be in this list nonetheless.

tanishqaggarwal: Detumble, startup, and manual are for sure in this list. Docking, docked, and initialization hold I'm not so sure. I'll reorganize this set of states in the future if necessary.

Yes, all faults are handled by the mission manager since it has to make the decision on how to change the overall mission states. However, faults are _detected_ by their subsystem.

tanishqaggarwal: Good point thanks

src/fsw/FCCode/MissionManager.cpp
```diff
@@ -55,26 +76,50 @@ MissionManager::MissionManager(StateFieldRegistry& registry, unsigned int offset
...
     if (state == mission_state_t::startup) {
         set(radio_state_t::disabled);
     }
-    else {
-        bool faulted = check_hardware_faults();
-        if (faulted) {
+
+    // Step 2. Change state if faults exist.
+    bool is_fault_responsive_state = false;
+    for(mission_state_t fault_responsive_state : fault_responsive_states) {
```
stewartaslan: What is this for loop doing exactly? Is it checking the spacecraft state against other states to see if we are in a "fault responsive state"? What does it mean to be in a fault responsive state?

src/fsw/FCCode/MissionManager.cpp
```diff
@@ -55,26 +76,50 @@ MissionManager::MissionManager(StateFieldRegistry& registry, unsigned int offset
...
+        if (state == fault_responsive_state) {
+            is_fault_responsive_state = true;
+            break;
+        }
+    }
+    if (is_fault_responsive_state) {
+        const bool prop_depressurized = failed_pressurize_fp->is_faulted();
```
stewartaslan: Do we only check faults if we're in a "fault responsive state"?

src/fsw/FCCode/MissionManager.cpp
```diff
@@ -55,26 +76,50 @@ MissionManager::MissionManager(StateFieldRegistry& registry, unsigned int offset
...
         }
+        else if (prop_depressurized) {
+            transition_to_state(mission_state_t::standby,
+                adcs_state_t::point_standby,
+                prop_state_t::idle);
+            return;
+        } 
     }
 
+    // Step 3. Handle state.
```
stewartaslan: On line 136, why transition to safe hold if the state is undefined? Maybe standby instead since it's a "safer" state.

src/fsw/FCCode/MissionManager.cpp
```diff
@@ -55,26 +76,50 @@ MissionManager::MissionManager(StateFieldRegistry& registry, unsigned int offset
...
-bool MissionManager::check_hardware_faults() {
-    return false;
-    // TODO
+bool MissionManager::check_adcs_hardware_faults() const {
+    return wheel1_adc_fault_fp->is_faulted() || wheel2_adc_fault_fp->is_faulted()
+            || wheel3_adc_fault_fp->is_faulted()
+            || wheel_pot_fault_fp->is_faulted();
 }
 
 void MissionManager::execute() {
```
stewartaslan: Does execute happen every control cycle?

src/fsw/FCCode/MissionManager.hpp
```diff
@@ -20,6 +21,9 @@ class MissionManager : public TimedControlTask<void> {
     WritableStateField<double> detumble_safety_factor_f;
     WritableStateField<double> close_approach_trigger_dist_f; // Meters
     WritableStateField<double> docking_trigger_dist_f; // Meters
+    static constexpr double initial_detumble_safety_factor = 0.2;
```
stewartaslan: What does this safety factor mean? With respect to what?

src/fsw/FCCode/MissionManager.hpp
```diff
@@ -20,6 +21,9 @@ class MissionManager : public TimedControlTask<void> {
...
+    static constexpr double initial_close_approach_trigger_dist = 100;
```
stewartaslan: Units

src/fsw/FCCode/MissionManager.hpp
```diff
@@ -20,6 +21,9 @@ class MissionManager : public TimedControlTask<void> {
...
+    static constexpr double initial_docking_trigger_dist = 0.4;
```
stewartaslan: Units

src/fsw/FCCode/MissionManager.hpp
```diff
@@ -34,12 +38,32 @@ class MissionManager : public TimedControlTask<void> {
...
+    static constexpr std::array<mission_state_t, 5> fault_responsive_states = {
+        mission_state_t::follower,
+        mission_state_t::leader,
+        mission_state_t::follower_close_approach,
+        mission_state_t::leader_close_approach,
+        mission_state_t::standby
+    };
+    // These states do not respond to fault conditions.
+    static constexpr std::array<mission_state_t, 7> fault_nonresponsive_states = {
+        mission_state_t::detumble,
```
stewartaslan: Why not respond to faults in detumble? We could still go into safe hold. Any fault that would send us into standby should definitely be ignored though.

src/fsw/FCCode/MissionManager.hpp
```diff
@@ -34,12 +38,32 @@ class MissionManager : public TimedControlTask<void> {
...
+        mission_state_t::safehold,
+        mission_state_t::startup,
+        mission_state_t::manual,
+        mission_state_t::docking,
+        mission_state_t::docked,
```
stewartaslan: Does the quake fault response tree run in docked?

src/fsw/FCCode/MissionManager.hpp
```diff
@@ -34,12 +38,32 @@ class MissionManager : public TimedControlTask<void> {
...
+        mission_state_t::initialization_hold
```
stewartaslan: Shouldn't we still respond to power faults like low battery voltage?

src/fsw/FCCode/MissionManager.cpp
tanishqaggarwal: This is a case that would require there to have been some kind of jump in software execution. It's a good idea to bring it to safe hold in that case. The safe hold auto-exit will, in any case, bring the spacecraft to standby eventually.

src/fsw/FCCode/MissionManager.cpp
tanishqaggarwal: It's checking if the current state is a state in which we should respond to faults. If so, it responds to any existing faults.

src/fsw/FCCode/MissionManager.cpp
tanishqaggarwal: Yes

src/fsw/FCCode/MissionManager.hpp
tanishqaggarwal: If the angular momentum of the spacecraft is less than this factor times the maximum momentum of a single wheel, we can be pretty sure we're detumbled. This factor lives on psim as well.

src/fsw/FCCode/MissionManager.hpp
tanishqaggarwal: I'll add these!

src/fsw/FCCode/MissionManager.hpp
tanishqaggarwal: Hm yeah I agree. In a next iteration of this fault handler, I'll create a table of states vs. faults and use that as a more general means of dictating which faults get handled in which state, as @shihaocao suggested.

src/fsw/FCCode/MissionManager.hpp
tanishqaggarwal: The response tree runs in every mode except startup. This is good! If we don't have comms in docked mode we want to initiate the Quake response.

src/fsw/FCCode/MissionManager.hpp
tanishqaggarwal: Yeah maybe. I'll address this in the fault-state table thing in #207

src/fsw/FCCode/MissionManager.cpp
```diff
@@ -55,26 +76,50 @@ MissionManager::MissionManager(StateFieldRegistry& registry, unsigned int offset
...
     mission_state_t state = static_cast<mission_state_t>(mission_state_f.get());
 
+    // Step 1. Disable radio if in startup.
     if (state == mission_state_t::startup) {
         set(radio_state_t::disabled);
```
athena255: Does mission manager run before radio state starts up since radio currently starts at config

src/fsw/FCCode/MissionManager.hpp
```diff
@@ -34,12 +38,32 @@ class MissionManager : public TimedControlTask<void> {
...
+    };
 
    protected:
     /**
      * @brief Returns true if there are hardware faults on the spacecraft.
      */
-    bool check_hardware_faults();
+    bool check_adcs_hardware_faults() const;
```
athena255: does this apply to all possible hardware faults or just hardware faults in certain systems?

src/fsw/FCCode/MissionManager.hpp
```diff
@@ -95,13 +119,22 @@ class MissionManager : public TimedControlTask<void> {
...
     // Fields provided by Piksi and orbital estimation subsystems
-    const ReadableStateField<unsigned char>* piksi_mode_fp; // Piksi reading state (fixed RTK, float RTK, SPP, or error state)
     const ReadableStateField<d_vector_t>* propagated_baseline_pos_fp; // Propagated baseline position
 
     // Information from docking subsystem
     WritableStateField<bool> docking_config_cmd_f;
     const ReadableStateField<bool>* docked_fp;
 
+    // True if the battery is below the threshold for safehold.
+    Fault* low_batt_fault_fp;
```
athena255: Why do we handle faults here instead of having systems handle their own faults? It is unclear to me what types of faults should be handled here and what types of faults should be handled in their own systems.

src/fsw/FCCode/QuakeManager.cpp
```diff
@@ -29,8 +29,8 @@
 // Quake driver setup is initialized when QuakeController constructor is called
 QuakeManager::QuakeManager(StateFieldRegistry &registry, unsigned int offset) : 
     TimedControlTask<bool>(registry, "quake", offset),
-    max_wait_cycles_f("radio.max_wait", Serializer<unsigned int>(0, 24 * 60 * 60 * 1000 / 120)),
-    max_transceive_cycles_f("radio.max_transceive", Serializer<unsigned int>(0, 24 * 60 * 60 * 1000 / 120)),
+    max_wait_cycles_f("radio.max_wait", Serializer<unsigned int>(0, PAN::one_day_ccno)),
+    max_transceive_cycles_f("radio.max_transceive", Serializer<unsigned int>(0, PAN::one_day_ccno)),
```
athena255: Is this saying the maximum amount of transitive cycles wasted is one_day?

test/test_gsw_uplink_producer/test_1.json
```diff
@@ -1,7 +1,7 @@
 {
   "adcs.mode": 6,
   "adcs.cmd_attitude": 1343443,
-  "pan.mode": 7,
```
athena255: this change exists on the test json file, but not on the actual uplink producer test. That test may fail.

src/fsw/FCCode/MissionManager.cpp
tanishqaggarwal: Yep, mission manager runs in the control loop prior to quake manager (see MainControlLoop.cpp)

src/fsw/FCCode/MissionManager.hpp
tanishqaggarwal: This function only computes if there is an ADCS hardware fault.

src/fsw/FCCode/MissionManager.hpp
tanishqaggarwal: The faults that this PR addresses cause changes in the mission trajectory, so the mission manager needs to be aware of them and be the handler for them.

src/fsw/FCCode/QuakeManager.cpp
tanishqaggarwal: Yep, it places a limit on the serializer and therefore a limit on the maximum possible commanded max_wait and max_transceive.

test/test_gsw_uplink_producer/test_1.json
tanishqaggarwal: Let's resolve this offline

PR #195, opened 2020-02-05
Auto-exit mechanism for safe hold

tanishqaggarwal:
After 24 hours in safe hold, the mission manager will now force the satellite to reboot.


shihaocao:
Everything looks good here, I would just be sure to merge this in after #194.

PR #199, opened 2020-02-06
possible quake and bitstream fixes

athena255:


tanishqaggarwal:
@athena255 Can we close this, since we know these changes don't cause the build failure to go away?

PR #201, opened 2020-02-07
Dev/dcdc controller

fatimayousuf:


src/fsw/FCCode/DCDCController.cpp
```diff
@@ -0,0 +1,49 @@
...
+      reset_cmd_f("dcdc.reset_cmd", Serializer<bool>())
+{
+  add_writable_field(adcs_dcdc_cmd_f);
+  add_writable_field(sph_dcdc_cmd_f);
+  add_writable_field(disable_cmd_f);
+  add_writable_field(reset_cmd_f);
+
+  // Set default values
+  adcs_dcdc_cmd_f.set(dcdc.adcs_enabled());
+  adcs_dcdc_cmd_f.set(dcdc.adcs_enabled());
```
tanishqaggarwal: should say "sph"

src/fsw/FCCode/Drivers/DCDC.hpp
```diff
@@ -25,6 +25,12 @@ class DCDC : public Device {
     bool is_functional();
     void disable();
     void reset();
+
+    // Returns whether or not adcs is enabled. Default is false
+    bool adcs = false;
+
+    // Returns whether or not spike and hold is enabled. Default is false
+    bool sph = false;
```
tanishqaggarwal: Let's make these protected/private variables.

test/test_fsw_dcdc/test_fsw_dcdc.cpp
```diff
@@ -0,0 +1,88 @@
...
+    TEST_ASSERT_EQUAL(false, tf.dcdc.sph_enabled());
+
+    tf.adcs_dcdc_cmd_fp->set(true);
+    tf.sph_dcdc_cmd_fp->set(true);
+
+    tf.disable_cmd_fp->set(true);
+    tf.dcdc_controller->execute();
+    TEST_ASSERT_EQUAL(false, tf.dcdc.adcs_enabled());
+    TEST_ASSERT_EQUAL(false, tf.dcdc.sph_enabled());
+
```
tanishqaggarwal: Let's add in a reset cmd and verify that both pins go high again.

Some more possible test item(s): 
- If one pin was on and another was off, disable() makes them turn both off.
- If one pin is on and another is off, reset() makes them turn both on.

src/fsw/FCCode/DCDCController.hpp
```diff
@@ -0,0 +1,46 @@
...
+     */
+    void execute() override;
+
+   protected:
+    Devices::DCDC &dcdc;
+
+    //command statefield tells control task to enable/disable ADCS pin for DCDC
+    WritableStateField<bool> adcs_dcdc_cmd_f;
+
+    //command statefield tells control task to enable/disable ADCS pin for DCDC
```
tanishqaggarwal: Should say SPH.

Also the SPH pin also controls the docking system--might be worth mentioning.

src/fsw/FCCode/Drivers/DCDC.cpp
```diff
@@ -23,29 +23,57 @@ bool DCDC::is_functional() {
...
     #ifndef DESKTOP
     digitalWrite(dcdc_motor_enable_pin, HIGH);
+    #else
+    adcs=true;
+    #endif
+}
+
+const bool DCDC::adcs_enabled() {
+    #ifndef DESKTOP
+    return digitalRead(dcdc_sph_enable_pin);
```
stewartaslan: I think this should be `dcdc_motor_enable_pin`. Also, to be consistent with the electrical schematics, it would be good to change the name of `dcdc_sph_enable_pin` to `SpikeDockDCDC_EN` and `dcdc_motor_enable_pin` to `ADCSMotorDCDC_EN`

src/fsw/FCCode/DCDCController.hpp
```diff
@@ -0,0 +1,47 @@
...
+     * 
+     * @param registry 
+     * @param offset
+     * @param dcdc converter 
+     */
+    DCDCController(StateFieldRegistry& registry, unsigned int offset,
+        Devices::DCDC &dcdc);
+
+    /**
+     * @brief Sets statefield variables and moves the motor to docking/undocking
```
stewartaslan: Is this left over from the docking motor controller?

src/fsw/FCCode/Drivers/DCDC.cpp
```diff
@@ -56,5 +84,10 @@ void DCDC::reset() {
     delay(10);
     enable_sph();
     enable_adcs();
+    #else
+    disable_adcs();
+    disable_sph();
```
stewartaslan: We should have a delay here as well, same as the DESKTOP implementation. I think 100ms would be a good delay.

tanishqaggarwal: Shit, I missed the delay stuff. We can't have such large delays in Flight Software. @fatimayousuf can you reconfigure the driver to call reset in two steps instead of having a large delay?

fatimayousuf: Wait so there should or shouldn't be a delay???

stewartaslan: In reality there should be a delay because we want to wait a little before turning them back on during a reset, however Tanishq is saying that we don't want to have delays at the driver level. We only want to have delays at the controller level, like we did in the Quake controller.

So it sounds like the solution here could be to have reset happen at the controller level. it would first call the driver's disable function, then delay for 100ms, then call the driver's enable function.

fatimayousuf: That makes sense, thanks for clarifying!

src/fsw/FCCode/DCDCController.cpp
```diff
@@ -42,8 +42,14 @@ void DCDCController::execute() {
     }
 
     if (reset_cmd_f.get()) {
-        dcdc.reset();
-        reset_cmd_f.set(false);
+        if (dcdc.adcs_enabled() || dcdc.sph_enabled()) {
+            dcdc.disable();
+        }
+        if (!dcdc.adcs_enabled() && !dcdc.sph_enabled()) {
```
stewartaslan: Where does the delay happen? Do we just delay one full control cycle since reset_cmf_f only gets checked once per cycle?

tanishqaggarwal: Does the unit test check that there's a one-cycle delay due to this logic?

fatimayousuf: @stewartaslan Yeah that's the idea. @tanishqaggarwal I just added to the test file

Issue #192, opened 2020-02-04
Downlink producer is broken on Teensy when "pan.deployed" is included


tanishqaggarwal:
To band-aid this issue I'm removing pan.deployed from the telemetry flows in #188 but this needs to be reversed.

tanishqaggarwal:
Fixed by #371

PR #202, opened 2020-02-07
Fix EEPROM build

tanishqaggarwal:
This PR fixes Teensy builds of FSW. For some reason including the "pan.deployed" flag in telemetry flows causes the FSW to crap out (see #192).


src/eeprom_configs.cpp
```diff
@@ -1,4 +1,4 @@
 #include "eeprom_configs.hpp"
 
-const std::vector<std::string> PAN::statefields = {"pan.mode", "pan.deployed", "pan.sat_designation", "pan.cycle_no"};
```
tanishqaggarwal: It's weird that this exists on master but doesn't break builds.

tanishqaggarwal:
Doesn't fix anything.

PR #206, opened 2020-02-07
Add faults to the state field registry

tanishqaggarwal:


tanishqaggarwal:
Some more restructuring is required.

Issue #20, opened 2019-05-29
Fault decision tree and implementation for Quake faults 


tanishqaggarwal:
If Quake radio fails to initiate a successful SBDIX in 24 hours:
- Force the mission state to standby, wait 24 hours in that state. This puts the satellite in an comms-maximizing attitude.
- If we still don't have comms by  then, power cycle the radio every 8 hours three times.
- If we still don't have comms by then, go into safe hold, which after 24 hours will reset the entire satellite.

This should probably be implemented as a separate state machine on top of the Quake manager, and should be a subset of the mission manager.

Issue #207, opened 2020-02-08
Handle mission manager faults with a state-fault table


tanishqaggarwal:
In a next iteration of the mission manager fault handler, I'll create a table of states vs. faults and use that as a more general means of dictating which faults get handled in which state, as @shihaocao suggested.

For example, currently detumble is marked as a non-responsive state, but ideally it responds only to the low power fault. Same with initialization hold.

_Originally posted by @tanishqaggarwal in https://github.com/pathfinder-for-autonomous-navigation/FlightSoftware/pull/194/review_comment/create_

PR #209, opened 2020-02-08
Tiered fault handler for Quake

tanishqaggarwal:
# Tiered fault handler for Quake

Implements #20 .

### Summary of changes
- A class called QuakeFaultHandler has been implemented that manages a state machine for the current faulted state of the Quake.
- Removed fault handling from Mission Manager until I can open a PR with a more systematic approach to fault handling.

### Testing
The class has been thoroughly unit tested to verify the state behavior and state transition behavior for the Quake Fault Machine.

No HOOTL/HITL testing is currently available since the class is not integrated into the MainControlLoop.

### Constants
The diff to the `constants` in the root-level repository is just line # changes.

### Documentation Evidence
Added a description of the Quake Fault Handler to RTD:
https://pan-software.readthedocs.io/en/latest/flight_software/fault_management.html

---
This change is [Reviewable](https://reviewable.io/reviews/pathfinder-for-autonomous-navigation/flightsoftware/209)


src/fsw/FCCode/QuakeFaultHandler.hpp
```diff
@@ -0,0 +1,140 @@
...
+      const bool desired_state_is_valid = 
+          (s == mission_state_t::standby) 
+          || (s == mission_state_t::safehold)
+          || (s == mission_state_t::manual);
+      if(!desired_state_is_valid) assert(false);
+      else state = s;
+    }
+  private:
+    mission_state_t state = mission_state_t::manual;
+};
```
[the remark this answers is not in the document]
tanishqaggarwal: I'm not sure how else to handle it. Since MissionManager contains the QuakeFaultHandler, the mock needs to be compiled into the MissionManager when unit-testing the latter. This requires the fault handler mock to exist inside FCCode doesn't it?

tanishqaggarwal:
Closing so that we can have a more serious discussion about restructuring the spaghetti nature of fault handling.

tanishqaggarwal:
Reopening after removing fault integration with MissionManager. I'll reintegrate when implementing #207

src/fsw/FCCode/QuakeFaultHandler.hpp
```diff
@@ -0,0 +1,140 @@
...
+    /**
+     * @brief Runs the fault state machine and returns the suggested
+     * mission state.
+     * 
+     * @return mission_state_t Can either be standby, safe hold, or
+     * manual. The mission manager should switch to standby or
+     * safe hold if this state machine requires it, but should take
+     * no action when manual is returned.
+     */
+    mission_state_t execute();
```
athena255: Why does executing QuakeFaultHandler return a mission_state_t? What do you mean by suggested?

src/fsw/FCCode/QuakeFaultHandler.hpp
```diff
@@ -0,0 +1,140 @@
...
+
+    /**
+     * @brief States of fault machine. Description:
+     * - Unfaulted: We've received comms in the last 24 hours
+     * - Forced standby: We haven't had comms in the past 24 hours so the
+     *   satellite is forced into standby mode to try to maximize comms.
+     * - Power cycle 1: Forced standby didn't work. So the entry into this
+     *   state causes a power cycle of the radio and waits 8 hours before
+     *   transitioning to...
+     * - Power cycle 2: Same behavior as power cycle 1.
```
athena255: This is minor, but why did we not just keep a count of power cycles?

src/fsw/FCCode/QuakeFaultHandler.hpp
```diff
@@ -0,0 +1,140 @@
...
+    unsigned int cur_state_entry_ccno = 0;
+
+    /**
+     * @brief Indicator fields used by the fault handler.
+     */
+    const InternalStateField<unsigned char>* radio_state_fp;
+    const InternalStateField<unsigned int>* last_checkin_cycle_fp;
+    WritableStateField<bool>* power_cycle_radio_fp;
+
+    const unsigned int& control_cycle_count = TimedControlTaskBase::control_cycle_count;
```
athena255: Why does this reference exist when the control cycle count is global?

src/fsw/FCCode/QuakeFaultHandler.hpp
```diff
@@ -0,0 +1,140 @@
...
+    }
+    mission_state_t get_output() const {
+      return state;
+    }
+    void set_output(mission_state_t s) {
+      const bool desired_state_is_valid = 
+          (s == mission_state_t::standby) 
+          || (s == mission_state_t::safehold)
+          || (s == mission_state_t::manual);
+      if(!desired_state_is_valid) assert(false);
```
athena255: How does Teensy handle asserts? Are they just ignored? If we are focused on recovering from faults, shouldn't we automatically set the state to safehold or something?

src/fsw/FCCode/QuakeFaultHandler.hpp
```diff
@@ -0,0 +1,140 @@
...
+    bool less_than_one_day_since_successful_comms() const;
+    bool in_state_for_more_than_time(const unsigned int time) const;
+    bool radio_is_disabled() const;
+};
+
+/**
+ * @brief Mocking class that can be used by Mission Manager's unit test to
+ * inject Quake faults in a easily controllable way.
+ */
+class QuakeFaultHandlerMock {
```
athena255: Why is this test class here and not inside the test directory? This does not seem to implement fault injection.

src/fsw/FCCode/QuakeFaultHandler.hpp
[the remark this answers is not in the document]
athena255: Why does the mock need to be compiled into MissionManager? That does not seem ideal.

src/fsw/FCCode/QuakeFaultHandler.cpp
```diff
@@ -0,0 +1,113 @@
...
+
+void QuakeFaultHandler::transition_to(fault_checker_state_t next_state) {
+    cur_state = next_state;
+    cur_state_entry_ccno = control_cycle_count;
+}
+
+mission_state_t QuakeFaultHandler::dispatch_unfaulted() {
+    if (!radio_is_disabled() && !less_than_one_day_since_successful_comms()) {
+        transition_to(fault_checker_state_t::forced_standby);
+        return mission_state_t::standby;
```
athena255: We would put mission manager on standby because we failed to receive comms?

src/fsw/FCCode/QuakeFaultHandler.cpp
```diff
@@ -0,0 +1,113 @@
...
+    }
+}
+
+mission_state_t QuakeFaultHandler::dispatch_forced_standby() {
+    if (in_state_for_more_than_time(PAN::one_day_ccno)) {
+        power_cycle_radio_fp->set(true);
+        transition_to(fault_checker_state_t::powercycle_1);
+        return mission_state_t::standby;
+    }
+    else if (radio_is_disabled() || less_than_one_day_since_successful_comms()) {
```
athena255: Why go to unfaulted just because the radio is disabled?

athena255: Also, if you're planning to go into unfaulted, shouldn't you set the power_cycle_radio_fp to false?

src/fsw/FCCode/QuakeFaultHandler.cpp
```diff
@@ -0,0 +1,113 @@
...
+}
+
+mission_state_t QuakeFaultHandler::dispatch_powercycle_2() {
+    if (in_state_for_more_than_time(PAN::one_day_ccno / 3)) {
+        power_cycle_radio_fp->set(true);
+        transition_to(fault_checker_state_t::powercycle_3);
+        return mission_state_t::standby;
+    }
+    else if (less_than_one_day_since_successful_comms()) {
+        transition_to(fault_checker_state_t::unfaulted);
```
athena255: shouldn't we set power_cycle_radio_fp to false?

src/fsw/FCCode/QuakeFaultHandler.cpp
```diff
@@ -0,0 +1,113 @@
...
+        return mission_state_t::manual;
+    }
+    else {
+        return mission_state_t::safehold;
+    }
+}
+
+bool QuakeFaultHandler::less_than_one_day_since_successful_comms() const {
+    return control_cycle_count - last_checkin_cycle_fp->get() < PAN::one_day_ccno;
```
athena255: Is there a possibility of integer underflow here?

test/test_fsw_quake_fault_handler/test_quake_fault_handler.cpp
```diff
@@ -0,0 +1,328 @@
...
+    std::shared_ptr<InternalStateField<unsigned int>> last_checkin_cycle_fp;
+    std::shared_ptr<WritableStateField<bool>> radio_power_cycle_fp;
+
+    std::unique_ptr<QuakeFaultHandler> qfh;
+
+    void disable_radio() {
+        radio_state_fp->set(static_cast<unsigned char>(radio_state_t::disabled));
+    }
+    void enable_radio() {
+        radio_state_fp->set(static_cast<unsigned char>(radio_state_t::wait));
```
athena255: This should be transitioned to config.

test/test_fsw_quake_fault_handler/test_quake_fault_handler.cpp
```diff
@@ -0,0 +1,328 @@
...
+    TEST_ASSERT_EQUAL(mission_state_t::standby, qfh.execute());
+
+    qfh.set_output(mission_state_t::safehold);
+    TEST_ASSERT_EQUAL(mission_state_t::safehold, qfh.execute());
+
+    qfh.set_output(mission_state_t::standby);
+    TEST_ASSERT_EQUAL(mission_state_t::standby, qfh.execute());
+}
+
+void test_qfh_initialization() { TestFixture tf; }
```
athena255: This test demonstrates nothing.

src/fsw/FCCode/QuakeFaultHandler.hpp
```diff
@@ -0,0 +1,140 @@
+#ifndef quake_fault_handler_HPP_
+#define quake_fault_handler_HPP_
+
+#include "TimedControlTask.hpp"
+#include "mission_state_t.enum"
+
+/**
+ * @brief This state machine class implements a tiered fault system
+ * for addressing comms failures. See Issue #20 on Flight Software
```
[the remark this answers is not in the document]
tanishqaggarwal: Thanks for catching this laziness, will fix

src/fsw/FCCode/MissionManager.cpp
```diff
@@ -80,50 +80,8 @@ MissionManager::MissionManager(StateFieldRegistry& registry, unsigned int offset
...
-            return;
-        }
-        else if (prop_depressurized) {
-            transition_to_state(mission_state_t::standby,
-                adcs_state_t::point_standby,
-                prop_state_t::idle);
-            return;
-        } 
-    }
+    const mission_state_t state = static_cast<mission_state_t>(mission_state_f.get());
```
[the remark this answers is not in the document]
tanishqaggarwal: It's just const because it's a local variable whose value is never changed after initialization. It's a bit verbose but I've gotten used to programming this way--it ensures that your compiler will catch mistakes related to modifying variables

src/fsw/FCCode/QuakeFaultHandler.hpp
tanishqaggarwal: Just so we don't have to drag around the fully-qualified name of the cycle count inside the implementation. It makes things slightly cleaner.

I could probably just move this line into the declaration though to unclutter the header file

test/test_fsw_quake_fault_handler/test_quake_fault_handler.cpp
tanishqaggarwal: Thanks

src/fsw/FCCode/QuakeFaultHandler.hpp
tanishqaggarwal: Teensy ignores asserts. The goal of this mocking class is to be used from inside the unit test, so this assert is just a small amount of protection to ensure that user of this mocking class doesn't use the class in a way that's incompatible with the QuakeFaultHandler interface specification.

src/fsw/FCCode/QuakeFaultHandler.hpp
tanishqaggarwal: You're right, that's a much better way to handle this. I'll refactor the logic to use counting once I standardize fault handling (#207).

I'm sort of obsessed with state machines

src/fsw/FCCode/QuakeFaultHandler.hpp
tanishqaggarwal: The QuakeFaultHandler recommends a state to the mission manager based on the current faulted state of the Quake. The mission manager will transition to the recommended state if the Mission Manager isn't in safehold.

src/fsw/FCCode/QuakeFaultHandler.hpp
tanishqaggarwal: I should've been clearer about "fault injection": what I mean is that it allows the unit test to easily set the QuakeFaultHandler's recommended mission state and thus test the Mission Manager's response to various recommendations, i.e. how it responds to Quake faults.

src/fsw/FCCode/QuakeFaultHandler.cpp
tanishqaggarwal: Yep, but only if we're in a mission state like follower/leader/close approach. If we've failed to receive comms then it makes sense to point in a direction that maximizes comms/power, which is the pointing strategy in standby.

test/test_fsw_quake_fault_handler/test_quake_fault_handler.cpp
tanishqaggarwal: If the QuakeFaultHandler isn't being fed the right state field inputs by the test fixture, then this test catches the error. I can remove it though since the field inputs are OK now

src/fsw/FCCode/QuakeFaultHandler.cpp
tanishqaggarwal: If the radio is manually disabled we don't want the fault state machine stepping through its states because then it will trigger fault responses (like powercycling the radio, recommending standby, etc.). This is particularly important during the deployment wait period.

At this point it's ok to not set power_cycle_radio_fp() to false because it's already false. This flag is nearly almost always false because the Gomspace Controller auto-falses power_cycle_radio_fp after power cycling.

src/fsw/FCCode/QuakeFaultHandler.cpp
tanishqaggarwal: See above.

src/fsw/FCCode/QuakeFaultHandler.cpp
tanishqaggarwal: I thought about this. Here's the sequence of events in the control cycle:

QuakeManager
----------------- < Control cycle boundary
Clock Manager
...
Mission Manager
  - QuakeFaultHandler

last_checkin_cycle_fp() must've been updated prior to the control cycle boundary, but clock manager just ran, so control_cycle_count should always be greater than or equal to the last_checkin_cycle_fp.

src/fsw/FCCode/QuakeFaultHandler.hpp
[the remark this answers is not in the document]
tanishqaggarwal: I agree. Do you have a suggestion on a better way to unit test the MissionManager with the QuakeFaultHandler inside it? When unit-testing MissionManager we shouldn't also be stepping through the complex internal logic of QuakeFaultHandler, which is why during unit testing I had the MissionManager conditionally compile a simpler substitute for the QuakeFaultHandler.

src/fsw/FCCode/MissionManager.cpp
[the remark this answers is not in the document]
athena255: I’m confused since this is a state machine. Shouldn’t the state be modifiable?

src/fsw/FCCode/QuakeFaultHandler.hpp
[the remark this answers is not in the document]
athena255: I don't have a good solution right now, but I will think about it. I don't think we should create a substitute for QuakeFaultHandler in order to test QuakeFaultHandler. That is not really testing QuakeFaultHandler.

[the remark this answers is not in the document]
tanishqaggarwal: That's exactly the point. We don't want to test QuakeFaultHandler; we want to test MissionManager's _response_ to the QuakeFaultHandler. Testing this response doesn't require a full-fledge QuakeFaultHandler underneath MissionManager; instead, we want to have a simple class that lets us easily test MissionManager's response to all possible outputs of QuakeFaultHandler.

tanishqaggarwal:
___
*[src/fsw/FCCode/MissionManager.cpp, line 84 at r3](https://reviewable.io/reviews/pathfinder-for-autonomous-navigation/flightsoftware/209#-M1kjAxUyYLERoTWwL8F-r3-84:-M1kk1Mu9l477GQqUcz4:blwd258) ([raw file](https://github.com/pathfinder-for-autonomous-navigation/flightsoftware/blob/327c8bf8cc35237d601f2ac9624258565fd8f0a5/src/fsw/FCCode/MissionManager.cpp#L84)):*
<details><summary><i>Previously, athena255 (Athena ) wrote…</i></summary><blockquote>

I’m confused since this is a state machine. Shouldn’t the state be modifiable?
</blockquote></details>

I'm just storing a temporary copy of the state, and I've marked this copy as const.

src/fsw/FCCode/MissionManager.cpp
```diff
@@ -80,50 +80,8 @@ MissionManager::MissionManager(StateFieldRegistry& registry, unsigned int offset
     set(sat_designation_t::undecided);
 }
 
-bool MissionManager::check_adcs_hardware_faults() const {
```
tanishqaggarwal: Removed this code so that we can better handle fault handling integration in #207

src/fsw/FCCode/QuakeFaultHandler.cpp
```diff
@@ -0,0 +1,115 @@
...
+        case fault_checker_state_t::unfaulted:      return dispatch_unfaulted(); break;
+        case fault_checker_state_t::forced_standby: return dispatch_forced_standby(); break;
+        case fault_checker_state_t::powercycle_1:   return dispatch_powercycle_1(); break;
+        case fault_checker_state_t::powercycle_2:   return dispatch_powercycle_2(); break;
+        case fault_checker_state_t::powercycle_3:   return dispatch_powercycle_3(); break;
+        case fault_checker_state_t::safehold:       return dispatch_safehold(); break;
+        default:
+            return mission_state_t::manual;
+    }
+    return mission_state_t::manual;
```
kkrol27: I may just be missing something but is there a reason for the extra `break` and `return` statements?

src/fsw/FCCode/MissionManager.cpp
```diff
@@ -156,8 +116,8 @@ void MissionManager::dispatch_startup() {
     // Step 2. Turn radio on, and check for hardware faults that would necessitate
     // going into an initialization hold. If faults exist, go into
     // initialization hold, otherwise detumble.
-    set(radio_state_t::wait);
-    if (check_adcs_hardware_faults()) {
+    set(radio_state_t::config);
+    if (false) {
```
kkrol27: Again probably just missing something but why do we have `if (false)` here currently?

src/fsw/FCCode/QuakeFaultHandler.cpp
tanishqaggarwal: Just safety. Return ends the function, break ends the switch statement; I tend to include both just to "satisfy" both statements

src/fsw/FCCode/MissionManager.cpp
tanishqaggarwal: I have this so that no hardware fault response happens yet. This PR is only meant to add the state machine to the codebase but not to the actual flow of execution

test/test_fsw_quake_fault_handler/test_quake_fault_handler.cpp
```diff
@@ -0,0 +1,326 @@
...
+
+    void disable_radio() {
+        radio_state_fp->set(static_cast<unsigned char>(radio_state_t::disabled));
+    }
+    void enable_radio() {
+        radio_state_fp->set(static_cast<unsigned char>(radio_state_t::config));
+    }
+
+    TestFixture(fault_checker_state_t initial_state = fault_checker_state_t::unfaulted) {
+        radio_state_fp = registry.create_internal_field<unsigned char>("radio.state");
```
athena255: Shouldn't you test this with the Quake Manager?

test/test_fsw_quake_fault_handler/test_quake_fault_handler.cpp
```diff
@@ -0,0 +1,326 @@
...
+        tf.step_and_expect(mission_state_t::standby, fault_checker_state_t::forced_standby);
+    }
+}
+
+void test_qfh_forced_standby() {
+    // If it's been more than 24 hours since comms, the fault handler
+    // should cause a power cycle and move to the powercycle_1 state.
+    {
+        TestFixture tf{fault_checker_state_t::forced_standby};
+        tf.set_cur_state_entry_ccno(one_day_ccno);
```
athena255: Perhaps we should check the boundary cases. like check that when it's one control cycle before moving into power cycle, that we are not in power cycle.

test/test_fsw_quake_fault_handler/test_quake_fault_handler.cpp
```diff
@@ -0,0 +1,326 @@
...
+    }
+}
+
+void test_qfh_powercycle_1() {
+    // If it's been more than 8 hours since comms, the fault handler
+    // should cause a power cycle and move to the powercycle_2 state.
+    {
+        TestFixture tf{fault_checker_state_t::powercycle_1};
+        tf.set_cur_state_entry_ccno(one_day_ccno);
+        cc_count = one_day_ccno + one_day_ccno / 3;
```
athena255: should check the boundary case.

test/test_fsw_quake_fault_handler/test_quake_fault_handler.cpp
```diff
@@ -0,0 +1,326 @@
...
+        cc_count = 0;
+    }
+
+
+    // If we've (recently) received comms within 8 hours since the state transition, the state
+    // should return to unfaulted immediately.
+    {
+        TestFixture tf{fault_checker_state_t::powercycle_1};
+        tf.check_state_returns_to_unfaulted_if_comms_recvd(one_day_ccno / 3);
+    }
```
athena255: It might be good to separate out tests and make sure that each test only tests one function. That way, we know that each function that we are testing is independent of all other tests.

test/test_fsw_quake_fault_handler/test_quake_fault_handler.cpp
tanishqaggarwal: We just want to create an input to the fault handler. I don't think it's a good idea to do that with the QuakeManager since it introduces the behavior of a complex, unrelated state machine into the _unit_ test.

test/test_fsw_quake_fault_handler/test_quake_fault_handler.cpp
tanishqaggarwal: I agree with your point generally about test independence. In this case, I wanted to minimize the number of one- or two-line functions and instead have one function per fault handler state. The brackets do a good job of separating test fixtures for independent test steps. The control cycle count is reset in the test fixture's construction, so this ensures a clean separation between test steps.

test/test_fsw_quake_fault_handler/test_quake_fault_handler.cpp
tanishqaggarwal: You were right to ask for these cases; there was an off by one issue 😳

tanishqaggarwal: Nevermind, not an off-by-one issue, but rather a lack of important assertions

src/fsw/FCCode/QuakeFaultHandler.cpp
```diff
@@ -0,0 +1,119 @@
...
+mission_state_t QuakeFaultHandler::execute() {
+    switch(cur_state) {
+        case fault_checker_state_t::unfaulted:      return dispatch_unfaulted(); break;
+        case fault_checker_state_t::forced_standby: return dispatch_forced_standby(); break;
+        case fault_checker_state_t::powercycle_1:   return dispatch_powercycle_1(); break;
+        case fault_checker_state_t::powercycle_2:   return dispatch_powercycle_2(); break;
+        case fault_checker_state_t::powercycle_3:   return dispatch_powercycle_3(); break;
+        case fault_checker_state_t::safehold:       return dispatch_safehold(); break;
+        default:
+            return mission_state_t::manual;
```
stewartaslan: Why is manual the default mission state?

src/fsw/FCCode/QuakeFaultHandler.cpp
```diff
@@ -0,0 +1,119 @@
...
+    return mission_state_t::manual;
+}
+
+void QuakeFaultHandler::transition_to(fault_checker_state_t next_state) {
+    cur_state = next_state;
+    cur_state_entry_ccno = control_cycle_count;
+}
+
+mission_state_t QuakeFaultHandler::dispatch_unfaulted() {
+    if (!radio_is_disabled() && !less_than_one_day_since_successful_comms()) {
```
stewartaslan: Why is it even possible to disable the radio? Seems a bit dangerous to have that option.

src/fsw/FCCode/QuakeFaultHandler.hpp
stewartaslan: I am still wondering why this has to be in flight code rather than in the test directory.

src/fsw/FCCode/QuakeFaultHandler.cpp
```diff
@@ -0,0 +1,119 @@
...
+#include "constants.hpp"
+#include "radio_state_t.enum"
+
+const unsigned int& control_cycle_count = TimedControlTaskBase::control_cycle_count;
+
+QuakeFaultHandler::QuakeFaultHandler(StateFieldRegistry& r) : ControlTask<mission_state_t>(r) {
+    radio_state_fp        = find_internal_field<unsigned char>("radio.state", __FILE__, __LINE__);
+    last_checkin_cycle_fp = find_internal_field<unsigned int>("radio.last_comms_ccno", __FILE__,
+                                                                  __LINE__);
+    power_cycle_radio_fp  = find_writable_field<bool>("gomspace.power_cycle_output1_cmd", __FILE__,
```
stewartaslan: How does GOMSpace power cycle reporting work? Is this state field an integer number that reports the control cycle on which GOMSpace power cycled output 1?

test/test_fsw_quake_fault_handler/test_quake_fault_handler.cpp
```diff
@@ -0,0 +1,342 @@
...
+
+        qfh = std::make_unique<QuakeFaultHandler>(registry);
+        qfh->cur_state = initial_state;
+    }
+
+    // Below are getter and setter methods for the test harness, listed in order of
+    // increasing invasiveness/complexity.
+
+    // Verify that the radio was powercycled (or not).
+    void check_powercycled() { TEST_ASSERT_TRUE(radio_power_cycle_fp->get()); }
```
stewartaslan: For HITL testing, does the Quake report a boot count or something so we can see if it actually rebooted?

src/fsw/FCCode/QuakeFaultHandler.cpp
tanishqaggarwal: The radio is initially disabled during startup and then never disabled autonomously.

src/fsw/FCCode/QuakeFaultHandler.cpp
tanishqaggarwal: The power cycle flag is just a boolean value that can be set "true" to cause a power cycle of the radio. The Gomspace Controller then gets notified, sets the flag to "false", and performs a reset

test/test_fsw_quake_fault_handler/test_quake_fault_handler.cpp
tanishqaggarwal: I'll add a ticket to make sure we keep track of the # of power cycles of a device.

tanishqaggarwal: See #303

src/fsw/FCCode/QuakeFaultHandler.cpp
tanishqaggarwal: This PR uses "manual" to denote that the fault handler is recommending no action. In #262 I made the interface better by having fault handlers return an enum value instead.

I think "manual" should be the default recommended state if the Quake fault handler is in some undefined state (which is only possible due to a radiation mistake). Maybe I should also add code there that resets the state machine to a reasonable value (like unfaulted)?

src/fsw/FCCode/QuakeFaultHandler.hpp
tanishqaggarwal: I'm going to remove this class from here since #262 makes a mock anyway.

src/fsw/FCCode/QuakeFaultHandler.cpp
tanishqaggarwal: I added the "reset to unfaulted."

src/fsw/FCCode/QuakeFaultHandler.cpp
stewartaslan: In that case, how are checking that a power cycle happened? Is it a recorded event?

tanishqaggarwal: We should make this part of the set of "recorded events" when we address #278

PR #214, opened 2020-02-14
Shift flows

fatimayousuf:


src/fsw/FCCode/DownlinkProducer.cpp
```diff
@@ -243,5 +243,6 @@ void DownlinkProducer::swap_flow_priorities(unsigned char id1, unsigned char id2
     bool temp = flows[idx1].is_active;
     flows[idx1].is_active = flows[idx2].is_active;
     flows[idx2].is_active = temp;
-    std::swap(flows[idx1], flows[idx2]);
-}
+    flows.insert(flows.begin() + std::min(idx1, idx2), flows[std::max(idx1, idx2)]);
+    flows.erase(flows.begin()+std::max(idx1,idx2)+1);
```
tanishqaggarwal: Insert and erase cause memory operations on the heap and are dangerous to keep performing since they'll cause heap fragmentation.

Let's try to find a non structure-changing solution (you'll probably have to create some temporary Flow objects in between somehow)?

src/fsw/FCCode/DownlinkProducer.cpp
```diff
@@ -243,11 +243,27 @@ void DownlinkProducer::shift_flow_priorities(unsigned char id1, unsigned char id
     bool temp = flows[idx1].is_active;
     flows[idx1].is_active = flows[idx2].is_active;
     flows[idx2].is_active = temp;
-    flows.insert(flows.begin() + idx2, flows[idx1]);
+
+    std::vector<Flow> shifted_flows;
```
tanishqaggarwal: Creating a vector also causes heap allocation! :(

fatimayousuf: Noooooooooo :(((((((

fatimayousuf: Making temporary Flow objects wouldn't cause heap allocation but making a vector of flow objects would???

src/fsw/FCCode/DownlinkProducer.cpp
```diff
@@ -243,5 +243,15 @@ void DownlinkProducer::swap_flow_priorities(unsigned char id1, unsigned char id2
...
-}
+
+    if (idx1>idx2) {
+        for (size_t i = idx1; i > idx2; i--) {
+            std::swap(flows[i], flows[i-1]);
+        }
+    }
+    else if (idx2>idx1) {
+        for (size_t i = idx1; i < idx2; i++) {
+            std::swap(flows[i],flows[i+1]);
```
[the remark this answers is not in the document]
tanishqaggarwal: You should test for this kind of boundary-case behavior.

src/fsw/FCCode/DownlinkProducer.cpp
```diff
@@ -243,5 +243,15 @@ void DownlinkProducer::swap_flow_priorities(unsigned char id1, unsigned char id2
     bool temp = flows[idx1].is_active;
     flows[idx1].is_active = flows[idx2].is_active;
     flows[idx2].is_active = temp;
```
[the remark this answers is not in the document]
fatimayousuf: It was there before when we were just swapping flows. I just assumed that was something we would want to keep?

src/fsw/FCCode/DownlinkProducer.cpp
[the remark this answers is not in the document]
fatimayousuf: Ok two things:

Firstly, the function throws an error if the id of the flow is greater than or equal to the length of the flows vector. So for example, if I called shift(1,5) with the flow vector [1,2,3,4,5], I would get an error because the id of flow 5 is equal to the length of the flow vector. Also, no flow can have an id of 0, so the first flow has to have an id of at least 1. So I had assumed that we never would want to shift the last flow. 

After I saw your comment, I realized I assumed wrong, so I changed it so that shift_priorities() throws an error if an id is greater than the length of the vector, not greater than or equal to.

Second, after I made this change. I tested the boundary-case behaviour and the shift function worked fine...

src/fsw/FCCode/DownlinkProducer.cpp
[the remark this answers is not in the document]
tanishqaggarwal: Nah I think the flows should be shifted in their entirety. If one of the two flows is inactive we should keep it that way. Maybe I just had placeholder code here and I forgot about it...sorry!

src/fsw/FCCode/DownlinkProducer.cpp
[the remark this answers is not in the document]
tanishqaggarwal: Ah, you were way more careful than I gave you credit for. Nice, thanks!!

test/test_fsw_downlink_producer/test_downlink_producer.cpp
```diff
@@ -335,6 +335,56 @@ void test_downlink_changes() {
...
+    // Test shifting backwards
+    tf.downlink_producer->shift_flow_priorities(5,2);
+
+    // Get the new flow vector and check that the flows have been reordered as desired
+    std::vector<DownlinkProducer::Flow> flows=tf.downlink_producer->get_flows();
+    std::vector<int> desired_ids={1,5,2,3,4,6};
+    for (size_t i = 0; i<flows.size(); i++){
+        unsigned char flow_id;
+        flows[i].id_sr.deserialize(&flow_id);
+        TEST_ASSERT_EQUAL(desired_ids[i], flow_id);
```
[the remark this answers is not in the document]
tanishqaggarwal: Nevermind, I don't think this is necessary. If you're checking flow IDs that's a good sign that the flows are also correct.

test/test_fsw_downlink_producer/test_downlink_producer.cpp
```diff
@@ -335,6 +335,64 @@ void test_downlink_changes() {
...
+    // Check that the priorities are swapped, too
+    TEST_ASSERT_EQUAL(false, flows[1].is_active);
+    TEST_ASSERT_EQUAL(true, flows[4].is_active);
+
+    // Test shifting forwards
+    tf.downlink_producer->shift_flow_priorities(1,2);
+
+    // Get the new flow vector and check that the flows have been reordered as desired
+    flows=tf.downlink_producer->get_flows();
+    desired_ids={5,2,1,3,4,6};
```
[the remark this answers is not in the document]
tanishqaggarwal: It's less confusing now, thanks.

src/fsw/FCCode/DownlinkProducer.cpp
```diff
@@ -243,5 +243,15 @@ void DownlinkProducer::swap_flow_priorities(unsigned char id1, unsigned char id2
...
-    std::swap(flows[idx1], flows[idx2]);
-}
+
+    if (idx1>idx2) {
+        for (size_t i = idx1; i > idx2; i--) {
+            std::swap(flows[i], flows[i-1]);
```
tanishqaggarwal: I thought a bit more about the swap function and realized that it accepts rvalues as parameters:
https://en.cppreference.com/w/cpp/utility/move

This is bad news! flows[x] is fundamentally an lvalue, and since we haven't defined a move assignment operator for flows, the only way the compiler can convert flows[x] to an rvalue is to create a temporary copy.

Can we resolve this by implementing a move assignment operator for the flow? We'll have to unit test this operator too.

tanishqaggarwal: Sorry that I just thought about this.

tanishqaggarwal: I'm sorry! Swap accepts lvalues.

tanishqaggarwal: Talked offline that swap doesn't actually take in rvalues, but it still requires its parameters to be MoveAssignable.

src/fsw/FCCode/DownlinkProducer.hpp
```diff
@@ -103,9 +103,43 @@ class DownlinkProducer : public TimedControlTask<void> {
...
+
+        /**
+        * @brief Copy assignment operator.
+        */
+        Flow& operator=(const Flow& rhs) {
+            is_active = rhs.is_active;
+            id_sr = std::move(rhs.id_sr);
+            field_list = rhs.field_list;
+            return *this;
+        }
```
tanishqaggarwal: No need to implement this; we don't care if the copy constructor exists or not. (I think once you've defined the move constructor, the copy constructor becomes implicitly deleted by the compiler, which is actually a behavior that we want--we don't want to create a copy of the vector list of fields contained inside the flow.)

src/fsw/FCCode/DownlinkProducer.hpp
```diff
@@ -103,9 +103,43 @@ class DownlinkProducer : public TimedControlTask<void> {
...
+            *this = other;
+        }
+
+        /**
+        * @brief Move assignment operator.
+        */
+        Flow& operator=(Flow&& rhs) {
+            is_active = rhs.is_active;
+            id_sr = std::move(rhs.id_sr);
+            field_list = rhs.field_list;
```
tanishqaggarwal: Here the vector list is getting copied, which will cause heap operations. Instead, loop through the array and copy the fields over individually.

src/fsw/FCCode/DownlinkProducer.hpp
```diff
@@ -103,9 +103,43 @@ class DownlinkProducer : public TimedControlTask<void> {
...
+            return *this;
+        }
+
+        /**
+        * @brief Move constructor.
+        */
+        Flow(Flow&& other) {
+            *this = other;
+        }
```
tanishqaggarwal: No need to implement the move constructor if you've implemented the move assignment operator. The constructor gets automatically produced by the compiler once the assignment operator is constructed.

src/fsw/FCCode/DownlinkProducer.hpp
fatimayousuf: The flow vector is const. So every time downlink_producer->get_flows() is called, the flows are copied instead of moved. But since we are deleting the copy constructor, should I make the flow vector not const? Or is it better to just keep the copy assignment operator?

fatimayousuf: Every time we call downlink_producer.get_flows(), all the flows in the flow vector are copied. I think this is because get_flows() returns a const vector of flows. So should I keep the copy constructor?

tanishqaggarwal: Oh I see, it was already there. Ok let's keep it

PR #216, opened 2020-02-16
Prop controller

athena255:
Prop Controller implementation with basic features. No fault handling included.

---
This change is [Reviewable](https://reviewable.io/reviews/pathfinder-for-autonomous-navigation/flightsoftware/216)


src/fsw/FCCode/PropController.cpp
```diff
@@ -0,0 +1,339 @@
...
+    sched_valve3_f("prop.sched_valve3", Serializer<unsigned int>(999)),
+    sched_valve4_f("prop.sched_valve4", Serializer<unsigned int>(999))
+{
+
+    add_writable_field(prop_state_f);
+    add_writable_field(fire_cycle_f);
+    add_writable_field(sched_valve1_f);
+    add_writable_field(sched_valve2_f);
+    add_writable_field(sched_valve3_f);
+    add_writable_field(sched_valve4_f);
```
tanishqaggarwal: Does the prop controller not cache a GNC firing? This interface allows GNC to continually change its planned firings. This is an interface detail I've probably forgotten. I guess since you've already implemented this we can defer firing caching to another control task, maybe something called PropPlannerTask

test/test_fsw_prop_controller/test_prop_controller.cpp
```diff
@@ -0,0 +1,266 @@
...
+}
+
+void test_presurize_to_await_firing()
+{
+  PropTestFixture tf;
+  tf.set_state(prop_state_t::idle);
+  unsigned int cycles_until_fire = PropState_Pressurizing::num_cycles_needed();
+  tf.fixture_set_schedule(200, 200, 200, 200, cycles_until_fire);
+  // we may finish pressurizing earlier than 20 cycles, in this case, the 
+  // state should be in await_firing until the actual firing cycle (21)
```
tanishqaggarwal: Are you actually testing the "if pressurizing  finishes early, transition to await_firing" behavior here? It seems like you wait the maximum pressurization duration.

test/test_fsw_prop_controller/test_prop_controller.cpp
```diff
@@ -0,0 +1,266 @@
...
+
+void test_firing_to_idle()
+{
+  PropTestFixture tf;
+  tf.set_state(prop_state_t::idle);
+  unsigned int cycles_until_fire = PropState_Pressurizing::num_cycles_needed();
+  tf.fixture_set_schedule(700, 200, 200, 800, cycles_until_fire);
+  tf.step(cycles_until_fire);
+  tf.check_state(prop_state_t::await_firing);
+  // For 800/140 control cycles, we should be firing
```
tanishqaggarwal: control cycle count is 120 ms (in flight / radio-only testing environments) and 170 ms (in USB-connected testing environments.) A more accurate statement would be 800 ms / (control cycle duration)

test/test_fsw_prop_controller/test_prop_controller.cpp
```diff
@@ -0,0 +1,266 @@
...
+{
+  // Test that the prop system is in the state associated with pressurizing
+  PropTestFixture tf;
+  tf.set_state(prop_state_t::idle);
+  unsigned int cycles_until_fire = PropState_Pressurizing::num_cycles_needed();
+  tf.fixture_set_schedule(700, 200, 200, 800, cycles_until_fire);
+  tf.step();
+  tf.check_state(prop_state_t::pressurizing);
+  //TODO: check the prop system states associated with pressurizing
+}
```
tanishqaggarwal: Is this different from the earlier idle-to-pressurize test?

test/test_fsw_prop_controller/test_prop_controller.cpp
```diff
@@ -0,0 +1,266 @@
...
+  PropTestFixture tf;
+  tf.set_state(prop_state_t::idle);
+  unsigned int cycles_until_fire = PropState_Pressurizing::num_cycles_needed() + 5;
+  tf.fixture_set_schedule(700, 200, 200, 800, cycles_until_fire);
+  tf.step(cycles_until_fire - 5);
+  tf.check_state(prop_state_t::await_firing);
+  tf.step(5);
+  tf.check_state(prop_state_t::await_firing);
+
+  // TODO: check state associated with await_firing
```
tanishqaggarwal: This TODO is resolved right?

test/test_fsw_prop_controller/test_prop_controller.cpp
```diff
@@ -0,0 +1,266 @@
...
+
+void test_firing()
+{
+  PropTestFixture tf;
+  tf.set_state(prop_state_t::idle);
+  unsigned int cycles_until_fire = PropState_Pressurizing::num_cycles_needed() + 5;
+  tf.fixture_set_schedule(700, 200, 200, 800, cycles_until_fire);
+  tf.step(cycles_until_fire + 1);
+  tf.check_state(prop_state_t::firing);
+  // TODO: check state associated with firing
```
tanishqaggarwal: This TODO is resolved right?

src/fsw/FCCode/PropController.cpp
```diff
@@ -0,0 +1,339 @@
...
+}
+
+bool PropState_Pressurizing::can_pressurize_in_time()
+{
+    return controller->fire_cycle_f.get() >= num_cycles_needed();
+}
+
+unsigned int PropState_Pressurizing::num_cycles_needed()
+{
+    return ( PropController::max_pressurizing_cycles + 1 )
```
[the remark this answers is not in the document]
tanishqaggarwal: Tbh +1 is cleaner.

src/fsw/FCCode/PropController.cpp
tanishqaggarwal: Can you create an issue ticket for the PropPlannerTask

src/fsw/FCCode/PropController.hpp
```diff
@@ -0,0 +1,254 @@
...
+    }
+
+    // Return the PropState associated with the given prop_state_t
+    PropState& get_state(prop_state_t);
+    
+    // ------------------------------------------------------------------------
+    // Constants
+    // ------------------------------------------------------------------------
+    static constexpr unsigned int max_pressurizing_cycles = 20;
+    static constexpr float threshold_firing_pressure = 25;
```
tanishqaggarwal: I believe these constants should be made ground-modifiable.

src/fsw/FCCode/PropController.hpp
```diff
@@ -0,0 +1,254 @@
...
+class PropState_Idle : public PropState
+{
+public:
+    PropState_Idle() : PropState(prop_state_t::idle) {}
+    bool can_enter() override;
+    prop_state_t evaluate() override;
+
+    // Start pressurizing when we are within this amount of control cycles of the time
+    //      at which we want to fire
+    static constexpr unsigned int num_cycles_within_firing_to_pressurize = 50;
```
tanishqaggarwal: Should be made ground-modifiable (i.e. a state field.)

src/fsw/FCCode/PropController.hpp
```diff
@@ -0,0 +1,254 @@
...
+
+    // Number of control cycles needed to pressurize Tank2
+    //      (i.e. number of control cycles in 20 pressurizing cycles)
+    static unsigned int num_cycles_needed();
+    static constexpr unsigned int firing_duration_ms = 1000;
+    static constexpr unsigned int cooling_duration_ms = 10*1000;
+
+    // Number of control cycles needed to execute an entire pressurizing
+    //      cycle (1 + 10 = 11 seconds worth of control cycles)
+    static constexpr unsigned int ctrl_cycles_per_pressurizing_cycle =
```
tanishqaggarwal: Let's make these constants ground-modifiable.

src/fsw/FCCode/PropController.cpp
athena255: What do you mean by cache firing? Technically, the true schedule is only set at one point in time. I’d preferred, I can have that set earlier

src/fsw/FCCode/PropController.cpp
```diff
@@ -0,0 +1,339 @@
...
+        {
+            // TODO: what to do if tried to enter a state and was not allowed?
+        }
+    }
+    // Tick all the countdown timers
+    CountdownTimer::tick();
+
+    // Decrement fire_cycle if it is not equal to 0
+    if (fire_cycle_f.get() != 0)
+        fire_cycle_f.set(fire_cycle_f.get() - 1);
```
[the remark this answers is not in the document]
athena255: I can easily change to using GPS time. The rationale had something to do with 64 bit overflows or the nondeterminism of testing with micros().

src/fsw/FCCode/PropController.hpp
```diff
@@ -0,0 +1,254 @@
...
+private:
+
+    // ------------------------------------------------------------------------
+    // Static Components
+    // ------------------------------------------------------------------------
+    static PropState_Disabled state_disabled;
+    static PropState_Idle state_idle;
+    static PropState_Pressurizing state_pressurizing;
+    static PropState_AwaitFiring state_await_firing;
+    static PropState_Firing state_firing;
```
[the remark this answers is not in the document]
athena255: Only one instance of each of these components should ever exist.

src/fsw/FCCode/PropController.hpp
```diff
@@ -0,0 +1,254 @@
...
+
+// Forward declaration of PropState classes
+class PropState;
+class PropState_Disabled;
+class PropState_Idle;
+class PropState_Pressurizing;
+class PropState_AwaitFiring;
+class PropState_Firing;
+class PropState_Venting;
+class PropState_HandlingFault;
```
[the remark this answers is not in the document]
athena255: Making them ControlTasks might invite systems outside of PropController to believe that they may modify these components.

src/fsw/FCCode/PropController.hpp
```diff
@@ -0,0 +1,254 @@
...
+    void reset_timer();
+
+private:
+    // Number of cycles until this timer is free
+    size_t cycles_left = 0;
+
+    // Called by PropController to advance each timer in tick_list by one control cycle
+    static void tick(); 
+    // All instantiated timers are automatically added to this tick_list
+    static std::vector<CountdownTimer*> tick_list;
```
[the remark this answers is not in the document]
athena255: I intended for Firing Time start time use this as well. The vector list is a static list of pointers to CountdownTimers. One call to tick() will tick them all.

test/test_fsw_prop_controller/test_prop_controller.cpp
```diff
@@ -0,0 +1,266 @@
...
+  PropTestFixture tf;
+  tf.check_state(prop_state_t::disabled);
+}
+
+void test_disable()
+{
+  PropTestFixture tf;
+  // Prop should remain in disabled state until manually set to some other state
+  tf.fixture_set_schedule(200, 300, 400,500, 21);
+  // Firing time is set 5 control cycles from now but we should not fire since
```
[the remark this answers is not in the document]
athena255: I will check the state of the valves to make sure they are not open

test/test_fsw_prop_controller/test_prop_controller.cpp
```diff
@@ -0,0 +1,266 @@
...
+  tf.set_state(prop_state_t::idle);
+  // Prop should ignore if requested firing time is less than 21 cycles into the future
+  // since we will not have time to pressurize Tank1
+  unsigned int cycles_until_fire = PropState_Pressurizing::num_cycles_needed() - 1;
+  tf.fixture_set_schedule(200, 400, 800, 100, cycles_until_fire);
+  tf.step();
+  // State should remain in idle
+  tf.check_state(prop_state_t::idle);
+  tf.step(cycles_until_fire - PropState_Idle::num_cycles_within_firing_to_pressurize);
+  tf.check_state(prop_state_t::idle);
```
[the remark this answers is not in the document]
athena255: Yes, I agree with this.

test/test_fsw_prop_controller/test_prop_controller.cpp
athena255: So I expect pressurizing to take 20 cycles, but if it takes 5 cycles, then the state should transition to await_firing until we are nearly at the firing time. In this case, it will wait in await_firing until it is near time to fire

test/test_fsw_prop_controller/test_prop_controller.cpp
athena255: No

src/fsw/FCCode/PropController.hpp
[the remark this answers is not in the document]
tanishqaggarwal: Fair enough, pretty standard interface.

src/fsw/FCCode/PropController.hpp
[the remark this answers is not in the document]
tanishqaggarwal: This is fine but I think we should minimize static initialization of things to prevent a static initialization fiasco. Each of these classes expects a pointer to the PropController instance to be initialized later which makes me uneasy--shouldn't good design enforce safe initialization via constructor initializer lists?

Your code works, though. I just have very picky, personal design preferences. Feel free to ignore this.

src/fsw/FCCode/PropController.hpp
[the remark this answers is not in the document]
tanishqaggarwal: I don't think so, I think this just helps the code have the same pattern everywhere. For example QuakeManager owns a control task called QuakeController, MainControlLoop owns all control tasks, etc. The design goal behind control tasks was that one could contain another.

src/fsw/FCCode/PropController.cpp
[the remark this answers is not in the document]
tanishqaggarwal: GPS time is safe to overflows, I think.

src/fsw/FCCode/PropController.cpp
tanishqaggarwal: The GNC algorithm is constantly pumping out a recommended firing every time it's run (probably once every minute or so) but we want to just store one of its recommended firings and run the propulsion state machine sequence on it to completion. What do you mean by "the true schedule is only set at one point in time"?

test/test_fsw_prop_controller/test_prop_controller.cpp
tanishqaggarwal: I agree with that behavior, but (as far as I can tell) the test behavior has the PropController step into pressurizing, wait 19 cycles, and then step into await_firing.

Also, it's not clear to me how you're actually reducing the pressurization time requirement in the test fixture. Why does 19 cycles in "pressurizing" trigger "await_firing" and not 20 cycles or 5 cycles?

test/test_fsw_prop_controller/test_prop_controller.cpp
tanishqaggarwal: What would it mean to resolve this TODO? I'm not totally clear what's missing in the test step here.

src/fsw/FCCode/PropController.hpp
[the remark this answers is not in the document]
athena255: ControlTask is a nontrivial class and inheriting from it adds complexity. It is also a class that is difficult (impossible at the moment) to build without platformio. It makes no sense to inherit from them for no reason especially since I don't use any of the methods or fields they have. Considering how flaky Quake Manager is, I'd rather all components be as transparent and isolated as possible. This will make it easier to test and reason locally.

src/fsw/FCCode/PropController.cpp
```diff
@@ -0,0 +1,339 @@
...
+    prop_state_t next_state = get_state(current_state).evaluate();
+    if (next_state != current_state)
+    {
+        if ( get_state(next_state).can_enter() )
+        {
+            prop_state_f.set(static_cast<unsigned int>(next_state));
+        }
+        else
+        {
+            // TODO: what to do if tried to enter a state and was not allowed?
```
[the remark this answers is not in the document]
athena255: I have not thought of a nice way to handle faults without making the code look terrible. Switch statements have the potential to allow a transition between states that have nothing to do with each other. I'm planning to define fault states similar to the way states are defined right now. That would be the most structurally consistent and least complicated.

src/fsw/FCCode/PropController.hpp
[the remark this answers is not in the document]
athena255: There is no manual initialization associated with these objects. They are initialized like any other static variable.

> -shouldn't good design enforce safe initialization via constructor initializer lists?

No. First, it is trivial to check that the pointer to PropController is non zero before using it. And second, they are declared private, so no class outside of PropController may reference these components. Third, initializer list has nothing to do with architectural design, so I don't think we should make architectural decisions based on the preference to use initializer lists.

test/test_fsw_prop_controller/test_prop_controller.cpp
athena255: If I set the _firing_ time to 21, and PropController accepts this firing time, then 21 cycles later, PropController had better be firing. However, at 20 cycles, PropController should not be firing, but it should be done pressurizing. Therefore, it should be in awaiting_firing.

test/test_fsw_prop_controller/test_prop_controller.cpp
athena255: As you can see in the definition of cycles_firing, I end up using PAN::control_cycle_time_ms.

test/test_fsw_prop_controller/test_prop_controller.cpp
athena255: Not right now. But it should be, and that is why there's a TODO.

test/test_fsw_prop_controller/test_prop_controller.cpp
athena255: In await_firing, we want some conditions to be true: 
- Tank2 is at threshold pressure
- Tank1 valve is not currently open
- firing schedule and firing time is valid
- There are no other hardware faults.

test/test_fsw_prop_controller/test_prop_controller.cpp
athena255: This comment is not good. I see why you are confused. By state, I refer to conditions that must be true when we are in a given state. I don't mean whether prop_state_t has the expected value.

src/fsw/FCCode/PropController.hpp
athena255: A ground-modifiable constant is a global variable and not a constant. Allowing other subsystems the capability to modify these variables is rather liberal.

src/fsw/FCCode/PropController.cpp
```diff
@@ -0,0 +1,339 @@
...
+    }
+
+    return this_state;
+}
+
+bool PropState_Idle::is_time_to_pressurize() const
+{
+    bool is_within_pressurizing_time = controller->fire_cycle_f.get() < num_cycles_within_firing_to_pressurize + PropState_Pressurizing::num_cycles_needed();
+    bool is_schedule_valid = controller->validate_schedule();
+    bool has_enough_time = PropState_Pressurizing::can_pressurize_in_time();
```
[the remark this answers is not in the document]
athena255: PropState_Pressurizing has_enough_time() if it can pressurize before the requested fire_cycle.
PropState_Pressurizing is_within_pressurizing time if  PropState_Pressurizing should start pressurizing right now. 
Now that you bring it up though, I think this is redundant because PropState_Idle checks for this.

test/test_fsw_prop_controller/test_prop_controller.cpp
tanishqaggarwal: Right I'm just saying the comment could be updated

src/fsw/FCCode/PropController.hpp
tanishqaggarwal: Sorry, what I mean is the ability to modify a state field containing this constant at runtime. I think it's a good idea to allow the ground to be able to modify any parameters that affect software behavior.

src/fsw/FCCode/PropController.cpp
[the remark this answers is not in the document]
tanishqaggarwal: At this point you still need a switch statement to handle the various cases for the illegal transitions though, right?

src/fsw/FCCode/PropController.hpp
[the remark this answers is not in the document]
tanishqaggarwal: Sure let's just go with the current mechanism, esp since we know QuakeController broke tests

src/fsw/FCCode/PropController.cpp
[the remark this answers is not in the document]
athena255: Yes, I think you are right.

test/test_fsw_prop_controller/test_prop_controller.cpp
```diff
@@ -0,0 +1,310 @@
...
+    tf.set_state(prop_state_t::idle);
+    tf.fixture_set_schedule(200, 400, 12, 800, PropState_Pressurizing::num_cycles_needed() + 10);
+    tf.step();
+    tf.check_state(prop_state_t::await_pressurizing);
+    TEST_ASSERT_EQUAL(9, tf.execute_until_state_change()); // goes into pressurizing after 9 cycles in await_pressurizing
+    tf.check_state(prop_state_t::pressurizing);
+}
+
+void test_pressurize_to_firing() {
+    // There is no going from pressurizing into firing
```
tanishqaggarwal: I think this comment should be expanded more because otherwise it feels extraneous. Of course we can't transition from pressurizing to firing, but the reason is that we have to go to await firing in between, right?

test/test_fsw_prop_controller/test_prop_controller.cpp
```diff
@@ -0,0 +1,310 @@
...
+    tf.fixture_set_schedule(200, 200, 200, 200, PropState_Pressurizing::num_cycles_needed() + 1);
+    tf.step();
+    tf.check_state(prop_state_t::pressurizing);
+    // Same as above except we execute 18 pressurizing cylces
+    unsigned int expected_cycles =
+            (PropController::max_pressurizing_cycles - 2) * PropState_Pressurizing::ctrl_cycles_per_pressurizing_cycle +
+            2;
+    TEST_ASSERT_EQUAL(expected_cycles, tf.execute_until_state_change());
+    tf.check_state(prop_state_t::await_firing);
+}
```
tanishqaggarwal: Is this test testing anything new?

test/test_fsw_prop_controller/test_prop_controller.cpp
```diff
@@ -0,0 +1,310 @@
...
+    unsigned int cycles_until_fire = PropState_Pressurizing::num_cycles_needed() + 5;
+    tf.fixture_set_schedule(700, 200, 200, 800, cycles_until_fire);
+    TEST_ASSERT_EQUAL(1, tf.execute_until_state_change()); // await pressurizing
+    tf.execute_until_state_change(); // pressurizing
+    tf.execute_until_state_change(); // await_firing
+    tf.check_state(prop_state_t::await_firing);
+
+    // TODO: check state associated with await_firing
+    // valves should be closed
+    // schedule should be valid
```
tanishqaggarwal: Please resolve this TODO

test/test_fsw_prop_controller/test_prop_controller.cpp
```diff
@@ -0,0 +1,310 @@
...
+    PropTestFixture tf;
+    tf.set_state(prop_state_t::idle);
+    unsigned int cycles_until_fire = PropState_Pressurizing::num_cycles_needed() + 5;
+    tf.fixture_set_schedule(700, 200, 200, 800, cycles_until_fire);
+    tf.step(cycles_until_fire);
+    tf.check_state(prop_state_t::firing);
+    TEST_ASSERT_TRUE(PropulsionSystem.is_firing())
+    unsigned int expected_firing_cycles = (800 / PAN::control_cycle_time_ms) + 1; // roundup
+    TEST_ASSERT_EQUAL(expected_firing_cycles, tf.execute_until_state_change());
+    // can't really check valve states unless we want to make a fake interval timer
```
tanishqaggarwal: Yeah that's not worth it at this level of FSW

test/test_fsw_prop_controller/test_prop_controller.cpp
```diff
@@ -0,0 +1,310 @@
...
+    RUN_TEST(test_idle_to_pressurizing);
+    RUN_TEST(test_idle_to_await_pressurize);
+    RUN_TEST(test_await_pressurize_to_pressurize);
+    RUN_TEST(test_pressurize_to_firing);
+    RUN_TEST(test_pressurize_to_await_firing);
+    RUN_TEST(test_firing_to_idle);
+    RUN_TEST(test_pressurizing);
+    RUN_TEST(test_pressurize_fail);
+    RUN_TEST(test_await_firing);
+    RUN_TEST(test_firing);
```
tanishqaggarwal: Can you make the order of tests here reflect the order of the tests in the file?

src/fsw/FCCode/PropController.cpp
```diff
@@ -1,14 +1,23 @@
...
-{
+#ifdef FEELS_BAD
+#include <iostream>
+#define DD(f_, ...) printf((f_), ##__VA_ARGS__)
+#else
+#define DD(f_, ...)
+#endif
+#ifdef DESKTOP
+size_t g_fake_pressure_cycle_count = 15; // global
+#endif
```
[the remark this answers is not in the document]
athena255: Lol, I'll change the macro name. This is just a debug print. If #FEELS_BAD is defined then, DD is a printf. If it is not define, then DD is NOP. It exists for debugging test cases. g_fake_pressure_cycle_count also exists for test cases. I need a condition for when the "sensor pressure" should say that it is at threshold. By setting the global variable to 15, I am getting the sensor to report that the Tank2 is pressurized upon entering the 15th pressurizing cycle.

src/fsw/FCCode/PropController.cpp
```diff
@@ -23,73 +32,78 @@ PropController::PropController(StateFieldRegistry& registry, unsigned int offset
...
+
+    prop_state_t next_state = get_state(current_state).evaluate();
+    if (next_state != current_state) {
+        // sanity check
+        if (get_state(next_state).can_enter()) {
+            prop_state_f.set(static_cast<unsigned int>(next_state));
+            get_state(next_state).enter();
+        } else {
+            // This should never happen, if it does, complain a lot.
+            // TODO: enter handling fault maybe?
```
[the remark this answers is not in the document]
athena255: This should literally never ever happen because the enter condition can_enter() is called by the previous state. However, that assumes that no one modifies prop_state_t, and this is not true.

test/test_fsw_prop_controller/test_prop_controller.cpp
athena255: Absolutely!

g_fake_pressure_cycle_count = 20 is a boundary case. This test proves that the behavior of that boundary case is the same as the behavior of the normal case (g_fake_pressure_cycle_count = 19).

test/test_fsw_prop_controller/test_prop_controller.cpp
athena255: It is testing that the transition happens at the expected control cycle. It is testing the transition itself.

test/test_fsw_prop_controller/test_prop_controller.cpp
athena255: Sure. I have a script for this.

src/fsw/FCCode/PropController.cpp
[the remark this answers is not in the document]
tanishqaggarwal: You could make FEELS_BAD just be FUNCTIONAL_TEST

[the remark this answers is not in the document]
tanishqaggarwal: I think some comments explaining what's going on here will still be useful

src/fsw/FCCode/PropController.cpp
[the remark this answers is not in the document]
tanishqaggarwal: Oh ah, I forgot about ground commands to prop_state_t. Ok let's punt this handling to a future PR

test/test_fsw_prop_controller/test_prop_controller.cpp
tanishqaggarwal: Right! thanks for the clarification

src/fsw/FCCode/PropController.cpp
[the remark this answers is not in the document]
athena255: we agree to not make firings too far into the future

src/fsw/FCCode/PropController.hpp
athena255: This constant does not exist anymore

src/fsw/FCCode/PropController.cpp
```diff
@@ -23,73 +32,78 @@ PropController::PropController(StateFieldRegistry& registry, unsigned int offset
...
 
-bool PropController::validate_schedule()
-{
-    return is_valid_schedule(sched_valve1_f.get(), sched_valve2_f.get(), sched_valve3_f.get(), sched_valve4_f.get(), fire_cycle_f.get());
+bool PropController::validate_schedule() {
+    return is_valid_schedule(sched_valve1_f.get(), sched_valve2_f.get(), sched_valve3_f.get(), sched_valve4_f.get(),
+                             fire_cycle_f.get());
 }
 
 bool PropController::is_at_threshold_pressure()
```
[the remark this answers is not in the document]
athena255: this function had been made static

src/fsw/FCCode/PropController.cpp
tanishqaggarwal: Figured it out, this is all getting deferred to prop planner.

src/fsw/FCCode/PropController.cpp
```diff
@@ -1,14 +1,23 @@
...
-    sched_valve1_f("prop.sched_valve1", Serializer<unsigned int>(999)),
-    sched_valve2_f("prop.sched_valve2", Serializer<unsigned int>(999)),
-    sched_valve3_f("prop.sched_valve3", Serializer<unsigned int>(999)),
-    sched_valve4_f("prop.sched_valve4", Serializer<unsigned int>(999))
-{
+#ifdef FEELS_BAD
+#include <iostream>
+#define DD(f_, ...) printf((f_), ##__VA_ARGS__)
+#else
+#define DD(f_, ...)
```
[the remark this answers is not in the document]
tanishqaggarwal: Fixed

src/fsw/FCCode/PropController.cpp
```diff
@@ -215,88 +244,102 @@ bool PropState_Pressurizing::should_use_backup()
...
-    if (controller->is_at_threshold_pressure())
-    {
+    if (controller->is_at_threshold_pressure()) {
+        DD("\tTank2 is at threshold pressure!\n");
         PropulsionSystem.close_valve(Tank1, valve_num);
-        return prop_state_t::await_firing;
+        if (controller->can_enter_state(prop_state_t::await_firing)) {
+            return prop_state_t::await_firing;
+        } else {
+            // TODO: sure want to disable?
```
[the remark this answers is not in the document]
tanishqaggarwal: Nvm, this is a bad comment; we could easily enter this control flow branch if the ground changes the schedule willy-nilly. Let's think about this more carefully.

src/fsw/FCCode/PropController.cpp
```diff
@@ -141,72 +138,104 @@ void CountdownTimer::reset_timer()
...
+    bool was_await_pressurizing = controller->check_current_state(prop_state_t::await_pressurizing);
+    // Allow from idle because sometimes we can immediately pressurize
+    bool was_idle = controller->check_current_state(prop_state_t::idle);
+
+    return (was_await_pressurizing || was_idle) && is_time_to_pressurize();
+}
+
+bool PropState_Pressurizing::is_time_to_pressurize() {
+    // It is only time to pressurize when we are 20 pressurizing cycles away
+    return controller->fire_cycle_f.get() == num_cycles_needed() ;
```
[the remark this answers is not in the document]
tanishqaggarwal: Decided to rename to "cycles_until_firing".

src/fsw/FCCode/PropController.cpp
```diff
@@ -215,88 +244,102 @@ bool PropState_Pressurizing::should_use_backup()
...
 {
     // If 1 second has past since we opened the valve, then close the valve
     if ( countdown.is_timer_zero() )
     {
         PropulsionSystem.close_valve(Tank1, valve_num);
-        // Start cooldown timer
-        countdown.set_timer_ms(cooling_duration_ms);
+        countdown.set_timer_cc(ctrl_cycles_per_cooling_period);
     }
+    return this_state;
```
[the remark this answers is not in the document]
tanishqaggarwal: Got it, rationale is that handle_pressurize_failed() changes the state to disabled.

src/fsw/FCCode/PropController.cpp
```diff
@@ -307,33 +350,26 @@ bool PropState_Firing::is_schedule_empty() const
...
+    return ( was_pressurizing && PropController::is_at_threshold_pressure() && controller->validate_schedule() );
 }
 
-prop_state_t PropState_AwaitFiring::evaluate()
-{
-    if ( is_time_to_fire() )
-    {
+void PropState_AwaitFiring::enter() {
+    DD("==> entered PropState_AwaitFiring\n");
+    // TODO: maybe should be fire_cycle time?
```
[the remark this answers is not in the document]
tanishqaggarwal: Don't need this comment.

src/fsw/FCCode/PropController.cpp
```diff
@@ -307,33 +350,26 @@ bool PropState_Firing::is_schedule_empty() const
...
-    if ( !PropController::is_at_threshold_pressure() )
-        return false;
-    // there should be a schedule and a firing time
-    if ( !controller->validate_schedule() )
-        return false;
-    return true;
+bool PropState_AwaitFiring::can_enter() const {
+    bool was_pressurizing = controller->check_current_state(prop_state_t::pressurizing);
+
+    return ( was_pressurizing && PropController::is_at_threshold_pressure() && controller->validate_schedule() );
```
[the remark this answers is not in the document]
athena255: We decided that we should check the pressure again.

test/test_fsw_prop_controller/test_prop_controller.cpp
tanishqaggarwal: Comment was fixed, hence resolved.

src/fsw/FCCode/PropController.cpp
```diff
@@ -0,0 +1,388 @@
...
+          sched_valve2_f("prop.sched_valve2", Serializer<unsigned int>(999)),
+          sched_valve3_f("prop.sched_valve3", Serializer<unsigned int>(999)),
+          sched_valve4_f("prop.sched_valve4", Serializer<unsigned int>(999)),
+          // TODO: verify these Serializer paramemters
+          max_pressurizing_cycles("prop.max_pressurizing_cycles", Serializer<unsigned int>(50)),
+          threshold_firing_pressure("prop.threshold_firing_pressure", Serializer<float>(10, 50, 4)),
+          ctrl_cycles_per_filling_period("prop.ctrl_cycles_per_filling", Serializer<unsigned int>(50)),
+          ctrl_cycles_per_cooling_period("prop.ctrl_cycles_per_cooling", Serializer<unsigned int>(50)),
+          tank1_valve("prop.tank1_valve", Serializer<unsigned int>(1)),
+          // TODO: Why does Fault take a control_cycle_count reference?
```
tanishqaggarwal: It uses the control cycle count to increment the persistence of the failure condition. Also, I think the persistence should be 1 here so that the fault is signaled as soon as the last allowable prop pressure firing occurs.

src/fsw/FCCode/PropController.cpp
```diff
@@ -0,0 +1,405 @@
...
+            // case prop_state_t::handling_fault:
+            //     return state_handling_fault;
+        default:
+            return state_disabled;
+    }
+}
+
+bool PropController::is_valid_schedule(unsigned int v1, unsigned int v2, unsigned int v3, unsigned int v4,
+                                       unsigned int ctrl_cycles_from_now) {
+    return (v1 < 1000 && v2 < 1000 && v3 < 1000 && v4 < 1000 && ctrl_cycles_from_now > 1);
```
tanishqaggarwal: We should probably also add a check that v1, v2, v3, v4 > 0 (so that zero-length schedules are not considered valid.)

src/fsw/FCCode/PropController.cpp
```diff
@@ -0,0 +1,405 @@
...
+#define DD(f_, ...) printf((f_), ##__VA_ARGS__)
+size_t g_fake_pressure_cycle_count = 15; // global
+#else
+#define DD(f_, ...) do{} while(0)
+#endif
+
+PropController::PropController(StateFieldRegistry &registry, unsigned int offset)
+        : TimedControlTask<void>(registry, "prop", offset),
+          prop_state_f("prop.state", Serializer<unsigned int>(6)),
+          cycles_until_firing("prop.fire_cycle", Serializer<unsigned int>(256)),
```
tanishqaggarwal: Can we rename the string itself as well, maybe like "prop.cycles_till_fire"

src/fsw/FCCode/PropController.cpp
```diff
@@ -0,0 +1,449 @@
...
+}
+
+void PropState_Disabled::enter()
+{
+    DD("==> entered PropState_Disabled\n");
+}
+
+prop_state_t PropState_Disabled::evaluate() {
+    // Call disable here because we might have entered this due to some problem
+    if (PropulsionSystem.is_functional())
```
stewartaslan: Didn't we decide that there's no way to tell if prop system is functional? So should we just go ahead and disable here no matter what?

src/fsw/FCCode/PropController.cpp
```diff
@@ -0,0 +1,449 @@
...
+        valve_num = 1;
+    // Reset the pressurizing cycles count to 0
+    pressurizing_cycle_count = 0;
+    // Reset timer to 0 (just in case)
+    countdown.reset_timer();
+}
+
+bool PropState_Pressurizing::should_use_backup()
+{
+    // TODO: valve choice is determined by the ground, is this ok?
```
stewartaslan: Yep this is good, I think we can remove this TODO

src/fsw/FCCode/PropController.cpp
tanishqaggarwal: Small (< 10 ms) firings are ignored anyway

src/fsw/FCCode/PropController.cpp
```diff
@@ -0,0 +1,339 @@
...
+        return false;
+    return true;
+}
+
+prop_state_t PropState_AwaitFiring::evaluate()
+{
+    if ( is_time_to_fire() )
+    {
+        // Copy the schedule values from the registry into Tank2
+        controller->write_tank2_schedule();
```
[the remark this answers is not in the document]
athena255: Schedule will only fire if cycles_until_firing is 0

PR #226, opened 2020-02-24
Piksi Upgrades

shihaocao:
- Added ability to mute Piksi by setting relevant outputs to NaN when Quake is in transceive mode
- Added set relevant outputs to NaN whenever no data is received (as opposed to leaving last output)
- Replaced piksi_state_t::dead with a Fault

---
This change is [Reviewable](https://reviewable.io/reviews/pathfinder-for-autonomous-navigation/flightsoftware/226)


test/test_fsw_piksi_control_task/test_piksi_control_task.cpp
```diff
@@ -268,6 +297,40 @@ void test_dead(){
...
+    tf.set_vel_ecef(tow, vel);
+    tf.set_baseline_ecef(tow, baseline);
+    tf.set_baseline_flag(1);
+    
+    // set the quake mode from last cycle as transceive
+    tf.radio_state_fp->set(static_cast<unsigned char>(radio_state_t::transceive));
+
+    tf.execute();
+    // times should now agree, and be in baseline
+    assert_piksi_mode(piksi_mode_t::fixed_rtk);
```
tanishqaggarwal: This behavior is incorrect, right? We expect the Piksi to go into no_fix since the data jumps to NaN and data mute is on?

src/fsw/FCCode/PiksiControlTask.cpp
```diff
@@ -12,7 +13,11 @@ PiksiControlTask::PiksiControlTask(StateFieldRegistry &registry,
     current_state_f("piksi.state", Serializer<unsigned int>(4)),
     fix_error_count_f("piksi.fix_error_count", Serializer<unsigned int>(1001)),
     time_f("piksi.time", Serializer<gps_time_t>()),
-    last_fix_time_f("piksi.last_fix_time")
+    last_fix_time_f("piksi.last_fix_time"),
+    bool_sr(),
+    data_mute_f("piksi.data_mute", bool_sr),
+    //if we haven't had a good reading in ~120 seconds the piksi is probably dead
+    piksi_fault("piksi.fault", 1000, control_cycle_count)
```
tanishqaggarwal: Need to add the fault to the registry

src/fsw/FCCode/PiksiControlTask.cpp
```diff
@@ -41,37 +59,32 @@ void PiksiControlTask::execute()
     //3 means CRC error on serial
     //5 means timing error exceed
     if(read_out == 3|| read_out == 4|| read_out == 5)
-        fix_error_count_f.set(fix_error_count_f.get() + 1);
+        piksi_fault.signal();
```
tanishqaggarwal: Rather than having a bunch of distributed signals() and unsignals(), I think it would be better to keep a running flag of whether or not Piksi should be signaled, and then have one if-else block at the end. In my belief this gives a better face-value guarantee that the Piksi fault is either signaled or unsignaled on every control cycle.

test/test_fsw_piksi_control_task/test_piksi_control_task.cpp
```diff
@@ -248,26 +280,63 @@ void test_dead(){
         TEST_ASSERT_FLOAT_WITHIN(0.1,mag_2(baseline),mag_2(tf.baseline_fp->get()));
 
         //simulate that the piksi is not sending any data for 1000 control cycles.
-        //Make sure that the counter state fields are set correctl.
+        //Make sure that the counter state fields are set correct.
```
tanishqaggarwal: correctly

test/test_fsw_piksi_control_task/test_piksi_control_task.cpp
```diff
@@ -248,26 +280,63 @@ void test_dead(){
...
-        assert_piksi_mode(piksi_mode_t::dead);
+        TEST_ASSERT(tf.piksi_fault_fp->is_faulted());
+        assert_piksi_mode(piksi_mode_t::no_data_error);
+
+        // one good execution releases the fault
+        tf.set_read_return(1);
+        TimedControlTaskBase::control_cycle_count++;
+        tf.execute();
+        assert_piksi_mode(piksi_mode_t::fixed_rtk);
+        TEST_ASSERT_FALSE(tf.piksi_fault_fp->is_faulted());
```
tanishqaggarwal: It should be clearer, in some preamble comments, that this chunk of code will test the fault flag.

src/fsw/FCCode/PiksiControlTask.cpp
shihaocao: Sounds good. Will do

test/test_fsw_piksi_control_task/test_piksi_control_task.cpp
shihaocao: To answer, i wanted to not let muting mask underlying values as much as possible. However, all this code will be replaced by changes to ADCSEstimator.

tanishqaggarwal:
This PR is so old it uses Travis. Can I suggest we close this PR until we're ready to address this properly?

shihaocao:
Lmao yeah, this PR is very dead, will revive in a bit

PR #227, opened 2020-02-26
Extend sf registry

fatimayousuf:


test/test_fsw_state_field_registry_mock/test_fsw_state_field_registry_mock.cpp
```diff
@@ -149,6 +149,17 @@ void test_create_writable_vector_field_args() {
     TEST_ASSERT_NOT_NULL(registry.find_writable_field("foo2"));
 }
 
+void test_create_fault() {
+    StateFieldRegistryMock registry;
+
+    registry.create_fault("foo", 1, 3);
+    TEST_ASSERT_NOT_NULL(registry.find_fault("foo"));
```
tanishqaggarwal: Let's also check that "foo" exists as a writable field?

test/test_fsw_state_field_registry_mock/test_fsw_state_field_registry_mock.cpp
```diff
@@ -164,11 +175,15 @@ void test_clear() {
     registry.create_internal_field<unsigned int>("foo3");
     TEST_ASSERT_NOT_NULL(registry.find_internal_field("foo3"));
 
+    registry.create_fault("foo4", 1, 300);
+    TEST_ASSERT_NOT_NULL(registry.find_fault("foo4"));
+
     registry.clear();
     TEST_ASSERT_NULL(registry.find_internal_field("foo"));
```
tanishqaggarwal: This as well should check that foo is writable, not internal.

test/test_fsw_state_field_registry_mock/test_fsw_state_field_registry_mock.cpp
tanishqaggarwal: Note that Fault derives from WritableStateField<bool> so foo should be a writable field.

test/StateFieldRegistryMock.hpp
```diff
@@ -292,26 +301,43 @@ class StateFieldRegistryMock : public StateFieldRegistry {
         return field_ptr;
     }
 
+    /**
+     * @brief Create a fault and add it to the registry.
+     * 
+     * @param name Name of fault to create.
+     * @return Pointer to fault that was created.
+     */
+    std::shared_ptr<Fault> create_fault(const std::string& name, const size_t _persistance, unsigned int control_cycle_count)
```
tanishqaggarwal: Persistence is spelled with an e not an a

test/test_fsw_state_field_registry_mock/test_fsw_state_field_registry_mock.cpp
```diff
@@ -149,6 +149,17 @@ void test_create_writable_vector_field_args() {
     TEST_ASSERT_NOT_NULL(registry.find_writable_field("foo2"));
```
tanishqaggarwal: Can you also add tests to the test_common_state_field_registry? It's important that the base-level functionality of adding and finding faults (i.e. without StateFieldRegistryMock) is tested, or flight software may break.

tanishqaggarwal: I think there were some segfaults when I tried doing it. Have fun 🙃

test/StateFieldRegistryMock.hpp
fatimayousuf: How very...meticulous

test/test_fsw_state_field_registry_mock/test_fsw_state_field_registry_mock.cpp
fatimayousuf: image

test/test_fsw_state_field_registry_mock/test_fsw_state_field_registry_mock.cpp
fatimayousuf: I'll change it, but I think you wrote that, not me lol

PR #232, opened 2020-02-29
add docking timeout

fatimayousuf:
This change is [Reviewable](https://reviewable.io/reviews/pathfinder-for-autonomous-navigation/flightsoftware/232)


src/fsw/FCCode/DockingController.cpp
```diff
@@ -5,16 +5,19 @@ DockingController::DockingController(StateFieldRegistry &registry, unsigned int
     : TimedControlTask<void>(registry, "docking_ct", offset), docksys(_docksys),
       docked_f("docksys.docked", Serializer<bool>()),
       dock_config_f("docksys.dock_config", Serializer<bool>()),
-      is_turning_f("docksys.is_turning", Serializer<bool>())
+      is_turning_f("docksys.is_turning", Serializer<bool>()),
```
tanishqaggarwal: The docking state timeout has nothing to do with the docking controller; the docking controller is just a hardware controller, but the docking state is managed entirely inside MIssionManager. I'd recommend moving this flag into MissionManager

src/fsw/FCCode/MissionManager.cpp
```diff
@@ -19,6 +20,8 @@ MissionManager::MissionManager(StateFieldRegistry& registry, unsigned int offset
     docking_trigger_dist_f("trigger_dist.docking", Serializer<double>(0, 100, 10)),
     max_radio_silence_duration_f("max_radio_silence",
         Serializer<unsigned int>(2 * PAN::one_day_ccno)),
+    docking_timeout_limit_f("docking_timeout_limit",
+        Serializer<unsigned int>(2 * PAN::one_day_ccno)),
```
tanishqaggarwal: Thanks for making this wait configurable

src/fsw/FCCode/MissionManager.cpp
```diff
@@ -269,6 +275,12 @@ void MissionManager::dispatch_leader_close_approach() {
 void MissionManager::dispatch_docking() {
     docking_config_cmd_f.set(true);
 
+    if(too_long_in_docking() && !docked_fp->get()) {
+        transition_to_state(mission_state_t::standby,
+            adcs_state_t::startup,
+            prop_state_t::disabled);
+    }
+
```
tanishqaggarwal: To make the logic slightly more explicit, I'd recommend replacing lines 278-291 as an if-elseif-else statement, with the if-statement running if the docking logic is pressed, the elseif-statement running if too long in docking, and the else statement running no behavior. It makes the logic here just a tad more explicit since a code reader doesn't have to parse through any boolean logic.

test/test_fsw_mission_manager/test_mission_manager.cpp
```diff
@@ -197,6 +198,19 @@ void test_dispatch_docking() {
...
+    // Check that the system is not docked
+    TEST_ASSERT_FALSE(tf2.docked_fp->get());
+
+    // Let a full day pass without docking
+    tf2.set_ccno(PAN::one_day_ccno);
+    tf2.step();
+    TEST_ASSERT_FALSE(tf2.docked_fp->get());
+
+    // Check that mission manager moves to standy
+    tf2.check(mission_state_t::standby);
```
tanishqaggarwal: Also should test
- The state does not change if there's been less than 1 day since the docking switch being pressed
- If a significant amount of time passes and then the docking switch is pressed, the state still transitions to docked

test/test_fsw_mission_manager/test_mission_manager.cpp
```diff
@@ -200,17 +200,35 @@ void test_dispatch_docking() {
...
+    tf2.step();
 
     // Check that the system is not docked
     TEST_ASSERT_FALSE(tf2.docked_fp->get());
 
-    // Let a full day pass without docking
-    tf2.set_ccno(PAN::one_day_ccno);
+    // Let a half day pass
+    tf2.set_ccno(MissionManager::control_cycle_count+0.5*PAN::one_day_ccno);
+    tf2.step();
```
tanishqaggarwal: How about at PAN::one_day_ccno - 1? Does it stay in docked?

PR #237, opened 2020-03-01
Added better control task tests

tanishqaggarwal:
I recognize I'm not testing failure behavior for the find/add functions, but this is currently impossible to test since failures cause a dynamic assertion failure. Any suggestions on making the behavior more testable?

### DIFF
--- a/src/fsw/FCCode/ControlTask.hpp
+++ b/src/fsw/FCCode/ControlTask.hpp
@@ -47,6 +47,7 @@ class ControlTask : protected debug_console {
   protected:
     StateFieldRegistry& _registry;
 
+  private:
     void check_field_added(const bool added, const std::string& field_name) {
         if(!added) {
             #ifdef UNIT_TEST
@@ -66,6 +67,12 @@ class ControlTask : protected debug_console {
             assert(false);
         }
     }
+
+  #ifdef UNIT_TEST
+  public:
+  #else
+  protected:
+  #endif
 
     template<typename U>
     void add_internal_field(InternalStateField<U>& field) {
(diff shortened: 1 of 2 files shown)


src/fsw/FCCode/ControlTask.hpp
```diff
@@ -119,37 +128,44 @@ class ControlTask : protected debug_console {
...
-        return *event_ptr;
+        return event_ptr;
     }
 
     Fault* find_fault(const char* fault, const char* file, const unsigned int line) {
-        auto fault_ptr = _registry.find_fault(fault);
+        Fault* fault_ptr = _registry.find_fault(fault);
         check_field_exists(fault_ptr, "fault", fault);
-        return *fault_ptr;
+        return fault_ptr;
```
tanishqaggarwal: This asterisk is what broke things @fatimayousuf

fatimayousuf: Oh yikes sorry about that

src/fsw/FCCode/ControlTask.hpp
```diff
@@ -67,6 +68,12 @@ class ControlTask : protected debug_console {
         }
```
fatimayousuf: This is just an idea. Could we put a "return;" statement on Line 56? That way, the function would still print the message, but it wouldn't be able to throw an assertion error. And if there is a way to check that "Field \"" << field_name << "\" is already in the registry." is being printed in a Desktop Unit Test, then we would also know for sure if an assertion error _would_ be thrown.

src/fsw/FCCode/ControlTask.hpp
tanishqaggarwal: No worries! I didn't catch it either when I reviewed your PR so it's on me as well

src/fsw/FCCode/ControlTask.hpp
tanishqaggarwal: Good idea! I'm still not sure how to catch prints inside a Unity framework unit test but I'll keep looking, it's a useful thing to have.

PR #238, opened 2020-03-02
Better initialization error handling

tanishqaggarwal:
This PR vastly improves the error handling infrastructure on Flight Software. Here are the important features of this PR:

1. Implementation of the `pan_assert` function, which throws an exception if running in a unit test, and causes a program crash if running in HOOTL/HITL. It has no behavior in flight. This function lets us test invalid program behavior inside the unit test.
2. A test for `pan_assert` (test_common_assertion) to demonstrate this behavior.
3. An invalid initialization test for the DockingController to show how pan_assert can be used inside unit tests.
4. Modifications of ControlTasks to include noexcept specifiers. I know that noexcept is not compile-time enforced in C++, which is a little unfortunate since it would be great if we could compile-time check that runtime functions cannot throw exceptions. Perhaps there exist static analysis tools that we can use for this purpose.
5. I commented out calls to assert() that occur during the control loop (in DownlinkProducer.cpp and QuakeManager.cpp), because they're kind of dangerous. We should instead try to log the error behavior either to an event or a state field so that we can understand what happened from the ground.

As a team, if we want to use this feature to broaden our tests, will have to create additional PRs to add invalid initialization testing to a bunch of FSW unit tests.

---
This change is [Reviewable](https://reviewable.io/reviews/pathfinder-for-autonomous-navigation/flightsoftware/238)


src/fsw/FCCode/ADCSCommander.cpp
```diff
@@ -28,7 +28,7 @@ ADCSCommander::ADCSCommander(StateFieldRegistry& registry, unsigned int offset)
     imu_mag_filter_f("adcs_cmd.imu_mag_filter", filter_sr),
     imu_gyr_filter_f("adcs_cmd.imu_gyr_filter", filter_sr),
     imu_gyr_temp_filter_f("adcs_cmd.imu_gyr_temp_filter", filter_sr),
-    k_sr(std::numeric_limits<float>::min(), std::numeric_limits<float>::max(),16),
+    k_sr(0,100,16),
```
tanishqaggarwal: I had to make this change because the min <= max comparison in the Serializer constructor was throwing an exception. Since we have to review all serializer constants anyway, I think this is OK.

src/common/assertion.hpp
```diff
@@ -0,0 +1,51 @@
...
+    if (!(condition)) {
+        #if defined(UNIT_TEST)
+            throw EXCEPTION(err);
+        #elif defined(FUNCTIONAL_TEST)
+            #ifdef DESKTOP
+                std::cout << "Error: " << err << std::endl;
+            #else
+                Serial.printf("Error: %s", err);
+            #endif
+            assert(false);
```
nhz2: What does assert do on teensy? Maybe it should just loop, printing the error message/blinking an LED forever?

tanishqaggarwal: On HITL/HOOTL it just prints the error. On flight it does nothing.

I feel like blinking an LED is probably a good idea since we want software to get "crashed". I'll put that in.

tanishqaggarwal:
We determined this is an extreme nice-to-have.

PR #239, opened 2020-03-02
Constants for prop driver sensors

tanishqaggarwal:
We've computed and added in the regressions.

The regression for pressure is a piecewise linear graph, as we've always assumed. The regression for temperature ended up looking like
y = A ln(x)^K + B
for constants A, K, B.

Evidence of regressions:
https://cornellprod-my.sharepoint.com/:x:/g/personal/saa243_cornell_edu/Edg-vGek6SBEoe0jhIUxmnIBPh84Y6g_Tro-SJWonhuVhA?e=EewWzU

### DIFF
--- a/src/fsw/FCCode/Drivers/PropulsionSystem.cpp
+++ b/src/fsw/FCCode/Drivers/PropulsionSystem.cpp
@@ -83,8 +83,13 @@ void _Tank2::setup()
 
 int Tank::get_temp() const
 {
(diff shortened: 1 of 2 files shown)


tanishqaggarwal:
Closing until we can add:
- A hardware unit test for verifying that the regressions produce reasonable values
- Functional ptest for checking that these values are available at a global level within FlightSoftware

PR #240, opened 2020-03-03
CPPCheck passthrough 1

tanishqaggarwal:
I passed our codebase through CPPCheck and fixed some things that I thought were bad. Notable changes:
1) A shift to u?int(8|16|32)_t for integer types. It makes it clearer what the size of the actual object is that we're dealing with.
2) Travis changes to reduce the build times (we don't need to be building all ADCS test scripts and both leader AND follower environments for Teensy)
3) Default initializations for many class members (particularly inside drivers)
4) Removed __FILE__ and __LINE__ constructs from find_x_field functions

It looks like a massive PR but it's really just the same kinds of changes over a large number of files.


src/fsw/FCCode/ADCSBoxMonitor.cpp
```diff
@@ -95,16 +95,16 @@ ADCSBoxMonitor::ADCSBoxMonitor(StateFieldRegistry &registry,
         add_fault(wheel_pot_fault);
     }
 
-bool exceed_bounds(const std::array<float, 3>& input, const float min, const float max){
```
tanishqaggarwal: I decided to remove const qualifiers from parameters that don't really need them, so that the code looks a little cleaner.

tanishqaggarwal:
Seems like some recent changes broke the build.

PR #249, opened 2020-03-06
Deployment to Standby testcase

tanishqaggarwal:
To run, invoke the command

    python -m ptest.run_simulation -c ptest/configs/fc_only_native.json -t DeploymentToStandby

from within the psim/python directory. Be sure that you've activated the virtual environment and edited ptest/configs/fc_only_native.json to point to the correct location of FlightSoftware (or, if you're on Windows, use fc_only_teensy.json and then modify the USB port to which the Teensy is connected)

---
This change is [Reviewable](https://reviewable.io/reviews/pathfinder-for-autonomous-navigation/flightsoftware/249)


ptest/cases/base.py
```diff
@@ -1,91 +1,196 @@
...
+            "follower_close_approach",
+            "leader_close_approach",
+            "docking",
+            "docked",
+            "safehold",
+            "manual"
+        ])
+
+        self.prop_states = FSWEnum([
+            "disabled",
```
athena255: this is wrong order, see prop_state_t.enum

ptest/cases/base.py
```diff
@@ -1,91 +1,196 @@
...
+            "zero_torque",
+            "zero_L",
+            "detumble",
+            "point_manual",
+            "point_standby",
+            "point_docking"
+        ])
+
+        self.radio_states = FSWEnum([
+            "disabled",
```
athena255: this is wrong order

tanishqaggarwal: Which item is incorrect?

tanishqaggarwal: Ok, I know that this comment is still an open item, but we can quickly resolve this. I'm going to merge this into master so that i can clean up and modernize the prop_driver_case branch

PR #251, opened 2020-03-07
HITL Band-Aid

shihaocao:
- Draft PR

---
This change is [Reviewable](https://reviewable.io/reviews/pathfinder-for-autonomous-navigation/flightsoftware/251)


ptest/state_session.py
```diff
@@ -189,6 +189,20 @@ def _write_state_basic(self, fields, vals, timeout = None):
         for field in fields:
             returned_vals.append(self._wait_for_state(field, timeout))
 
+        if returned_vals[0] is None:
```
tanishqaggarwal: There's a feature in flight software (in QuakeManager) that can also return telemetry packets #157. Can you make sure that this new code doesn't break that feature?

src/fsw/FCCode/ADCSBoxController.cpp
```diff
@@ -93,14 +93,29 @@ void ADCSBoxController::execute(){
...
+        bool disable_get = havt_cmd_disable_vector_fp[idx]->get();
+        if(disable_get){
+            temp_cmd_table.set(idx, disable_get);
+            send_cmd_table = true;
+        }
     }
-    adcs_system.set_havt_disable(temp_cmd_table);
+    
+    if(send_cmd_table)
+        adcs_system.set_havt_disable(temp_cmd_table);
```
tanishqaggarwal: Can you add some comments for this logic? It's a little hard to follow at this point.

src/fsw/FCCode/ADCSBoxMonitor.cpp
```diff
@@ -129,7 +129,8 @@ void ADCSBoxMonitor::execute(){
     float gyr_temp = 0.0;
 
     //ask the driver to fill in values
-    adcs_is_functional.set(adcs_system.is_functional());
+    adcs_is_functional.set(adcs_system.i2c_ping());
```
tanishqaggarwal: Did you switch this to get rid of the persistent err_count behavior?

src/fsw/FCCode/DebugTask.cpp
```diff
@@ -16,8 +16,8 @@ DebugTask::DebugTask(StateFieldRegistry &registry, unsigned int offset)
 
 void DebugTask::execute() {
 #ifdef FUNCTIONAL_TEST
-  start_cycle_f.set(false);
-  while (!start_cycle_f.get())
+  // start_cycle_f.set(false);
+  // while (!start_cycle_f.get())
```
tanishqaggarwal: Can you put this back in

ptest/state_session.py
```diff
@@ -189,6 +189,20 @@ def _write_state_basic(self, fields, vals, timeout = None):
...
+            return False
+
+        returned_vals = returned_vals[0].split(",")
+        returned_vals = [x for x in returned_vals if x is not ""]
+        
+        if (returned_vals[0].replace('.','')).isnumeric():
```
kkrol27: Negative numbers here?

src/adcs/havt_devices.hpp
```diff
@@ -36,6 +36,7 @@ enum Index : unsigned char {
   SSA_ADC1,
   SSA_ADC2,
   SSA_ADC3,
+
```
kkrol27: This extra line intentional?

src/adcs/rwa.cpp
```diff
@@ -196,6 +196,8 @@ void actuate(unsigned char rwa_mode, lin::Vector3f rwa_cmd) {
       // Set wheel configurations
       for (unsigned int i = 0; i < 3; i++) {
         if (wheels[i].is_functional()) {
+          LOG_INFO_header
+          LOG_INFO_println("wheel act: " + String(rwa_cmd(i)))
```
kkrol27: Wondering here if we really want this at the `INFO` logging level. I don't see why we wouldn't just set the `rwa` module itself to the `TRACE` level and let everything else be?

src/adcs/state_controller.cpp
```diff
@@ -17,6 +17,10 @@
   #define LOG_LEVEL UMB_LOG_LEVEL
 #endif
 
+#undef LOG_LEVEL
+#define LOG_LEVEL 0
+// ^ The above is a fix to prevent serial debug output during the i2c interrupt.
```
kkrol27: This is fine for now. In the future though we should probably just zero out the logging level in the config file and call it good.

src/adcs/havt_devices.hpp
shihaocao: Yep just helps us count devices

src/fsw/FCCode/ADCSBoxMonitor.cpp
shihaocao: is_functional() is tied to the error_count field which is latching. So I changed it to i2c_ping() so that it would always reflect the true functionality of the ADCSC in the current moment (which I think is more desirable).

ptest/state_session.py
```diff
@@ -195,7 +195,7 @@ def _write_state_basic(self, fields, vals, timeout = None):
         returned_vals = returned_vals[0].split(",")
         returned_vals = [x for x in returned_vals if x is not ""]
         
-        if (returned_vals[0].replace('.','')).isnumeric():
+        if (returned_vals[0].replace('.','').replace('-','')).isnumeric():
```
tanishqaggarwal: Why was this needed? Are negative numbers not numeric?

src/adcs/state_controller.cpp
shihaocao: Should be cleaner now, realized can use #undef to fix compilation errors for log level selection

src/adcs/rwa.cpp
shihaocao: Oops left that in accidentally

ptest/state_session.py
shihaocao: Fixed

ptest/state_session.py
shihaocao: Verified over PM that seems like still working

ptest/state_session.py
shihaocao: isnumeric() is true iff all the string characters contain one of 1234567890

ptest/cases/spin_motors_case.py
```diff
@@ -0,0 +1,13 @@
+# SpinMotorsCase. Gets satellite read to spin motors.
+from .base import SingleSatOnlyCase
+
+class SpinMotorsCase(SingleSatOnlyCase):
+    def setup_case_singlesat(self):
+        self.sim.flight_controller.write_state("pan.state", 11) # Mission State = Manual
+        self.sim.flight_controller.write_state("adcs.state", 5) # ADCS State = Manual
+        self.sim.flight_controller.write_state("adcs_cmd.rwa_mode", 1) # Speed Control
+        self.sim.flight_controller.write_state("adcs_cmd.rwa_speed_cmd", "0, 0, 0") # 0 speed to begin with
```
kkrol27: It'd be cool if we could potentially expand the `write_state` function to take `*args` instead of just one `arg`. Separate issue though for sure.

ptest/cases/spin_motors_case.py
```diff
@@ -0,0 +1,13 @@
...
+        self.sim.flight_controller.write_state("dcdc.ADCSMotor_cmd", "true")
```
kkrol27: Also potentially support for booleans?

src/adcs/main.cpp
```diff
@@ -216,6 +216,7 @@ void update_ssa() {
 
 #if LOG_LEVEL >= LOG_LEVEL_INFO
 static unsigned long cycles = 0;
+static unsigned long last_info_time = millis();
```
kkrol27: Working off time is definitely a better move!

shihaocao: :)

ptest/cases/spin_motors_case.py
shihaocao: yeah true, that would be a good upgrade.

PR #252, opened 2020-03-07
Install script

nhz2:
This is mostly Athena's `install.sh` script. I also added some try except blocks around opening the radio and flask key files.

---
This change is [Reviewable](https://reviewable.io/reviews/pathfinder-for-autonomous-navigation/flightsoftware/252)

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -5,7 +5,7 @@ source venv/bin/activate
 pip install -r requirements.txt
 matlab -batch '"'"pyversion $(pwd)/venv/bin/python"'"' -nodisplay -nojvm -nosplash
 export PYTHONPATH=$(pwd)/venv
-if [ -d "/home/pan/MATLAB/R2019b/extern/engines/python" ]
+if [ -d "$HOME/MATLAB/R2019b/extern/engines/python" ]
 then
 	cd /home/pan/MATLAB/R2019b/extern/engines/python
 else
@@ -15,4 +15,3 @@ python setup.py install
 cd -
 git submodule update --init --recursive
 pio run -e fsw_native_leader
-
--- a/install.sh
+++ b/install.sh
@@ -1,0 +1,18 @@
+#!/bin/bash
+
+python3.7 -m venv venv
+source venv/bin/activate
+pip install -r requirements.txt
+matlab -batch '"'"pyversion $(pwd)/venv/bin/python"'"' -nodisplay -nojvm -nosplash
+export PYTHONPATH=$(pwd)/venv
+if [ -d "/home/pan/MATLAB/R2019b/extern/engines/python" ]
+then
+	cd /home/pan/MATLAB/R2019b/extern/engines/python
+else
+	cd /Applications/MATLAB_R2019b.app/extern/engines/python
+fi
+python setup.py install
+cd -
+git submodule update --init --recursive
+pio run -e fsw_native_leader
+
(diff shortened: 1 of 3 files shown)


ptest/run_simulation.py
```diff
@@ -269,39 +269,46 @@ def stop_all(self, reason_for_stop, is_error = True):
...
+                if not v.validate(flask_keys_config, flask_keys_schema):
+                    print(
+                        "Malformed flask keys file. The following errors were found. Exiting."
+                    )
+                    print(v.errors)
+                    sys.exit(1)
+        except FileNotFoundError as err:
+            print("Warning:")
+            print(args.ground_conf+" key file not found")
+            flask_keys_config=None
```
kkrol27: I know for most test cases this sorta logic is probably alright, but I'd assume this will cause some other cases to crash without explanation. Not sure if it's the cleanest way to do it...

install.sh
```diff
@@ -0,0 +1,18 @@
+#!/bin/bash
+
+python3.7 -m venv venv
```
kkrol27: This is also only going to run on arch?

ptest/run_simulation.py
tanishqaggarwal: Why will this cause a crash?

install.sh
```diff
@@ -0,0 +1,18 @@
...
+source venv/bin/activate
+pip install -r requirements.txt
+matlab -batch '"'"pyversion $(pwd)/venv/bin/python"'"' -nodisplay -nojvm -nosplash
+export PYTHONPATH=$(pwd)/venv
+if [ -d "/home/pan/MATLAB/R2019b/extern/engines/python" ]
```
kkrol27: Perhaps replace `/home/pan` with `$HOME` - it may be slightly more applicable at that point.

ptest/configs/fc_only_native.json
```diff
@@ -1,12 +1,12 @@
...
     "seed" : 2,
-    "sim_duration" : 10,
     "single_sat_sim" : true,
+    "sim_duration" : 1000,
     "devices" : [
         {
             "name": "FlightController",
             "run_mode" : "native",
-            "binary_filepath" : "/Users/tanishqaggarwal/Documents/pan/repositories/FlightSoftware/.pio/build/fsw_native_leader/program"
+            "binary_filepath" : ".pio/build/fsw_native_leader/program"
```
tanishqaggarwal: Wow, the relative filepath ends up working? Nice.

ptest/run_simulation.py
kkrol27: If you need that radio keys for a quake test or something like that? If we don't require the radio keys for anything right now then just leave it.

ptest/configs/fc_only_native.json
nhz2: yeah I think it works now because everything is in `FlightSoftware`

ptest/run_simulation.py
tanishqaggarwal: I think it's better to always check that we have all config files. Why let a user have a partially configured ptest on their system when they could easily have a fully-configured one?

install.sh
tanishqaggarwal: I think it would be better to allow passing a bunch of directories to the install script since the paths vary wildly across systems (since people can have MATLAB, FlightSoftware, etc. located in random places)

ptest/configs/fc_only_native.json
kkrol27: You'd have to be sure to run in from the root flight software directory - otherwise it's not going to work. In Python, you could actually use `__file__` and `os.path.*` functions to figure it out at runtime potentially...

tanishqaggarwal:
Closing as "wontfix".

PR #256, opened 2020-03-08
Constants tracker

tanishqaggarwal:
This will make recurring constants review a lot easier since the diff of every PR to the constants will be very explicit.

---
This change is [Reviewable](https://reviewable.io/reviews/pathfinder-for-autonomous-navigation/flightsoftware/256)


src/common/constant_tracker.hpp
```diff
@@ -0,0 +1,6 @@
+#ifndef CONSTANT_TRACKER_HPP_
+#define CONSTANT_TRACKER_HPP_
+
+#define TRACKED_CONSTANT(type, name, val) type name = val;
```
kkrol27: To keep things less verbose we could potentially do:

    #define TRACKED_CONSTANT(type, name, val) static constexpr type name = val

I don't think anything would ever not be `static constexpr`.

tools/constant_reporter.py
```diff
@@ -0,0 +1,25 @@
...
+                        "val" : const[2].strip(),
+                        "file" : path
+                    })
+
+log = "Keep this file in version control so that changes to constants are well-known.\n\n"
+for const in consts:
+    log += f"{const['file']}: \"{const['name']}\" = \"{const['val']}\"\n"
+
+with open("constants", 'w') as f:
+    f.write(log)
```
kkrol27: Also, not at all necessary but I think we could do this in a 5 line bash script lol

kkrol27:
One additional thought I had, we should design this with variables other than scalars in mind. For example, we'd want to be able to define gnc related constants like the prop system thruster matrix and such. Most likely we keep that as a `static constexpr lin::Matrix<float, 4, 3>` or something.

tools/constant_reporter.py
tanishqaggarwal: This allows flexibility if we want to create CSVs or other formats for the constants.

src/common/constant_tracker.hpp
tanishqaggarwal: See `constants.hpp`; some constants can't be static. I could totally imagine wanting to create a non-constexpr constant as well (such as a vector of constants) so I feel like this is the minimum macro we can use.

I could add some more macros like TRACKED_CONSTANT_S, TRACKED_CONSTANT_SC, etc. to fix this

PR #257, opened 2020-03-08
Dev/event storage

afoarce:
This update makes events store the control cycle count as a pointer to the state field instead of having it passed in to the constructor, allowing for the value to be easily written back during deserialize. The event storage tests are also updated to allow for reading the data of the previously signaled event by decrementing the event storages's pointer (which is auto incremented in signal).

---
This change is [Reviewable](https://reviewable.io/reviews/pathfinder-for-autonomous-navigation/flightsoftware/257)


test/test_fsw_event/test_fsw_event.cpp
```diff
@@ -1,83 +1,151 @@
...
     TEST_ASSERT_EQUAL(32 + 2, tf.event.bitsize());
 
     // Set values for test
-    tf.control_cycle_count = 20;
+    tf.control_cycle_count_ptr->set(ccno);
     tf.data1_f.set(true);
+
+    //std::cout << "first tf.data1_f " << tf.data1_f.get() << std::endl;
+    //std::cout << "first ba[32] " << ba[32] << std::endl;
+
```
tanishqaggarwal: Can you remove these comments if they're no longer useful

test/test_fsw_event/test_fsw_event.cpp
```diff
@@ -1,83 +1,151 @@
...
     TEST_ASSERT_EQUAL(false, ba[32]);
     TEST_ASSERT_EQUAL(true, ba[33]);
 
-    const char* print_result = tf.event.print();
-    TEST_ASSERT_EQUAL_STRING("E: time: 20, data: 0, 1", print_result);
+    // Test that the event is correctly printed when a print is requested.
+    const char *print_result = event.print();
+    const char *expected_fmt_string = "E: time: %d, data: 0, 1";
+    char expected_string[100];
+    memset(expected_string, 0, 100);
```
tanishqaggarwal: Turns out sprintf automatically adds in the null character to terminate the string, so there's no need for the memset

test/test_fsw_event/test_fsw_event.cpp
```diff
@@ -1,83 +1,151 @@
...
+public:
+    StateFieldRegistryMock registry;
+
     ReadableStateField<bool> data1_f;
     ReadableStateField<bool> data2_f;
-    std::vector<ReadableStateFieldBase*> event_data;
-    unsigned int control_cycle_count = 0;
+    std::vector<ReadableStateFieldBase *> event_data;
+    ReadableStateField<unsigned int> *control_cycle_count_ptr;
+    size_t *event_ptr;
```
tanishqaggarwal: Why not move event_ptr into TestFixtureEventStorage?

tanishqaggarwal: Let's also initialize the values of pointers to be nullptr by default

test/test_fsw_event/test_fsw_event.cpp
```diff
@@ -1,83 +1,151 @@
...
+    tf.registry.find_readable_field("event.3");
+    // Test that a field was added to the registry for every single sub-event.
+    TEST_ASSERT_EQUAL(99, tf.registry.readable_fields.size());
+
+    // Event storage should behave the same as an event.
+    for (int i = 0; i < 200; i++)
+    {
+        //printf("%d\n", i);
+        test_single_event(tf, tf.event_storage, i);
+        //TEST_ASSERT_EQUAL(i * 2, tf.event_storage.event_ptr);
```
tanishqaggarwal: You're not testing the event pointer being incremented here, or that a "new" event is going to be signaled

tanishqaggarwal:
It would be great if you could also push some description of Events and Event Storage to ReadTheDocs. https://github.com/pathfinder-for-autonomous-navigation/Documentation

test/test_fsw_event/test_fsw_event.cpp
afoarce: If event_ptr is moved, then test_event_storage() can't directly use test_single_event() since I need to decrement the pointer to verify that the data was written correctly. Would I write a different test case for single vs storage to account for the event_ptr not existing for single events?

src/fsw/FCCode/Event.cpp
```diff
@@ -85,12 +82,11 @@ void Event::deserialize()
 
     for (ReadableStateFieldBase *field : data_fields)
     {
-        bit_array field_bits;
+        bit_array field_bits = const_cast<bit_array &>(field->get_bit_array());
```
tanishqaggarwal: This creates a copy of the field's bit array. Copy vs move vs reference semantics are tricky! The right declaration here would be

`bit_array& field_bits = const_cast<bit_array &>(field->get_bit_array());`

(The ampersand declares field_bits as a reference instead of a copy of the returned bit array.)

src/fsw/FCCode/Event.cpp
```diff
@@ -24,10 +24,7 @@ Event::Event(const std::string &name,
 
 Event::Event(Event &&other) : StateField<bool>(other.name(), true, false),
                               data_fields(other.data_fields),
-                              field_data(std::move(other.field_data))
-{
-    ccno = other.ccno;
-}
+                              field_data(std::move(other.field_data)) {}
 
 void Event::serialize()
```
tanishqaggarwal: I don't think the serialize() and deserialize() functions are getting unit tested. The tests don't need to happen as part of this PR but I would suggest making them an issue so that we don't forget to write them.

src/fsw/FCCode/Event.hpp
```diff
@@ -3,48 +3,104 @@
 
 #include <common/StateField.hpp>
 
+/**
+ * @brief This event base class exposes methods for reading and
+ * writing to an event.
+ */
+class EventBase
```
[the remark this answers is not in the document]
tanishqaggarwal: Seems resolved. Thanks!

src/fsw/FCCode/Event.cpp
tanishqaggarwal: I thought of this because the bit_array copy vs reference issue would've definitely been found by a unit test before I found it.

afoarce: I think serialize is tested by signal, and I have now added a test for deserialize. I'm not sure how to test the coverage for these, though.

src/fsw/FCCode/EventStorage.cpp
```diff
@@ -0,0 +1,61 @@
...
+        sub_events.emplace_back(name + std::string(x),
+                                _data_fields, _print_fn);
+    }
+}
+
+void EventStorage::add_events_to_registry(StateFieldRegistry &registry)
+{
+    for (Event &e : sub_events)
+    {
+        registry.add_readable_field(static_cast<ReadableStateFieldBase *>(&e));
```
fatimayousuf: You can use `registry.add_event(e)` instead of casting it to a Readable Statefield

afoarce:
Moved to dev/event_storage2 branch

Issue #210, opened 2020-02-09
Piksi fault management.


tanishqaggarwal:
There are three kinds of fault handling required for Piksi. All of the following errors should revert the satellite to standby if the satellite is currently in a fault-responsive state:
- If the Piksi has not sent data in a long time (Piksi's state is `dead`)
- If we entered a close approach state but haven't gotten CDGPS in X time (X is a ground-configurable wait initially set to 3 hours)
- If we got CDGPS during close approach, but are still in close approach and haven't received CDGPS in the last Y hours (Y=3 initially but is ground-configurable)

PR #262, opened 2020-03-10
Fault handling software

### LINKED ISSUE #207: Handle mission manager faults with a state-fault table

tanishqaggarwal:
# Fault handling software

Fixes #207, but not with a fault table, but rather with a set of fault machines.

**See https://pan-software.readthedocs.io/en/latest/flight_software/fault_management.html for documentation.**

### Summary of changes
- Created an inheritance structure based off a class called `FaultHandlerMachine`, which is a `ControlTask<fault_response_t>`. `fault_response_t` is a struct containing a recommendation to either go to safehold, go to standby, or do nothing.
- Made stub implementations for QuakeFaultHandler and PiksiFaultHandler off of this machine, until the stubs can be filled out.
- Made implementations for SimpleFaultHandler and SuperSimpleFaultHandler, which are fault handler machines based off of a single fault flag.
- Made implementation for MainFaultHandler, which combines the result of several fault machines into one for reporting to the MissionManager. MainFaultHandler can be enabled or disabled by HITL/HOOTL so that ptest testcases that don't cause faults can be written in a simplified manner. Question: is it worth conditionally compiling this option out of flight builds?
- Integrated MainFaultHandler into MissionManager, but prevented it from being enabled in main flight software until I've written better end-to-end unit testing.

### Remaining items
- Completing and unit-testing the stub implementations of QuakeFaultHandler (this will require some rework in #209) and PiksiFaultHandler
- Completing the stub implementation of PropOverpressureFaultHandler, and unit-testing it. @athena255 We should coordinate on how to work this fault handler into your stuff
- Unit-testing MainFaultHandler end-to-end, with real underlying FaultHandlerMachines instead of mocks.
- Unit-testing the behavior of MainFaultHandler inside MissionManager using a mocked MainFaultHandler class.
- Enabling MainFaultHandler in MainControlLoop by disabling the flag override.

This work is being tracked in #291 #210 #295 #296 #297

### Testing
- The unit testing should produce 100% coverage of simple fault handlers and main fault handlers.
- I took care to apply every possible input to the simple fault handler and main fault handler and verified it has the correct response.
- I also test that the main fault handler responds correctly to its enable/disable flag.

I haven't yet performed a HITL or HOOTL, choosing instead to disable fault handling from flight software until I can write up an integration test. Additional testing is being tracked on: #290, #292, #298, #299

### Constants
The only changes to the `constants` files are line number changes due to reordering inside the files in which they're declared.

### Documentation Evidence
Pushed to RTD:
https://pan-software.readthedocs.io/en/latest/flight_software/fault_management.html

---
This change is [Reviewable](https://reviewable.io/reviews/pathfinder-for-autonomous-navigation/flightsoftware/262)


src/fsw/FCCode/SimpleFaultHandler.cpp
```diff
@@ -0,0 +1,56 @@
...
+
+SimpleFaultHandler::SimpleFaultHandler(StateFieldRegistry& r, Fault& f,
+    const std::vector<mission_state_t>& _active_states,
+    mission_state_t rs) :
+        FaultHandlerMachine(r),
+        fault(f),
+        active_states(_active_states),
+        recommended_state(rs)
+{
+    assert(rs == mission_state_t::safehold || rs == mission_state_t::standby);
```
stewartaslan: Is it ok to have this assert here in the flight code? Under what conditions might it fail? And what would happen if it failed?

src/fsw/FCCode/SimpleFaultHandler.cpp
```diff
@@ -0,0 +1,56 @@
...
+const std::vector<std::vector<mission_state_t>> SimpleFaultHandler::active_state_lists {
+    // List 0 is used by the Gomspace's low-battery fault
+    {
+        mission_state_t::initialization_hold, mission_state_t::detumble,
+        mission_state_t::standby, mission_state_t::follower, mission_state_t::leader,
+        mission_state_t::follower_close_approach, mission_state_t::leader_close_approach
+    },
+    // List 1 is used by the ADCS and propulsion faults
+    {
+        mission_state_t::standby, mission_state_t::follower, mission_state_t::leader,
```
stewartaslan: I think we should allow overpressure/overtemperature faults to occur in every state since there is a chance of things exploding.

stewartaslan: Actually hold the phone on this. We might have to follow some DNH regulation about not actuating any valves until we have communication with the ground. Let me look into this more and also re-asses how much risk of explosion there actually is.

test/test_fsw_fault_handler/cartesian_product.hpp
```diff
@@ -0,0 +1,75 @@
+#ifndef CARTESIAN_PRODUCT_HPP_
```
stewartaslan: This is not used in the flight code right? Just for testing?

src/fsw/FCCode/SimpleFaultHandler.cpp
tanishqaggarwal: I think it should be fine. It would fail if the SimpleFaultHandler was constructed with a "recommended state" that was neither safehold nor standby. Since all construction of objects happens instantaneously during the initial setup of flight software, we would catch this assertion failure immediately during HOOTLs.

test/test_fsw_fault_handler/cartesian_product.hpp
tanishqaggarwal: Yep, just for testing.

tanishqaggarwal:
No, there isn't. But if the wheel is persistenly broken even after the satellite reboots, shouldn't we go into safehold anyway? Going into any other state could be disastrous for attitude or angular rate

src/fsw/FCCode/FaultHandlerMachine.hpp
```diff
@@ -0,0 +1,23 @@
...
+
+/**
+ * @brief Abstract class for fault handler state machines, which must
+ * return a recommended mission state from their execute function.
+ */
+class FaultHandlerMachine : public ControlTask<fault_response_t> {
+  public:
+    FaultHandlerMachine(StateFieldRegistry& r) : ControlTask<fault_response_t>(r) {}
+    virtual fault_response_t execute() = 0;
+};
```
kkrol27: Do we want to explicitly delete the default constructor, copy constructor, and/or move constructor?

src/fsw/FCCode/MainFaultHandler.cpp
```diff
@@ -0,0 +1,79 @@
...
+    for(std::unique_ptr<FaultHandlerMachine>& m : fault_handler_machines) {
+        const fault_response_t response = m->execute();
+
+        if (response == fault_response_t::none) continue;
+        else if (response == fault_response_t::safehold) {
+            return fault_response_t::safehold;
+        }
+        else {
+            ret = fault_response_t::standby;
+        }
```
kkrol27: This logic works for the time being, but is there any chance that this gets more complicated in the future?

src/fsw/FCCode/MissionManager.cpp
```diff
@@ -155,6 +134,14 @@ void MissionManager::execute() {
     }
 }
 
+bool MissionManager::check_adcs_hardware_faults() const {
+    return  adcs_functional_fault_fp->is_faulted()
+            || wheel1_adc_fault_fp->is_faulted() 
+            || wheel2_adc_fault_fp->is_faulted()
+            || wheel3_adc_fault_fp->is_faulted()
+            || wheel_pot_fault_fp->is_faulted();
+}
```
kkrol27: Are we not worried about sensors here?

src/fsw/FCCode/SimpleFaultHandler.hpp
```diff
@@ -0,0 +1,67 @@
...
+    /**
+     * @brief Determines the recommended state to the main fault handler
+     * on the basis of the current mission state and the fault flag.
+     * 
+     * @return fault_response_t Recommended state (either nothing, standby, or safehold.)
+     */
+    fault_response_t determine_recommended_state() const;
+
+    static const WritableStateField<unsigned char>* mission_state_fp;
+    Fault& fault;
```
kkrol27: Do we potentially want to make this a `Fault *`? Just because all the constructors take in a dereferenced pointer.

src/fsw/FCCode/SimpleFaultHandler.cpp
kkrol27: I think this is a perfectly fine use of an assert. Even if something like this was in a control tasks execute function, I'd say it's still fine.

kkrol27:
> No, there isn't. But if the wheel is persistenly broken even after the satellite reboots, shouldn't we go into safehold anyway? Going into any other state could be disastrous for attitude or angular rate

I think in the current state of GNC we have no way to control for this. The only other option we would have is to keep the satellite in detumble. In that case, we should gets comms sooner or later.

src/fsw/FCCode/SimpleFaultHandler.hpp
tanishqaggarwal: Sure I could do that to make things more consistent.

src/fsw/FCCode/MissionManager.cpp
tanishqaggarwal: These are faults that cause a transition into safehold according to our previous sensors. If there's a reason to also consider sensors let's schedule a meeting to work it into our fault response?

src/fsw/FCCode/MainFaultHandler.cpp
tanishqaggarwal: Not sure. It'd only get more complex if faults recommend states other than safehold or standby--which i don't envision us doing.

src/fsw/FCCode/FaultHandlerMachine.hpp
tanishqaggarwal: sure I'll do that for safety

src/fsw/FCCode/SimpleFaultHandler.cpp
stewartaslan: This seems good. Adding in more states like detumble would be different than what we said in our FCC application so let's just leave it as-is.

src/fsw/FCCode/MainFaultHandler.cpp
kkrol27: Yeah true - should be alright then

src/fsw/FCCode/SimpleFaultHandler.hpp
kkrol27: Doesn't have to happen, but just noticed all of the `*(r.find_fault(...))`.

src/fsw/FCCode/fault_response_t.enum
```diff
@@ -0,0 +1,10 @@
+#ifndef FAULT_RESPONSE_T_ENUM_
+#define FAULT_RESPONSE_T_ENUM_
+
+enum class fault_response_t {
+  none,
+  safehold,
+  standby
```
shihaocao: are there ever plans to expand this?

tanishqaggarwal: Don't see a reason as to why we would expand.

src/fsw/FCCode/FaultHandlerMachine.hpp
```diff
@@ -0,0 +1,25 @@
...
+
+#include "TimedControlTask.hpp"
+#include "mission_state_t.enum"
+#include "fault_response_t.enum"
+#include <common/Fault.hpp>
+
+/**
+ * @brief Response structure for fault handler state machines.
+ */
+
```
athena255: where is the response structure?

src/fsw/FCCode/FieldCreatorTask.hpp
```diff
@@ -19,13 +19,15 @@ class FieldCreatorTask : public ControlTask<void> {
...
       Fault failed_pressurize_f;
+      Fault overpressured_f;
 
       FieldCreatorTask(StateFieldRegistry& r) : 
         ControlTask<void>(r),
         pos_f("orbit.pos", Serializer<d_vector_t>(0,100000,100)),
         pos_baseline_f("orbit.baseline_pos", Serializer<d_vector_t>(0,100000,100)),
         prop_state_f("prop.state", Serializer<unsigned char>(1)),
-        failed_pressurize_f("prop.failed_pressurize", 1, TimedControlTaskBase::control_cycle_count)
+        failed_pressurize_f("prop.failed_pressurize", 1, TimedControlTaskBase::control_cycle_count),
```
athena255: There seems to be many ways to access control_cycle_count. Perhaps we should consolidate it into one g_control_cycle_count?

src/fsw/FCCode/MainFaultHandler.cpp
```diff
@@ -0,0 +1,79 @@
...
+    }
+
+    for(Fault* fault : active_list_1_standby_super_simple_faults) {
+        fault_handler_machines.push_back(
+            std::make_unique<SuperSimpleFaultHandler>(_registry, fault,
+                SimpleFaultHandler::active_state_lists[1], mission_state_t::standby)
+        );
+    }
+
+    fault_handler_machines.push_back(std::make_unique<PropOverpressureFaultHandler>(_registry));
```
athena255: Why these unique pointers?

src/fsw/FCCode/MainFaultHandler.cpp
```diff
@@ -0,0 +1,79 @@
...
+    fault_handler_machines.push_back(std::make_unique<QuakeFaultHandler>(_registry));
+    fault_handler_machines.push_back(std::make_unique<PiksiFaultHandler>(_registry));
+}
+
+fault_response_t MainFaultHandler::execute() {
+    // By default, or if the fault handling is globally disabled, recommend no action.
+    fault_response_t ret = fault_response_t::none;
+    if (!fault_handler_enabled_f.get()) return ret;
+
+    for(std::unique_ptr<FaultHandlerMachine>& m : fault_handler_machines) {
```
athena255: Why are you declaring this reference const?

src/fsw/FCCode/SimpleFaultHandler.cpp
```diff
@@ -0,0 +1,56 @@
...
+            return fault_response_t::safehold;
+        }
+    }
+    else return fault_response_t::none;
+}
+
+const std::vector<std::vector<mission_state_t>> SimpleFaultHandler::active_state_lists {
+    // List 0 is used by the Gomspace's low-battery fault
+    {
+        mission_state_t::initialization_hold, mission_state_t::detumble,
```
athena255: Style-wise, I'd suggest making this one item per line. It's hard to read otherwise.

src/fsw/FCCode/SimpleFaultHandler.hpp
```diff
@@ -0,0 +1,67 @@
...
+    static void set_mission_state_ptr(WritableStateField<unsigned char>* ptr);
+
+  protected:
+    /**
+     * @brief Determines the recommended state to the main fault handler
+     * on the basis of the current mission state and the fault flag.
+     * 
+     * @return fault_response_t Recommended state (either nothing, standby, or safehold.)
+     */
+    fault_response_t determine_recommended_state() const;
```
athena255: Why is this only the "recommended" state as oppose to simply determining the state?

test/test_fsw_fault_handler/cartesian_product.hpp
```diff
@@ -0,0 +1,75 @@
...
+// Integer exponentiation.
+template<size_t A, int X>
+constexpr size_t intpow() {
+    static_assert(X >= 0, "This function computes an integer-valued "
+                         "exponentiation and requires a positive exponent"
+                         "argument.");
+    static_assert(A != 0 || X > 0, "0^0 is undefined.");
+
+    size_t result = 1;
+    for(int i = X; i > 0; i--) result *= A;
```
athena255: Why are we manually doing integer exponentiation?

test/test_fsw_fault_handler/cartesian_product.hpp
```diff
@@ -0,0 +1,75 @@
...
+
+// Integer exponentiation.
+template<size_t A, int X>
+constexpr size_t intpow() {
+    static_assert(X >= 0, "This function computes an integer-valued "
+                         "exponentiation and requires a positive exponent"
+                         "argument.");
+    static_assert(A != 0 || X > 0, "0^0 is undefined.");
+
+    size_t result = 1;
```
athena255: Perhaps make this a uint64_t since size_t is compiled to long long on x86_64 but not on ARM?

src/fsw/FCCode/MainFaultHandler.cpp
tanishqaggarwal: It's not being declared const

src/fsw/FCCode/SimpleFaultHandler.cpp
tanishqaggarwal: Sure

src/fsw/FCCode/SimpleFaultHandler.hpp
tanishqaggarwal: Each sub- fault handler makes a recommendation to the MainFaultHandler. The MainFaultHandler chooses the most "severe" recommendation and forwards it to the MissionManager. The mission manager then acts upon MainFaultHandler's recommendation.

test/test_fsw_fault_handler/cartesian_product.hpp
tanishqaggarwal: I couldn't find an STL function that does constexpr integer exponentation. Do you know of one?

test/test_fsw_fault_handler/cartesian_product.hpp
tanishqaggarwal: This is only used for unit testing but I agree, let's reduce instances of size_t wherever possible

src/fsw/FCCode/MainFaultHandler.cpp
tanishqaggarwal: The fault handlers uniquely belong to the MainFaultHandler, so I decided to make this relationship compiler-enforced via `unique_ptr`.

src/fsw/FCCode/FaultHandlerMachine.hpp
tanishqaggarwal: Sorry, this is an outdated comment. Thanks for catching.

src/common/Fault.hpp
```diff
@@ -82,7 +82,7 @@ class Fault : public WritableStateField<bool> {
     using WritableStateField<bool>::set;
     using WritableStateField<bool>::get;
 
-    unsigned int& cc; // Control cycle count
+    const unsigned int& cc; // Control cycle count
```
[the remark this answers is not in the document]
tanishqaggarwal: You're right, I'll make a new PR to get rid of this since faults are pretty global at this point.

src/common/Fault.cpp
```diff
@@ -1,7 +1,7 @@
 #include "Fault.hpp"
 
 Fault::Fault(const std::string& name,
-      const size_t _persistence, unsigned int& control_cycle_count) : 
+      const size_t _persistence, const unsigned int& control_cycle_count) : 
     WritableStateField<bool>(name, Serializer<bool>()),
     _name(name),
     fault_bool_sr(),
```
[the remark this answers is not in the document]
tanishqaggarwal: 1. If it's the first control cycle then both cc and last_fault_time are zero. If the fault should be signaled on the first control cycle, this logic lets it happen.

2. No bc sometimes in our tests we skip control cycles. In normal operation, it still counts consecutive cycles since both numbers (cc and last_fault_time) go up by 1 on each consecutive signal.

3. unsignal_f is only used by the ground to try to temporarily clear a fault. So there's no need to set it from unsignal().

src/common/Fault.hpp
```diff
@@ -82,7 +82,7 @@ class Fault : public WritableStateField<bool> {
     using WritableStateField<bool>::set;
```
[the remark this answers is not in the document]
tanishqaggarwal: The is_faulted() function is what actually determines if the fault is being signaled or not. The other state fields are there for the purposes of ground control.

The state fields aren't being made private; instead, the inherited functions from WritableStateField are being made private. I suppose we could make all of the created state fields private as well, but we've been playing it fast and loose anyway with our OO design, and the value of rework vs the value gained is a little dubious

src/fsw/FCCode/FieldCreatorTask.hpp
tanishqaggarwal: I think it's much better to indicate where the control cycle count comes from by always referencing it as TimedControlTaskBase::control_cycle_count. This makes a little clearer exactly which file to look in in order to find the constant (TimedControlTaskBase.hpp). Conversely, naming it `g_control_cycle_count` means that you have to have IntelliSense working in order to find the constant's definition.

tanishqaggarwal: What other ways are we currently accessing control_cycle_count?

src/common/Fault.hpp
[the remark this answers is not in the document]
tanishqaggarwal: #310

test/test_fsw_fault_handler/cartesian_product.hpp
tanishqaggarwal: Nevermind, I wasn't able to fix this without breaking tests. Changing to uint64_t breaks tests on Teensy.

src/fsw/FCCode/FieldCreatorTask.hpp
shihaocao: Looking through files in VSCode, it looks like in ADCSBoxMonitor, GomspaceController, MissionManager, etc we just use `control_cycle_count` instead of including the `TimeControlTaskBase::` prefix. Maybe that's what @athena255 was talking about?

src/fsw/FCCode/SimpleFaultHandler.cpp
```diff
@@ -0,0 +1,63 @@
...
+            return fault_response_t::standby;
+        }
+        else {
+            return fault_response_t::safehold;
+        }
+    }
+    else return fault_response_t::none;
+}
+
+const std::vector<std::vector<mission_state_t>> SimpleFaultHandler::active_state_lists {
```
shihaocao: Is it necessary to enclose this as a vector of vectors? If we name each active state list such as `gomspace_active_state_lists`, I think that will be much more clear and less human error-prone than `active_state_lists[0]`

src/fsw/FCCode/MainFaultHandler.cpp
```diff
@@ -0,0 +1,79 @@
...
+        find_fault("adcs_monitor.wheel1_fault", __FILE__, __LINE__),
+        find_fault("adcs_monitor.wheel2_fault", __FILE__, __LINE__),
+        find_fault("adcs_monitor.wheel3_fault", __FILE__, __LINE__),
+        find_fault("adcs_monitor.wheel_pot_fault", __FILE__, __LINE__),
+        find_fault("prop.failed_pressurize", __FILE__, __LINE__),
+        find_fault("prop.overpressured", __FILE__, __LINE__)
+    };
+
+    std::vector<Fault*> active_list_1_standby_super_simple_faults {
+        find_fault("prop.failed_pressurize", __FILE__, __LINE__),
```
shihaocao: `prop.failed_pressurize` is in here twice, is this intended?

src/fsw/FCCode/MissionManager.cpp
```diff
@@ -113,6 +134,14 @@ void MissionManager::execute() {
     }
 }
 
+bool MissionManager::check_adcs_hardware_faults() const {
```
shihaocao: It seems like `check_adcs_hardware_faults()` is no longer used anymore. This should be removed within the .cpp and .hpp?

src/fsw/FCCode/FieldCreatorTask.hpp
tanishqaggarwal: Ah, this is because those classes inherit from `TimedControlTask<T>`, but `FaultHandlerMachine`s don't (they only inherit from `ControlTask<T>`.) That's why control_cycle_count is available without name qualification in those tasks, but not in fault handler machines.

src/fsw/FCCode/FaultHandlerMachine.hpp
```diff
@@ -0,0 +1,21 @@
...
+/**
+ * @brief Abstract class for fault handler state machines, which must
+ * return a recommended mission state from their execute function.
+ */
+class FaultHandlerMachine : public ControlTask<fault_response_t> {
+  public:
+    explicit FaultHandlerMachine(StateFieldRegistry& r) : ControlTask<fault_response_t>(r) {}
+    FaultHandlerMachine& operator=(const FaultHandlerMachine& other) = delete;
+    FaultHandlerMachine& operator=(const FaultHandlerMachine&& other) = delete;
```
kkrol27: You don't have to (and I don't think should) mark the argument of the move constructor and `const` when deleting it. Move constructor taking rvalue references don't have to take a `const` argument.

src/fsw/FCCode/MissionManager.cpp
```diff
@@ -44,6 +45,10 @@ MissionManager::MissionManager(StateFieldRegistry& registry, unsigned int offset
     add_readable_field(deployment_wait_elapsed_f);
     add_writable_field(sat_designation_f);
 
+    main_fault_handler = std::make_unique<MainFaultHandler>(registry);
+    static_cast<MainFaultHandler*>(main_fault_handler.get())->init();
```
kkrol27: Why exactly can the state field registry be casted to `MainFaultHandler*`? That seems a little odd.

src/fsw/FCCode/MissionManager.cpp
kkrol27: Hmmmm... okay yeah. I think @stewartaslan @nhz2 and I may need to talk about that. This is good for the time being though

test/test_fsw_fault_handler/cartesian_product.hpp
kkrol27: Doesn't really matter here, but just as a fun aside you could also do the exponentiation with structs:
    
    #include <type_traits>

    template <int A, int X, typename = void>
    struct power : std::integral_constant<int, A * power<A, X - 1>::value> { };

    template <int A>
    struct power<A, 0> : std::integral_constant<int, 1> { };

    // Something here to handle a negative exponent case
    template <int A, int X>
    struct power<A, X, std::enable_if_t<(X < 0)>> : std::integral_constant<int, 0> { };

src/fsw/FCCode/MissionManager.cpp
tanishqaggarwal: It's not a cast. Registry is the argument to the MainFaultHandler's constructor.

src/fsw/FCCode/FaultHandlerMachine.hpp
tanishqaggarwal: Fixed

src/fsw/FCCode/SimpleFaultHandler.cpp
tanishqaggarwal: The thing is that multiple fault handlers share the same list, so I'd have to name some things like `power_adcs_gomspace_active_state_lists`, which IMO is less clear. Do you have a suggestion for this?

src/fsw/FCCode/MainFaultHandler.cpp
tanishqaggarwal: Yikes, good catch. It should only be in the second

src/fsw/FCCode/MissionManager.cpp
tanishqaggarwal: It should be used. I added it to be used.

src/fsw/FCCode/SimpleFaultHandler.cpp
shihaocao: I agree that for some it'd be bulkier, but I feel that's still more clear than an arbitrary index number. It's not a big deal tho, just my opinion

shihaocao: We could also just duplicate `power_adcs_prop_active_state_lists` and then give them separate names?

tanishqaggarwal: Marking this as closed for now because I think we'll rarely revisit this.

src/fsw/FCCode/MissionManager.cpp
```diff
@@ -113,6 +134,14 @@ void MissionManager::execute() {
...
+bool MissionManager::check_adcs_hardware_faults() const {
+    return  adcs_functional_fault_fp->is_faulted()
+            || wheel1_adc_fault_fp->is_faulted() 
+            || wheel2_adc_fault_fp->is_faulted()
+            || wheel3_adc_fault_fp->is_faulted()
+            || wheel_pot_fault_fp->is_faulted();
+}
+
 void MissionManager::dispatch_startup() {
     set(radio_state_t::disabled);
```
shihaocao: Is this line necessary? You already have a guard on line 102

src/fsw/FCCode/MissionManager.cpp
shihaocao: Wait I think Kyle is talking about line 49?

src/fsw/FCCode/MainFaultHandler.cpp
shihaocao: I think Athena is talking about line 66 (right below this comment)

src/fsw/FCCode/PropOverpressureFaultHandler.cpp
```diff
@@ -0,0 +1,15 @@
...
+        r,
+        r.find_fault("prop.overpressured"),
+        SimpleFaultHandler::active_state_lists[1],
+        mission_state_t::standby
+    )
+{}
+
+fault_response_t PropOverpressureFaultHandler::execute() {
+    // TODO cause venting response
+    return determine_recommended_state();
```
shihaocao: I'm a little confused here, how does PropOverPressureFaultHandler command/execute a venting response if it can only tell the mission manager to go into safehold or standby?

src/fsw/FCCode/MainFaultHandler.cpp
tanishqaggarwal: Can you screenshot the line? I'm still confused

src/fsw/FCCode/MissionManager.cpp
tanishqaggarwal: Thanks for catching. I'll remove the guard on line 102

src/fsw/FCCode/PropOverpressureFaultHandler.cpp
tanishqaggarwal: When this fault handler is fully implemented, it'll grab the state fields that it needs and set them in the right way to cause a venting response.

src/fsw/FCCode/MissionManager.cpp
tanishqaggarwal: I don't think so, the use of registry is on line 48.

src/fsw/FCCode/PropOverpressureFaultHandler.cpp
shihaocao: Ah. Makes sense.

src/fsw/FCCode/MissionManager.cpp
```diff
@@ -90,9 +95,20 @@ MissionManager::MissionManager(StateFieldRegistry& registry, unsigned int offset
...
+
+    // Step 1. Change state if faults exist.
+    const fault_response_t fault_response = main_fault_handler->execute();
+    if (fault_response == fault_response_t::safehold) {
+        transition_to_state(mission_state_t::safehold, adcs_state_t::zero_torque, prop_state_t::disabled);
+        return;
+    }
+    else if (fault_response == fault_response_t::standby) {
+        transition_to_state(mission_state_t::standby, adcs_state_t::point_standby, prop_state_t::idle);
+        return;
```
shihaocao: `SimpleFaultHandler.cpp` active list 1 has standby as a responsive state. As it currently stands, active_list_1 standby, has one fault: `prop.failed_pressurize`. Calling `return` on line 108 means we will never reach `dispatch_standby()`

I guess this is okay, but I think that just means we need to make sure that the `prop.failured_pressurize` fault is cleared eventually. However, I could not find an `unsignal()` within PropController. Does this mean this fault needs to be manually cleared from the ground?

tanishqaggarwal: This is a great point. I should add checks that ignore the if-blocks if the mission state is already safehold (on line 102) or standby (on line 106).

A fault like that should definitely be cleared from the ground. Prop fault handling design is still in the works so your questions don't have a well-defined answer yet, but I think adding the mission state checks should resolve the issue. Thanks for catching.

PR #272, opened 2020-03-11
HAVT CMD Clear

shihaocao:
This PR includes a small bug fix that autonomously clears a HAVT cmd once it has been processed.

---
This change is [Reviewable](https://reviewable.io/reviews/pathfinder-for-autonomous-navigation/flightsoftware/272)


tanishqaggarwal:
Is this HITLed?

shihaocao:
Not yet, I'll mark as ready for review when that is.

src/fsw/FCCode/ADCSBoxController.cpp
```diff
@@ -103,6 +103,9 @@ void ADCSBoxController::execute(){
         if(reset_get) {
             temp_cmd_table.set(idx, reset_get);
             send_cmd_table = true;
+
+            // clear the state field now that it's loaded into temp_cmd_table
+            const_cast<WritableStateField<bool>*>(havt_cmd_reset_vector_fp[idx])->set(false);
```
kkrol27: Why do we need to use `const_cast` here? I mean immediately in this context I get it but is this convention pervasive across FSW right now?

shihaocao: Sorry for taking so long to respond, it's not pervasive across FSW. I just thought the const cast was more appropriate here because in all contexts controller should merely actuate a command and not modify the command. But it's an exception here that it needs to clear the command. Because it's an exception, I think `const_cast` here is appropriate.

shihaocao:
@tanishqaggarwal @kkrol27 Manually verified to pass HITL

PR #304, opened 2020-03-16
Added Macros for Finding Fields, Events, and Faults

### LINKED ISSUE #136: Macro-ify find_readable_field and find_writable_field
_Originally posted by @tanishqaggarwal in https://github.com/pathfinder-for-autonomous-navigation/FlightSoftware/pull/126/review_comment/create_

kkrol27:
Fixes #136.

### Summary of changes
- Adds macros to wrap calls to `find_internal_field`, `find_readable_field`, `find_writable_field`, `find_event`, and `find_fault`.
- Simply defaulted the virtual destructor for `ControlTask` as opposed to an empty implementation.

Going forward, we can transition to the use of the macros as opposed to the explicit function calls.

An example use would be:

    WritableStateField<unsigned char> const *adcs_state_fp = FIND_WRITABLE_FIELD(unsigned char, adcs.state);

### Testing
NA.

### Constants
NA.

### Documentation Evidence
Documented inline where the macros are defined.


src/fsw/FCCode/ControlTask.hpp
```diff
@@ -169,7 +169,13 @@ class ControlTask : protected debug_console {
     }
 };
 
-template<typename T>
-ControlTask<T>::~ControlTask() {}
+/* Convenient macros to find fields, events, and faults. The field/event/fault
+ * argument will be automatically stringified so no need to include quotation
+ * marks. */
+#define FIND_INTERNAL_FIELD(type, field) find_internal_field<type>(#field, __FILE__, __LINE__)
```
tanishqaggarwal: Making the stringification part of the macro removes the syntax highlighting around the string. I'm not sure that's better

kkrol27: That makes sense. Although personally the reason I used stringify was to make it seem like you were retrieving a variable itself (sorta like hiding implementation details) - using the quotes seems a tad verbose.

PR #308, opened 2020-03-18
File-based EEPROM in HOOTL

### LINKED ISSUE #306: Add file-based EEPROM to HOOTLs
So that the EEPROM logic may be tested.

tanishqaggarwal:
Fixes #306 

### Summary of changes
- Adds a JSON file-based EEPROM to HOOTL builds. The file, called `eeprom.json`, gets created in the root-level flight software if that is where the FlightSoftware binary is executed from.
- Refactored unit tests slightly so that they work both on desktop and Teensy.
- Added a "blink" program to common software targets so that we can quickly troubleshoot Teensy USB connectivity issues without loading up the full flight software.

### Testing
Unit tests pass, both on the desktop and Teensy platforms. This is pretty solid evidence for the feature working.

HOOTLs are unaffected by the change (I tested DeploymentToStandby).

### Constants
No constants were added.

### Documentation Evidence
TODO


src/common/targets/blink.cpp
```diff
@@ -0,0 +1,17 @@
+// REMOVE BEFORE FLIGHT "RESET" CODE
+// Resets satellite to original configuration prior to flight.
```
stewartaslan: What does this comment mean? Do we need to delete this file before flight even if we're not compiling this environment? I feel like the comment is a bit confusing and could lead to problems later.

tanishqaggarwal: Oops this comment is a remnant of copying from the preflight.cpp file. Thanks for catching

Issue #246, opened 2020-03-06
Add serializer testing for floats, doubles, vectors, quaternions to DownlinkParser


tanishqaggarwal:

kkrol27:
Just out of curiosity, are we talking `lin::Vector*` types or the `vector_t` things?

tanishqaggarwal:
vector_t

tanishqaggarwal:
I think this is related to #329 and we can probably close this out.

PR #313, opened 2020-03-22
Uplink producer changes

### LINKED ISSUE #144: Uplink producer upgrades
Currently uplink producer/consumer only has the ability to edit integer state fields. We need to expand this ability to all kinds of state fields.

fatimayousuf:
Uplink Producer Upgrades

Fixes #144

### Summary of changes
This pr allows us to set vector statefields, quaternion statefields, floats, doubles, chars, etc. with uplinks.

### Testing
I didn't add anything to the existing tests for Uplink Producer. It would be good to have a way to deserialize vector and quaternion statefields (Issue #246), which could allow us to more rigorously test that the writable fields specified in json file are actually set to the value specified in the json file.

I did change test1.json and test2.json slightly. I changed pan.sat_designation from 3 to 1 in test_1.json. This is because the maximum value of the serializer for pan.sat_designation is 2, so we could only set that statefield to a value from 0-2. I also changed the docking config commands from 0 and 1 to false and true, because the statefield represents a boolean value

### Constants
N/A


src/gsw/parsers/src/UplinkProducer.cpp
```diff
@@ -22,6 +22,94 @@ UplinkProducer::UplinkProducer(StateFieldRegistry& r):
     }
  }
 
+template<typename UnderlyingType>
+UnderlyingType find_value(nlohmann::json j, std::string name) {
+    for (auto& e : j.items()) {
+        if (e.key() == name) return e.value();
+    }
+    throw std::runtime_error("cannot find value of " + name + " in statefield registry");
```
tanishqaggarwal: This error message is incorrect, I think. find_value is used to retrieve the value from the JSON specification of the uplink, not from the state field registry.

src/gsw/parsers/src/UplinkProducer.cpp
```diff
@@ -22,6 +22,94 @@ UplinkProducer::UplinkProducer(StateFieldRegistry& r):
...
+}
+
+template<typename UnderlyingType>
+bool UplinkProducer::try_add_field(bitstream bs, std::string key, nlohmann::json j) {
+    // Get pointer to that field in the registry
+    WritableStateField<UnderlyingType>* ptr = dynamic_cast<WritableStateField<UnderlyingType>*>(registry.find_writable_field(key));
+
+    // If the statefield of the given underlying type doesn't exist in the registry, return false. Otherwise, get the value of the key
+    if (!ptr) return false;
+    uint64_t val = find_value<uint64_t>(j, key);
```
tanishqaggarwal: Do you always want to be retrieving a 64-bit integer? What if the value is a float, double, bool, unsigned char, unsigned int, etc. I thought you would want to retrieve `UnderlyingType` here.

src/gsw/parsers/src/UplinkProducer.cpp
```diff
@@ -22,6 +22,94 @@ UplinkProducer::UplinkProducer(StateFieldRegistry& r):
...
+template<typename UnderlyingType>
+UnderlyingType find_value(nlohmann::json j, std::string name) {
+    for (auto& e : j.items()) {
+        if (e.key() == name) return e.value();
+    }
+    throw std::runtime_error("cannot find value of " + name + " in statefield registry");
+}
+
+template<typename UnderlyingType>
+bool UplinkProducer::try_add_field(bitstream bs, std::string key, nlohmann::json j) {
```
tanishqaggarwal: You can use static_assert to force the compiler to ensure this function is only used for specific types. Like so:

`static_assert(std::is_same<TypeThatIWant, UnderylingType>::value || std::is_same<AnotherTypeThatIWant, UnderlyingType>::value, "Some descriptive error msg");

src/gsw/parsers/src/UplinkProducer.cpp
```diff
@@ -22,6 +22,94 @@ UplinkProducer::UplinkProducer(StateFieldRegistry& r):
...
+
+template<typename UnderlyingType>
+bool UplinkProducer::try_add_vector_field(bitstream bs, std::string key, nlohmann::json j) {
+    // Get pointer to that field in the registry
+    using UnderlyingVectorType = std::array<UnderlyingType, 3>;
+    WritableStateField<UnderlyingVectorType>* ptr = dynamic_cast<WritableStateField<UnderlyingVectorType>*>(registry.find_writable_field(key));
+
+    // If the statefield of the given underlying type doesn't exist in the registry, return false. Otherwise, get the values of the key
+    if (!ptr) return false;
+    std::array<uint64_t, 3> vals = find_value<std::array<uint64_t, 3>>(j, key);
```
tanishqaggarwal: Same comment here about always retrieving a uint64_t; you actually only want to be retrieving either a float or a double. The UnderlyingType should be `float` or `double`.

src/gsw/parsers/src/UplinkProducer.cpp
```diff
@@ -22,6 +22,94 @@ UplinkProducer::UplinkProducer(StateFieldRegistry& r):
...
+    // Get pointer to that field in the registry
+    WritableStateField<UnderlyingType>* ptr = dynamic_cast<WritableStateField<UnderlyingType>*>(registry.find_writable_field(key));
+
+    // If the statefield of the given underlying type doesn't exist in the registry, return false. Otherwise, get the values of the key
+    if (!ptr) return false;
+    auto vals = find_value<std::array<uint64_t, 3>>(j, key);
+    size_t field_index=field_map[key];
+
+    // Set the statefield pointer to the new values. Floats will be implicitly casted to doubles for d_quat_t
+    ptr->set({static_cast<float>(vals.at(0)), static_cast<float>(vals.at(1)),
```
tanishqaggarwal: You don't want to always static_cast to a float.

src/gsw/parsers/src/UplinkProducer.cpp
```diff
@@ -22,6 +22,94 @@ UplinkProducer::UplinkProducer(StateFieldRegistry& r):
...
+    WritableStateField<UnderlyingVectorType>* ptr = dynamic_cast<WritableStateField<UnderlyingVectorType>*>(registry.find_writable_field(key));
+
+    // If the statefield of the given underlying type doesn't exist in the registry, return false. Otherwise, get the values of the key
+    if (!ptr) return false;
+    std::array<uint64_t, 3> vals = find_value<std::array<uint64_t, 3>>(j, key);
+    size_t field_index=field_map[key];
+
+    ptr->set({static_cast<UnderlyingType>(vals.at(0)), static_cast<UnderlyingType>(vals.at(1)), static_cast<UnderlyingType>(vals.at(2))});
+    ptr->serialize();
+
```
tanishqaggarwal: You should check that the magnitude of the vector isn't below or above the serialization limits.

src/gsw/parsers/src/UplinkProducer.cpp
```diff
@@ -22,6 +22,94 @@ UplinkProducer::UplinkProducer(StateFieldRegistry& r):
...
+    // If the statefield of the given underlying type doesn't exist in the registry, return false. Otherwise, get the values of the key
+    if (!ptr) return false;
+    auto vals = find_value<std::array<uint64_t, 3>>(j, key);
+    size_t field_index=field_map[key];
+
+    // Set the statefield pointer to the new values. Floats will be implicitly casted to doubles for d_quat_t
+    ptr->set({static_cast<float>(vals.at(0)), static_cast<float>(vals.at(1)), 
+        static_cast<float>(vals.at(2)), static_cast<float>(vals.at(3))});
+    ptr->serialize();
+
```
tanishqaggarwal: You should check that the magnitude of the quaternion is in the range [0, 1].

src/gsw/parsers/src/UplinkProducer.cpp
```diff
@@ -22,6 +22,94 @@ UplinkProducer::UplinkProducer(StateFieldRegistry& r):
...
+    found_field_type |= try_add_field<signed int>(bs, key, j);
+    found_field_type |= try_add_field<unsigned char>(bs, key, j);
+    found_field_type |= try_add_field<signed char>(bs, key, j);
+    found_field_type |= try_add_field<float>(bs, key, j);
+    found_field_type |= try_add_field<double>(bs, key, j);
+    found_field_type |= try_add_field<bool>(bs, key, j);
+    found_field_type |= try_add_vector_field<float>(bs, key, j);
+    found_field_type |= try_add_vector_field<double>(bs, key, j);
+    found_field_type |= try_add_quat_field<f_quat_t>(bs, key, j);
+    found_field_type |= try_add_quat_field<d_quat_t>(bs, key, j);
```
tanishqaggarwal: Underlying type should probably be float or double here.

src/gsw/parsers/src/UplinkProducer.cpp
```diff
@@ -22,6 +22,94 @@ UplinkProducer::UplinkProducer(StateFieldRegistry& r):
...
+    ptr->serialize();
+
+    // Make sure the value we want to set does not exceed the max possible
+    // value that the field can be set to
+    size_t field_index=field_map[key];
+    uint64_t max_val = (1ul << (get_field_length(field_index) + 1)) - 1;
+    if (val > max_val) {
+        throw std::runtime_error("cannot assign " + std::to_string(val) + " to field " + key + ". max value: " + std::to_string(max_val));
+        return false;
+    }
```
tanishqaggarwal: You can get the minimum/maximum via `ptr->get_serializer_min()`, `ptr->get_serializer_max()`.

tanishqaggarwal: Should also check the minimum value.

src/gsw/parsers/src/UplinkProducer.cpp
```diff
@@ -22,6 +22,94 @@ UplinkProducer::UplinkProducer(StateFieldRegistry& r):
...
+    ptr->serialize();
+
+    // Add the updated value to the bitstream
+    add_entry(bs, ptr->get_bit_array(), field_index);
+    return true;
+}
+
+bool UplinkProducer::add_field_to_bitstream(bitstream bs, std::string key, nlohmann::json j) {
+    bool found_field_type = false;
+    found_field_type |= try_add_field<unsigned int>(bs, key, j);
```
tanishqaggarwal: We also need gps_time_t in this list.

src/gsw/parsers/src/UplinkProducer.cpp
fatimayousuf: I can control what types this function is used for in `add_field_to_bitstream()`

src/gsw/parsers/src/UplinkProducer.cpp
```diff
@@ -22,6 +22,127 @@ UplinkProducer::UplinkProducer(StateFieldRegistry& r):
...
+                  || (std::is_same<UnderlyingType, float>::value && std::is_same<UnderlyingQuatType, f_quat_t>::value),
+        "Can't collect quaternion field info for a vector of non-float or non-double type.");
+    WritableStateField<UnderlyingQuatType>* ptr = dynamic_cast<WritableStateField<UnderlyingQuatType>*>(registry.find_writable_field(key));
+
+    // If the quaternion statefield of the given underlying type doesn't exist in the registry, return false. Otherwise, get the values of the key
+    if (!ptr) return false;
+
+    // Check that the magnitude of the values in the JSON file is less than or equal to 1
+    std::array<UnderlyingType, 4> vals = j[key];
+    unsigned int quat_mag = sqrt(pow(vals[0], 2) + pow(vals[1], 2) + pow(vals[2], 2) + pow(vals[3], 2));
```
tanishqaggarwal: The magnitude should be a float/double

src/gsw/parsers/src/UplinkProducer.cpp
```diff
@@ -22,6 +22,127 @@ UplinkProducer::UplinkProducer(StateFieldRegistry& r):
...
+    return true;
+}
+
+template<typename UnderlyingType, typename UnderlyingQuatType>
+bool UplinkProducer::try_add_quat_field(bitstream bs, std::string key, nlohmann::json j) {
+    // Get pointer to that field in the registry
+    static_assert((std::is_same<UnderlyingType, double>::value && std::is_same<UnderlyingQuatType, d_quat_t>::value)
+                  || (std::is_same<UnderlyingType, float>::value && std::is_same<UnderlyingQuatType, f_quat_t>::value),
+        "Can't collect quaternion field info for a vector of non-float or non-double type.");
+    WritableStateField<UnderlyingQuatType>* ptr = dynamic_cast<WritableStateField<UnderlyingQuatType>*>(registry.find_writable_field(key));
```
tanishqaggarwal: Instead of having two template arguments, consider replacing this line with

`WritableStateField<std::array<UnderlyingType, 4>>* ptr = ...`

tanishqaggarwal: I like what you did with UnderlyingVectorType in the function above.

src/gsw/parsers/src/UplinkProducer.cpp
```diff
@@ -22,6 +22,127 @@ UplinkProducer::UplinkProducer(StateFieldRegistry& r):
...
+    if (vector_mag>max || vector_mag<min) throw std::runtime_error("Magnitude of vector must be in range [" + 
+        std::to_string(min) + "," + std::to_string(max) + "]");
+
+    ptr->set(vals);
+    ptr->serialize();
+
+    // Add the updated value to the bitstream
+    size_t field_index=field_map[key];
+    add_entry(bs, ptr->get_bit_array(), field_index);
+    return true;
```
tanishqaggarwal: Rather than returning true from these functions, it might be more correct to return if the result of add_entry was greater than zero. I think (but I'm not sure) that bitstream's editN function returns zero bytes if there were no more bytes available in the bitstream to write to.

tanishqaggarwal: Can you add some code that throws a runtime error if the uplink is too large? Or defer this work to an issue?

src/gsw/parsers/src/UplinkProducer.cpp
tanishqaggarwal: Once you change it to float/double remember to check that the number is both greater than zero and less than 1

fatimayousuf: But isn't the magnitude of a vector/quaternion by definition greater than 0 (i.e the square root of squared numbers)?

tanishqaggarwal: There could be floating-point errors in the sqrt computation that reduce the magnitude to be below zero.

Also, my mistake, the magnitude of a quaternion should always be 1. So what we should really check for is if |magnitude - 1| < e, where e is some small error value (maybe 1e-10).

Finally, consider using std::sqrt and std::pow since they're templated versions of the math functions, which automatically resolve float vs double computation

src/gsw/parsers/src/UplinkProducer.cpp
fatimayousuf: Don't lines 158-159 in UplinkProducer
`if (!bs.has_next())
                throw std::runtime_error("bitstream is not large enough");`
already check if an we are trying to write too many values in a single uplink?

tanishqaggarwal: Ah, good point. Sorry that I missed that

src/gsw/parsers/src/UplinkProducer.cpp
```diff
@@ -22,6 +22,125 @@ UplinkProducer::UplinkProducer(StateFieldRegistry& r):
...
+    ptr->serialize();
+
+    // Add the updated value to the bitstream
+    size_t field_index=field_map[key];
+    return add_entry(bs, ptr->get_bit_array(), field_index);
+}
+
+size_t UplinkProducer::add_field_to_bitstream(bitstream bs, std::string key, nlohmann::json j) {
+    size_t bits_written = 0;
+    bits_written |= try_add_field<unsigned int>(bs, key, j);
```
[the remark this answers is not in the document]
fatimayousuf: Oh yeah I forgot to change that

PR #316, opened 2020-03-23
Event Storage

afoarce:
#257 

### Summary of changes
- makes events store the control cycle count as a pointer to the state field instead of having it passed in to the constructor, allowing for the value to be easily written back during deserialize
- implemented deserialize and added unit tests for it
- added event storage tests   

### Testing
-Event class is tested with test_single_event, where pointers to an event's ccno and data fields are used to set the values and test that serializing correctly writes those values to the event's bitset. Changes to the ccno/data fields are also tested by updating them, signaling (serializing) and checking their bitset again. Print is checked so that the correct data values and format are outputted.  To test deserializing, the ccno and data fields are changed and then we check that deserializing writes them back to their originally serialized values. 
-EventStorage should act the same as a single event, so we call test_single_event for twice the storage size of the EventStorage object and check that the event pointer is updating to the next event each time. Pointers are kept to the current event pointer and to the previous so we can read an event's values after they are written and the pointer increments during signal. 

### Documentation Evidence
documentation issue ticket created:
https://github.com/pathfinder-for-autonomous-navigation/Documentation/issues/15


test/test_fsw_event/test_fsw_event.cpp
```diff
@@ -21,54 +25,147 @@ struct TestFixture {
...
+    tf.registry.find_readable_field("event.1");
+    tf.registry.find_readable_field("event.2");
+    tf.registry.find_readable_field("event.3");
+    // Test that a field was added to the registry for every single sub-event.
+    TEST_ASSERT_EQUAL(99, tf.registry.readable_fields.size());
+
+    // Event storage should behave the same as an event.
+    for (int i = 0; i < 200; i++)
+    {
+        TEST_ASSERT_EQUAL((2 * i) % 99, *tf.event_ptr); //event signaled twice per single event, storage size 99
```
tanishqaggarwal: Why is the event signaled twice?

src/common/Event.hpp
```diff
@@ -3,51 +3,122 @@
...
-    // The value of events should never be set via deserialize or set_bit_array(), so these
-    // functions will have stub implementations.
-    void deserialize() override;
-    bool deserialize(const char *val) override;
-    void set_bit_array(const bit_array& arr) override;
+   //  // The value of events should never be set via deserialize or set_bit_array(), so these
+   //  // functions will have stub implementations.
+   //  void deserialize() override;
+   //  bool deserialize(const char *val) override;
+   //  void set_bit_array(const bit_array& arr) override;
```
tanishqaggarwal: I think this comment should be deleted.

src/common/EventStorage.cpp
```diff
@@ -0,0 +1,67 @@
...
+
+void EventStorage::deserialize() 
+{
+    sub_events[event_ptr].deserialize();
+}
+
+void EventStorage::set_bit_array(const bit_array &arr)  
+{
+    sub_events[event_ptr].set_bit_array(arr);
+}
```
tanishqaggarwal: I think it would be better if these functions were unimplemented, since we don't expect to use them. Unless you're using them in the unit test?

src/common/EventStorage.cpp
```diff
@@ -0,0 +1,67 @@
...
+        sub_events.emplace_back(name + std::string(x),
+                                _data_fields, _print_fn);
+    }
+}
+
+void EventStorage::add_events_to_registry(StateFieldRegistry &registry)
+{
+    for (Event &e : sub_events)
+    {
+        registry.add_readable_field(static_cast<ReadableStateFieldBase *>(&e));
```
fatimayousuf: You can use registry.add_event() instead of casting it to a Readable StateField

test/test_fsw_event/test_fsw_event.cpp
```diff
@@ -21,54 +25,147 @@ struct TestFixture {
...
+};
+
+// Test that the event storage correctly manages pointers for event storage.
+void test_event_storage()
+{
+    TestFixtureEventStorage tf(99);
+    // Fields should have been created inside the state field registry.
+    tf.registry.find_readable_field("event.1");
+    tf.registry.find_readable_field("event.2");
+    tf.registry.find_readable_field("event.3");
```
fatimayousuf: what you did here is fine by me, but if you pull from master, you can use find_event() and registry.events.size()

test/test_fsw_event/test_fsw_event.cpp
fatimayousuf: It's signaled twice when `test_single_event()` is called

src/common/Event.hpp
```diff
@@ -3,51 +3,122 @@
 
 #include <common/StateField.hpp>
 
+/**
+ * @brief This event base class exposes methods for reading and
+ * writing to an event.
+ */
+class EventBase
```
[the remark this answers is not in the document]
afoarce: The EventBase class acts as an interface for Event and EventStorage, as they are meant to function in the same way for the user with EventStorage abstracting its storage of multiple Event objects.

src/common/EventStorage.cpp
afoarce: deserialize is unit tested, but I have now removed other unused functions from EventStorage

PR #317, opened 2020-03-23
Merge ADCS Sensor Characterization Scripts

kkrol27:
# Merge ADCS Sensor Characterization Scripts

### Summary of changes
- Added two ADCS targets to pull magnetometer and IMU data for sensor characterization.
- Added a Python script to record CSV data from the sensor characterization code.

### Testing
Tested during sensor characterization sessions.

### Constants
NA

### Documentation Evidence
See inline documentation.


.gitignore
```diff
@@ -9,6 +9,7 @@ geograv_wrapper.mexmaci64
 geomag_wrapper.mexmaci64
 
 test/output_export.cpp
+tmp
```
tanishqaggarwal: What creates this tmp folder? I don't see it in the Python script

kkrol27: Oh I guess we could add that in. I use tmp in general to store random stuff every now and then. It doesn't have to be where you write the output file to though.

PR #319, opened 2020-03-23
Updated to new lin library and made it a direct submodule

kkrol27:
# Updated `lin` library and made it a submodule of this repo

@shihaocao because I know you use `lin` objects a decent amount.

### Summary of changes
- See here for more information and a summary of changes within `lin` (https://github.com/pathfinder-for-autonomous-navigation/psim/pull/154)
- One small initializer list change within flight software because of how they are handled now (honestly, was surprised that after change 3500+ lines everything seemed to behave well)

### Testing
Ran tests locally.

### Constants
NA

### Documentation Evidence
Will be working on documentation for `lin` in general soon! In the short term, I'd recommend looking at the following files in roughly the specified order if you're interested:
 - `lin/core/tensor/stream.hpp`
 - `lin/core/tensor/base.hpp`
 - `lin/core/tensor/tensor.hpp`
 - `lin/core/matrix/*.hpp`
 - `lin/core/vector/*.hpp`
 - `lin/core/tensor/operations.hpp`
 - `lin/core/tensor/operators.hpp`
Those will give you a solid understanding of everything brought in with `#include <lin/core.hpp>`.

__EDIT:__ Just a tip for reading through code, if you see the `::Traits` member all over the place, just take it as magic and the code will make more sense for a quick overview haha. It really comes from `lin/core/traits/*` though.


test/test_fsw_mission_manager/test_fixture.cpp
```diff
@@ -160,7 +160,7 @@ void TestFixture::set_comms_blackout_period(int ccno) {
 
 // Set the angular rate of the spacecraft.
 void TestFixture::set_ang_rate(float rate) {
-    adcs_ang_momentum_fp->set({rate, 0, 0}); // TODO will need to change this once the inertia tensor
+    adcs_ang_momentum_fp->set({rate, 0.0f, 0.0f}); // TODO will need to change this once the inertia tensor
```
shihaocao: I have never paid too much attention to explicitly giving float/double args when working with FSW, was this change necessary to have things compile?

kkrol27: The actual issue here was the mixed type initializer list. `rate` was a `float` and `0` is an integer so the constructor couldn't be deduced (a little unsure as to why this changed but that was the problem).

With the new version of `lin` though it is possible to assign a double initializer list to a float vector and the cast will happen implicitly - e.g. the below code is now permissible:

   lin::Vector3f u = { 1.0, 2.0, 3.0 };

PR #324, opened 2020-03-26
Downlink parser event upgrades

### LINKED ISSUE #217: Downlink Parser Event Upgrades
The downlink parser needs to be aware of when a state field is an event and parse it appropriately for consumption. This also requires changes across how the ground software parses the JSONed downlink information.

fatimayousuf:
# Downlink Producer/Parser Event Upgrades

Fixes #217 .

### Summary of changes
- Allows us to send events through downlinks and parse them from the ground
- Fix config paths in ground software
- Small changes in the test files involving events, since control cycle count is no longer a constructor argument

### Testing
I added onto the downlink parser test file. I added an event to the test fixture statefield registry, and added the name of the event to the flow data. The test file ran the downlink producer and the downlink parser, and I checked that the downlink parser returned a json object formatted as expected. 

I also sent a json file to the pan email account and ran ground software (main.py) to see if there were any additional changes need to be made to ground software, and no further changes were needed. Note that this test didn't run the downlink parser, since I already tested that the downlink parser would return a json document in the correct format. this was just to make sure that a json document in this format could be added to the elasticsearch database.


src/fsw/FCCode/DownlinkProducer.cpp
```diff
@@ -146,10 +146,19 @@ void DownlinkProducer::execute() {
             downlink_frame_offset);
 
         for(auto& field : flow.field_list) {
-            field->serialize();
-            const bit_array& field_bits = field->get_bit_array();
-            add_bits_to_downlink_frame(field_bits, snapshot_ptr, packet_offset,
-                downlink_frame_offset);
+            if (field->name().find("event") != std::string::npos) {
```
tanishqaggarwal: You know, looking at the code that instantiates the flows I feel like a stronger and cleaner check for events would be

`if (_registry.find_event(field->name()))`

What do you think?

src/gsw/parsers/src/DownlinkParser.cpp
```diff
@@ -126,17 +126,52 @@ std::string DownlinkParser::process_downlink_packet(const std::vector<char>& pac
...
             for(ReadableStateFieldBase* field : flow->field_list) {
-                const std::vector<bool>::iterator field_end_it =
-                    frame_bits.begin() + field->get_bit_array().size();
-
-                const std::vector<bool> field_bits(frame_bits.begin(), field_end_it);
-                field->set_bit_array(field_bits);
-                field->deserialize();
-                ret["data"][field->name()] = std::string(field->print());
-                frame_bits.erase(frame_bits.begin(), field_end_it);
+                if (field->name().find("event") != std::string::npos) {
```
tanishqaggarwal: Same thing here, maybe, as I suggested above?

src/fsw/FCCode/DownlinkProducer.cpp
fatimayousuf: I'm thinking that makes more sense than what I did XD

PR #327, opened 2020-03-27
Cleaner Write State Syntax for Vectors and Bools in PTest

### LINKED ISSUE #255: Add "Native" Python Type Support to Write State
- ~Have Ptest automatically call cycle as an option~
- Have Ptest take *args
- Support for multiple arg types

__EDIT:__ See @kkrol27's comment below.

kkrol27:
# Cleaner Write State Syntax for Vector and Bools in PTest

Fixes #255.

### Summary of changes

Slightly reworked the write state interface to allow for:

* Feeding a variable number of arguments to the write state function to specify a vector:

        sim.write_state('some.vector.state.field', 0.0, 0.0, 0.0)

* Passing native booleans to the write state function as opposed to strings:

        sim.write_state('some.boolean.state.field', True)

* The `timeout` argument is now passed as a keyword argument and still default to `None`.

### Testing

@shihaocao will be using these features in the new ADCS PTest cases that are currently in development. We'll defer testing till then.

Otherwise, the changes seem to be completely backwards compatible and passed CI.

### Constants

NA

### Documentation Evidence

See inline documentation.

### DIFF
--- a/ptest/state_session.py
+++ b/ptest/state_session.py
@@ -205,7 +205,7 @@ class StateSession(object):
 
         return returned_vals == vals
 
-    def write_multiple_states(self, fields, vals, timeout = None):
+    def write_multiple_states(self, fields, vals, timeout=None):
         '''
         Write multiple states and check the write operation with feedback.
 
@@ -222,7 +222,26 @@ class StateSession(object):
 
         return self._write_state_basic(list(fields), list(vals), timeout)
 
-    def write_state(self, field, val, timeout=None):
+    def _val_to_str(self, val):
+        '''
+        Convert a state value or list of values into a single string writable to
+        a state.
+
+        Currently, the supported types are integers, doubles, integer vectors,
+        double vectors, and booleans.
+        '''
+        if type(val) is not list:
+            if type(val) is bool:
+                return 'true' if val else 'false'
+            else:
+                return str(val)
+        else:
+            string = ''
+            for _val in val:
+                val_str += self._val_to_str(_val) + ', '
+            return val_str[:len(val_str) - 2]
+
+    def write_state(self, field, *args, **kwargs):
         '''
         Write state and check write operation with feedback.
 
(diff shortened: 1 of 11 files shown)


kkrol27:
@tanishqaggarwal I'm slightly worried about the shift to keyword arguments having some backwards compatibility issues; do you know of any off the top of your head?

tanishqaggarwal:
If ptest runs on all of the existing cases then I don't think there should be any compatibility issues. The existing mission test cases use ptest functionality pretty comprehensively

kkrol27:
Okay, sounds good. I'll mark this ready for review then and I think @shihaocao is going to use it in the new ADCS test cases so if it's alright to defer feature testing to him I think that's the way to go.

ptest/state_session.py
```diff
@@ -222,26 +222,44 @@ def write_multiple_states(self, fields, vals, timeout = None):
...
         '''
         Override state and check write operation with feedback.
 
         Behaves the same way as write_state(), but is strictly written for a state variable that is overriden
         by the user, i.e. is no longer set by the simulation.
         '''
-
         self.overriden_variables.add(field)
-        return self._write_state_basic(field, val, timeout)
+        return self._write_state_basic(field, [self._val_to_str(args)], kwargs.get('timeout'))
```
kkrol27: @tanishqaggarwal Why are these are to `_write_state_basic` not wrapped up as lists?

ptest/state_session.py
```diff
@@ -222,36 +222,53 @@ def write_multiple_states(self, fields, vals, timeout = None):
...
-        try:
-            self.overriden_variables.remove(field)
-        except KeyError:
-            # It doesn't matter if you try to take the override off of a field that wasn't
-            # actually being overridden, so just ignore it.
-            return
+        If the state wasn't currently being overriden, then this functions just
+        acts as a no-op.
+        '''
+        self.overriden_variables.discard(field)
```
kkrol27: And I just replaced this with the `discard` function which doesn't do anything is the element wasn't in the set.

ptest/state_session.py
tanishqaggarwal: Not sure, they definitely should be. I've never used this function which is probably why it's not complaining.

PR #331, opened 2020-03-29
Eeprom booleans

### LINKED ISSUE #307: Add support for boolean state fields in EEPROM
We need to be able to store true/false values in EEPROM as well; this might be dependent on #305 so that EEPROMController can figure out the type of a variable upon initialization.

tanishqaggarwal:
# Add ability to save booleans to EEPROM

Fixes #331 #337.

### Summary of changes
- Made "EEPROM-savedness" a property of ReadableStateFields. The advantage of this approach is that it allows the declaration of EEPROM-saving to be much more localized. A field  is EEPROM-saved if its `eeprom_save_period` is greater than zero.
  - Created functions to allow conversion of a state field value to an EEPROM-saveable value and back, called `get_eeprom_repr` and `set_from_eeprom`.
- Made all previously EEPROM-saveable fields also EEPROM-saveable (`pan.deployed`,  `pan.deployment.elapsed`, `pan.state`, `pan.sat_designation`).
- Wherever possible, I replaced the use of `.at(x)` with `[x]` since it's cleaner.
- Modified the telemetry info generator to also report the EEPROM save period of state fields. This allows the EEPROM-saving to be integrated into our telemetry review.

### Testing
`DeploymentToStandby` case still passes, so HOOTLs are in good working order.

The get_eeprom_repr and set_eeprom_repr functions have been tested in `test_state_field.cpp`.

The attached unit test log demonstrates that get_eeprom_repr and set_from_eeprom work on Teensy.

[build.txt](https://github.com/pathfinder-for-autonomous-navigation/FlightSoftware/files/4399566/build.txt)

### Constants
I've added a new constant called `eeprom_size`, equal to 4096 (the number of bytes available in EEPROM on a Teensy 3.5/3.6). This value is used to bound the number of EEPROM-saved state fields via an assertion in EEPROMController.cpp.

### Documentation Evidence
Opened #338 to defer documentation.


src/common/StateFieldTypes.inl
```diff
@@ -48,10 +52,106 @@ class SerializableStateField : public StateField<T>, virtual public Serializable
...
+      {
+        // The EEPROM value is corrupted. Don't set from EEPROM.
+        return;
+      }
+
+      this->_val = static_cast<T>(val);
+    }
+
+    template<class Q = void>
+    typename std::enable_if<!is_eeprom_saveable(), Q>::type
```
kkrol27: Also, `std::enable_if_t<!is_eeprom_saveable(), Q>` should be sufficient and is less verbose.

src/common/Event.hpp
```diff
@@ -89,6 +89,13 @@ class Event : public ReadableStateFieldBase, public StateField<bool>, public Eve
       bool deserialize(const char *val) override;
       const char *print() const override;
 
+      // Functions from the SerializableStateField interface. These are
+      // going to receive stub implementations since events should not
+      // be EEPROM-saveable.
+      unsigned int eeprom_save_period() const override;
+      unsigned int get_eeprom_repr() const override;
+      void set_from_eeprom(unsigned int val) override;
```
kkrol27: Is it okay if these are not designated `virtual`? Sorry I'm not quite familiar enough with the whole state field hierarchy system.

src/common/StateFieldTypes.inl
```diff
@@ -48,10 +52,106 @@ class SerializableStateField : public StateField<T>, virtual public Serializable
...
+    unsigned int eeprom_save_period() const override { return __eeprom_save_period; }
+
+    /**
+     * @brief Gets a representation of the state field's value that can be
+     * stored in EEPROM.
+     * 
+     * @return unsigned int 
+     */
+    template<class Q = unsigned int>
+    typename std::enable_if<is_eeprom_saveable(), Q>::type
```
kkrol27: Actually, nevermind, I see what's going on here.

src/common/Event.hpp
tanishqaggarwal: These are overriding the virtual methods defined in ReadableStateFieldBase.

tanishqaggarwal:
I tried making a new state field class but the multiple inheritance was impossible to navigate, which is why I went down this route.

src/fsw/FCCode/EEPROMControllerCommon.cpp
```diff
@@ -1,33 +1,26 @@
...
 
-void EEPROMController::init(const std::vector<std::string>& statefields, const std::vector<unsigned int>& periods){
-  for (size_t i = 0; i<statefields.size(); i++){
-    // copy the string name of the statefield into a char array
-    char field[statefields.at(i).length() + 1];
-    strcpy(field, statefields.at(i).c_str());
-    // get the pointer to that statefield and add it to the pointer array
-    pointers.push_back(find_readable_field<unsigned int>(field, __FILE__, __LINE__));
+void EEPROMController::init() {
+  for (size_t i = 0; i<_registry.eeprom_saved_fields.size(); i++){
```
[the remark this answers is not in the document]
tanishqaggarwal: From teensy docs the EEPROM size is 4096 bytes so that should be the number of memory addresses. I’ll add some assertion code to protect from overpopulating the EEPROM

src/common/StateFieldTypes.inl
```diff
@@ -48,10 +52,106 @@ class SerializableStateField : public StateField<T>, virtual public Serializable
...
-          _serializer(s) {}
+          _serializer(s),
+          __eeprom_save_period(0) {}
+
+    static constexpr bool is_eeprom_saveable() {
+      return std::is_same<T, unsigned int>::value
+       || std::is_same<T, signed int>::value
+       || std::is_same<T, unsigned char>::value
+       || std::is_same<T, signed char>::value
+       || std::is_same<T, bool>::value;
```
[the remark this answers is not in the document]
tanishqaggarwal: Don’t think so, because we can’t fit vectors inside the space of an unsigned int.
The intent behind EEPROM-saving is to store state values that drive software behavior—I can’t think of any vectors that are either desirable to save or can’t be generated from saved ints/bools.

PR #335, opened 2020-03-29
ADCS Ptest Case + Ptest Upgrades

shihaocao:
# ADCS Ptest Case + Ptest Upgrades

Addresses #264 partially.

### Summary of changes
- Added `smart_read()` to state_session.py
- Added helpful features to base.py
- Added a basic HAVT + sensor checkout to ADCS Ptest case
- DCDCController goes right after MissionManager

Note: ADCS Ptest case is not done, SSA and Motors still need to be checked, will come in a future PR.

### Testing
HITL tested with the minimal ADCS setup (MAG1, MAG2, GYR on ADCSC + FC).
HITL tested with the EDU sat, diagnostics as expected. Also recorded (terminal output only) in the HITL logs folder.

### Constants
REORDERED CONTROL TASK EXECUTION. DCDCController is now right after MissionManager.

### Documentation Evidence
Post to ReadTheDocs, or
- Coming to ReadTheDocs ASAP


kkrol27:
Couple conflicts to clean up it looks like. Probably from the PR earlier today.

shihaocao:
lol ye fixing rn

ptest/cases/adcs_checkout_case.py
```diff
@@ -0,0 +1,195 @@
...
+        self.step()
+
+        self.print_rs("adcs_monitor.functional")
+        assert(self.rs("adcs_monitor.functional")), f"ADCSC Not Functional"
+
+        self.ws("pan.state", 11)
+        self.ws("adcs.state", 5)
+        self.ws("adcs_cmd.rwa_mode", 1)
+        self.ws("adcs_cmd.rwa_speed_cmd", [0,0,0])
+        self.ws("dcdc.ADCSMotor_cmd", True)
```
kkrol27: Just another "syntax" observation. Would it be worth making some sort of mode enum set for these state fields? Or are we just going to stick with the raw integer numbers the entire time?
If the latter, it'd be nice to have a comment specifying what is happening.

ptest/cases/adcs_checkout_case.py
```diff
@@ -0,0 +1,195 @@
...
+        edu_sat        = "110111011100" + "000000"
+
+        test_beds = {barebones_hitl:"BAREBONES HITL", minimal_hitl:"MINIMAL HITL", edu_sat:"EDU SAT"}
+
+        binary_string_havt_read = ''.join(["1" if x else "0" for x in self.havt_read])
+
+        if(binary_string_havt_read not in test_beds):
+            self.print_header("UN-RECOGNIZED HITL TEST BED. MAKE SURE HAVT_READ IS AS EXPECTED!")
+        else:
+            self.print_header(f"HAVT TABLE MATCHES WITH: {test_beds[binary_string_havt_read]}")
```
kkrol27: All of this is a very good idea. Although I'm wondering if we should be putting a lot of this code in a seprate location so it can be used in multiple PTest cases easily. Specifically, I'm talking about the

* Testbed definitions (image some ADCS object with functions `has_minimal_hitl`, `has_edu_sat`, `has_barebones_hitl`, etc.)
* Automatically resets and re-enables ADCS devices on startup and shutdown and detects what's functional (which supports the above feature)
* Disables all devices on startup
* et cetera et cetera

ptest/cases/adcs_checkout_case.py
```diff
@@ -0,0 +1,195 @@
...
+            print("Mag readings: ")
+            for i in range(10):
+                mag = mag_of(list_of_mag_rds[i])
+                print(f"{list_of_mag_rds[i]}, mag: {mag}")
+                self.soft_assert((25e-6 < mag and mag < 65e-6),
+                    "Mag reading out of expected (earth) bounds.")
+
+            # check readings changed over time
+            self.soft_assert(sum_of_differentials(list_of_mag_rds) > 0,
+                "Mag readings did not vary across readings.")
```
kkrol27: I like this

ptest/cases/base.py
```diff
@@ -136,6 +159,52 @@ def _setup_case(self):
...
+        Soft assert prints a fail message if the condition is False
+        
+        If specificied with a fail message, then a pass message, 
+        it will also print a pass message if condition is True.
+        '''
+        if condition: 
+            if len(args) == 1:
+                pass
+            else:
+                print(args[1])
```
kkrol27: What is exactly is up with this `if` `else`?

ptest/cases/adcs_checkout_case.py
kkrol27: Scratch this, we should just make an issue about supporting the `FSWEnum` type in `write_state`. I just saw the class definition below.

ptest/cases/adcs_checkout_case.py
```diff
@@ -0,0 +1,195 @@
...
+                list_of_mag_rds += [self.rs("adcs_monitor.mag_vec")]
+
+            # for each reading check, magnitude bounds
+            # earth's mag field is between 25 to 65 microteslas - Wikipedia
+            print("Mag readings: ")
+            for i in range(10):
+                mag = mag_of(list_of_mag_rds[i])
+                print(f"{list_of_mag_rds[i]}, mag: {mag}")
+                self.soft_assert((25e-6 < mag and mag < 65e-6),
+                    "Mag reading out of expected (earth) bounds.")
```
kkrol27: This may be broken all the time because of fields from the wheels, docking system, MTRs et cetera.

ptest/cases/adcs_checkout_case.py
tanishqaggarwal: I think the best way to address this would be to actually use the enums:

`self.ws("pan.state", self.mission_states.get_by_name("manual")`

etc.

ptest/cases/base.py
shihaocao: The `if else` is for allowing calling `soft_assert(1 == 1, "Fail msg", "Pass msg")` if you wanted a message to be printed when the test passed.

shihaocao: @tanishqaggarwal I wanted something to not halt the program if it failed. That way the user wouldn't have their tests "bottlenecked" by a faulty sensor early on in the program.

ptest/cases/adcs_checkout_case.py
shihaocao: Oops yeah, didn't update this block to use the enums, fixed

ptest/cases/adcs_checkout_case.py
shihaocao: True, I'll have different bounds once we're spinning wheels at the same time as doing readings etc.

ptest/cases/base.py
```diff
@@ -136,6 +159,52 @@ def _setup_case(self):
...
+        '''
+        print(f"{name} is {self.rs(name)}")
+    
+    def ws(self, name, val):
+        '''
+        Writes a state, and also confirms that the read command matches the applied state.
+        '''
+        self.sim.flight_controller.write_state(name, val)
+        read_val = self.rs(name)
+        assert(read_val == val), f"Write state not applied, expected: {val}, got {read_val} instead"
```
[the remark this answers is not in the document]
shihaocao: I'll move to SingleSatOnlyCase for now, will add support for multiple later.

ptest/cases/adcs_checkout_case.py
```diff
@@ -0,0 +1,195 @@
...
+class ADCSCheckoutCase(SingleSatOnlyCase):
+
+    @property
+    def havt_read(self):
+        read_list = [False for x in range(self.havt_length)]
+        for x in range(self.havt_length):
+            read_list[x] = self.rs("adcs_monitor.havt_device"+str(x))
+        return read_list
+
+    def step(self):
```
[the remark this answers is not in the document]
shihaocao: Just merged `step()` functionality under `cycle()`

ptest/cases/base.py
kkrol27: I actually like the idea of a soft assert. We have a look of conditions that are not well defined. For example, everything could actually be behaving nominally when some (and only some) of these soft asserts fail. We wouldn't want the testcase to halt, for example, if the magnetometer was reading slightly large values.

ptest/cases/adcs_checkout_case.py
```diff
@@ -0,0 +1,195 @@
...
+        list_of_list = [string_of_binary_list[4*i:(4*i)+4] for i in range((int)(self.havt_length/4)+1)]
+        final = [x + [" "] for x in list_of_list]
+
+        final_string = ''.join([''.join(x) for x in final])
+        print("HAVT Read: "+str(final_string))
+
+    def setup_case_singlesat(self):
+        pass
+
+    def run_case_singlesat(self):
```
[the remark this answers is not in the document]
shihaocao: Done

ptest/cases/adcs_checkout_case.py
shihaocao: I put this section in for easy tracking of which devices are expected to be functional for each setup, and to notify the user if there's an unexpected setup (ex. forgot to plug in something while running with the `minimal_setup`. I agree, that this should get moved to something higher up, though I'm thinking that i'll be a chunky PR on its own. I can toss up an issue for it and defer this to a ptest upgrade?

ptest/cases/base.py
tanishqaggarwal: Agree with soft assert so that we have multiple failure reports. It also allows for the case to run in HOOTL without failing.

ptest/cases/adcs_checkout_case.py
```diff
@@ -0,0 +1,202 @@
...
+            
+        # I forgot what it actually is, will update on next PR, after testing with EDU SAT
+        edu_sat        = "110111011100" + "000000"
+
+        test_beds = {barebones_hitl:"BAREBONES HITL", minimal_hitl:"MINIMAL HITL", edu_sat:"EDU SAT"}
+
+        binary_string_havt_read = ''.join(["1" if x else "0" for x in self.havt_read])
+
+        if(binary_string_havt_read not in test_beds):
+            self.print_header("UN-RECOGNIZED HITL TEST BED. MAKE SURE HAVT_READ IS AS EXPECTED!")
```
tanishqaggarwal: Shouldn't this cause a test case failure?

ptest/cases/adcs_checkout_case.py
```diff
@@ -0,0 +1,202 @@
...
+
+        # disable all devices
+        for x in range(self.havt_length):
+            self.ws(f"adcs_cmd.havt_disable{x}", True)
+
+        self.cycle()
+
+        self.logger.put("Post disabling all devices:")
+        self.print_havt_read()
+        self.soft_assert(([0 for x in range(self.havt_length)] == self.havt_read),
```
tanishqaggarwal: Slightly more succinct syntax might be

`self.soft_assert(all([x == 0 for x in self.havt_read]), "...")`

ptest/cases/adcs_checkout_case.py
```diff
@@ -0,0 +1,202 @@
...
+            self.cycle()
+            list_of_gyr_rds += [self.rs("adcs_monitor.gyr_vec")]
+
+        # for each reading check, magnitude bounds
+        # expected rotation???
+        self.logger.put("GYR readings: ")
+        for i in range(10):
+            mag = mag_of(list_of_gyr_rds[i])
+            self.logger.put(f"{list_of_gyr_rds[i]}, mag: {mag}")
+            # 3.8 ~= sqrt((125 * 0.017)^2 * 3)
```
tanishqaggarwal: I kind of understand what this comment means but a slightly longer one would be much clearer

ptest/cases/adcs_checkout_case.py
```diff
@@ -0,0 +1,202 @@
...
+        # If either mag1 or mag2 are up, run a check on it.
+        if self.rs("adcs_monitor.havt_device1") or self.rs("adcs_monitor.havt_device2"):
+            self.mag_checkout()
+
+        # Run checks on GYR if GYR is up.
+        if self.rs("adcs_monitor.havt_device0"):
+            self.gyr_checkout()
+
+
+        # TODO FURTHER CHECKOUTS
```
tanishqaggarwal: Can you add issues to track the TODOs?

ptest/cases/adcs_checkout_case.py
```diff
@@ -0,0 +1,202 @@
...
+                self.logger.put(f"Device #{x}, {self.havt_devices.get_by_num(x)} is not functional")
+
+        self.logger.put("Initial HAVT Table:")
+        self.print_havt_read()
+
+        # Note IMUGYR on left
+        # Just FC and ADCSC
+        barebones_hitl = "000111011100" + "000000"
+
+        # FC + ADCSC + MAG1 + MAG2 + GYR
```
tanishqaggarwal: If the flight controller is included in the HAVT table for the ADCS, and you disable all devices in the ADCS, will the communication between the FC and ADCS break (and therefore the testcase will stop responding?)

ptest/cases/adcs_checkout_case.py
```diff
@@ -0,0 +1,209 @@
...
+
+    Ex: sum_of_differentials([[1,1,1],[1,2,3],[1,2,2]]) evaluates to 4
+    '''
+    total_diff = [0 for x in lists_of_vals[0]]
+    for i in range(len(lists_of_vals) - 1):
+        diff = [abs(lists_of_vals[i][j] - lists_of_vals[i+1][j]) for j in range(len(lists_of_vals[i]))]
+        total_diff = [diff[x] + total_diff[x] for x in range(len(total_diff))]
+
+    return sum(total_diff)
+class ADCSCheckoutCase(SingleSatOnlyCase):
```
tanishqaggarwal: Add an extra line to separate the function from the class

tanishqaggarwal:
Can you add an issue ticket about the deferred documentation work?

ptest/cases/adcs_checkout_case.py
kkrol27: Wouldn't we want this to be something more like this:

    for testbed in testbeds:
        if (detected ^^ testbed) && testbed:
            testbed isn't supported

ptest/cases/adcs_checkout_case.py
kkrol27: There's no way to get the ADCS to stop listening to the flight computer.

shihaocao: Those are all ADCS HAVT tables, so the FC is not included in those strings.
So if you disable all devices, the connection between FC and ADCSC is unchanged.

ptest/cases/adcs_checkout_case.py
shihaocao: TIL about all()

ptest/cases/adcs_checkout_case.py
shihaocao: My logic is that if we suddenly have a new testbed, we might want to just run this script anyway to run checks on it. 

Or, if we know a bed has problems, this would just confirm those problems. Then we can still test whatever sensors/stuff is still functional.

shihaocao:
Sure. Won't merge until after testing.

ptest/cases/adcs_checkout_case.py
tanishqaggarwal: This shouldn't serve as a "development" ptest case, I think it should be a hard check on functionality. If a testbed suddenly has problems, I feel like the test case should fail hard. Everything else that's a soft assert I agree with because sensor values can fluctuate unpredictably. Unrecognized testbeds, though, indicate undefined behavior.

ptest/cases/adcs_checkout_case.py
```diff
@@ -0,0 +1,195 @@
...
+        self.soft_assert(([0 for x in range(self.havt_length)] == self.havt_read), 
+            "Disabling all devices failed")
+
+        # reset all devices
+        for x in range(self.havt_length):
+            self.ws(f"adcs_cmd.havt_reset{x}", True)
+        self.step()
+        print("Post resetting all devices:")
+        self.print_havt_read()
+        self.soft_assert((initial_up_devices == self.havt_read),
```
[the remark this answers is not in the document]
tanishqaggarwal: Actually, why isn't this a test case failure as well? If devices fail to come up that's a pretty bad sign about the health of the hardware. We shouldn't continue with tests that might damage it.

ptest/cases/adcs_checkout_case.py
```diff
@@ -0,0 +1,211 @@
...
+
+        # disable all devices
+        for x in range(self.havt_length):
+            self.ws(f"adcs_cmd.havt_disable{x}", True)
+
+        self.cycle()
+
+        self.logger.put("Post disabling all devices:")
+        self.print_havt_read()
+        self.soft_assert(all([x == 0 for x in self.havt_read]),
```
tanishqaggarwal: Same here, re: TestCaseFailure. If devices cannot be turned on/off that's a bad sign about the hardware's health.

kkrol27:
> I strongly think we should be using TestCaseFailure in more places. If the HAVT is not performing correctly, it's a strong sign hardware might be broken and that we shouldn't continue with more, potentially damaging tests.

I agree here with Tanishq that we should be using `TestCaseFailure` more often in this PR althought for some errors I much prefer the idea of a soft assert. Although, tbh we sorta need to see how the checkout behaves first before we'll really know what makes the most sense. So I'm not sure if it's the best use of our time to go back and forth on this right now. Perhaps setting up a call would be more effective.

tanishqaggarwal:
When running this test case for the first time on hardware today, Id much rather prefer earlier forced stops to continued execution. Forcing things to work correctly the first time by having hard barriers ensures greater correctness of our tests.

kkrol27:
> When running this test case for the first time on hardware today, Id much rather prefer earlier forced stops to continued execution. Forcing things to work correctly the first time by having hard barriers ensures greater correctness of our tests.

Fair that is true. We can always back the checks off but erring on the side of too many fatal errors is safer.

shihaocao:
Hm. After testing today, I think we're still a long way from full hardware reliability. There's still a lot of bugs within FSW as a whole.

As such, I think it might be worthwhile splitting ADCSCheckoutCase into two files, ADCSHITLCheckoutCase, which would explicitly be used for verifying that a PR doesn't break anything, and a normal ADCSCheckoutCase, who's sole purpose would be to report as much debug information as possible.

Especially after today, it was really useful to have ADCSCheckoutCase, as a diagnostic reporting tool, more than it was for a specific test that required passing. And more explicitly, I think the fact that it used `soft_assert()` was key to having all diagnostic information available, as many of the errors we encountered were multi-level and really nuanced.

As it stands for this PR, I really think we should keep it as `soft_assert()` because I'm using it _manually_ as a diagnostic tool the most.

tanishqaggarwal:
Makes sense to me, let's split into two cases

src/fsw/FCCode/MainControlLoop.cpp
```diff
@@ -99,13 +99,13 @@ void MainControlLoop::execute() {
 
     attitude_estimator.execute_on_time();
     mission_manager.execute_on_time();
+    dcdc_controller.execute_on_time();
```
shihaocao, line 102: This change is necessary, along with the ones in `MCL.hpp` to make sure the DCDC is on before ADCS commands are applied so that wheels can turn.

ptest/cases/adcs_checkout_case.py
```diff
@@ -0,0 +1,215 @@
...
+        self.ws("dcdc.ADCSMotor_cmd", True)
+
+        # Necessary so that motor commands are pre-empted by ADCS DCDC being on
+        self.cycle()
+
+        self.ws("adcs.state", self.adcs_states.get_by_name("point_manual"))
+        self.ws("adcs_cmd.rwa_mode", self.rwa_modes.get_by_name("RWA_SPEED_CTRL"))
+        self.ws("adcs_cmd.rwa_speed_cmd", [0,0,0])
+    
+        self.print_header("Finished Initialization")
```
shihaocao, line 67: Re-ordered this section

Issue #309, opened 2020-03-19
Add Build System and Macro Documentation


kkrol27:
Currently, we don't have much documentation about our use of the PlatformIO build system or the macros involved in flight software.

We should generate documentation to cover these main topics:

* Macros defined in the build system itself and their purpose (e.g. `PAN_LEADER`, `PAN_FOLLOWER`, `DESKTOP`, etc.).
* The major differences between Teensy builds and desktop builds (including compiler flags that aren't explicitly defined).
* Macros defined in C/C++ source code depending on those set by the builds system/compiler (e.g. custom cast macros or assertion macros).

tanishqaggarwal:
I kind of did some of this, just now, but also missed some. Writing this here so I remember to track this item more systematically

PR #336, opened 2020-03-29
Added Casting Macros that are Dependent on Compiler RTTI Support

### LINKED ISSUE #305: Add Dynamic Cast Support to FSW when Building for a Desktop
Currently, we have not been using `dynamic_cast<...>(...)`  because RTTI is disabled for teensy builds. However, this leaves us vulnerable to potentially undefined behavior that could be very difficult to debug. More specifically, the `find_*_field<...>(...)` functions in `ControlTask` look like this:

    template<typename U>
    InternalStateField<U>* find_internal_field(const char* field, const char* file, const unsigned int line) {
        InternalStateFieldBase* field_ptr = _registry.find_internal_field(field);
        check_field_exists(field_ptr, "internal", field);
        return static_cast<InternalStateField<U>*>(field_ptr);
    }

The issue here is if we specify an improper type `U`.

To help curb this, we should add support to use dynamics casting here for flight builds only.

kkrol27:
# Adding Casting Macros that are Dependent on the Presence of Compiler RTTI Support

Fixes #305.

### Summary of changes
- Added the `DYNAMIC_CAST(type, var)` and `STATIC_CAST(type, var)` macros.
- Updated `find_*_field` code to use dynamic casting if the compiler supports RTTI.
- Added the `check_template_field_exists` function to check at runtime if a dynamic casts succeeds upon field retrieval (obviously this only really does something if RTTI is enabled).

### Testing
Was able to build and run existing CI tests locally. I also double checked the preprocessor output of the following header:

    #include <common/casts.hpp>

    STATIC_CAST(float, 0.0);
    DYNAMIC_CAST(float, 0.0);

with the following four compiler invocations and everything checked out:

    g++ -std=gnu++14 -E              // Dynamic casting was enabled
    g++ -std=gnu++14 -E -fno-rtti    // Dynamic casting was disabled
    clang++ -std=c++14 -E            // Dynamic casting was enabled
    clang++ -std=c++14 -E -fno-rtti  // Dynamic casting was disabled

### Constants

NA

### Documentation Evidence

See inline documentation. Ideally we would also add documentation in tandem with [this issue](https://github.com/pathfinder-for-autonomous-navigation/FlightSoftware/issues/309).


kkrol27:
Unsure atm why exactly it's failing, but I'm seeing it locally as well with:

    pio run -e fsw_natrive_leader
    python test/test_real_fsw.py

kkrol27:
Actually, no scratch that I think those fields may be accessed in goofy ways somewhere. I'm only getting the error on a limited number of state fields even though many more are being accessed.

kkrol27:
Alright, I think I figured it out but have absolutely no idea how to fix it. The new EEPROM controller expects all three of those states listed above (`pan.state`, `pan.deployed`, `pan.sat_designation`) to be `unsigned int` state fields. However, just a quick grep show some issues with that assumption.

Just as examples:

```
@jezero ~/git/pan/FlightSoftware git:(kkrol/casts|+3) venv:(3.7.7)
% grep -r "pan.state" src/   
src/fsw/FCCode/MissionManager.cpp:    mission_state_f("pan.state", Serializer<unsigned char>(12)),
src/fsw/FCCode/mission_state_t.enum:// pan.state
src/flow_data.cpp:{1, "true", {"pan.state", "pan.sat_designation"}},
src/eeprom_configs.cpp:const std::vector<std::string> PAN::statefields = {"pan.state", "pan.deployed", "pan.sat_designation", "pan.cycle_no"};
src/flow_data.csv:Core state fields,true,pan.state,pan.deployed,pan.sat_designation
```
and

```
@jezero ~/git/pan/FlightSoftware git:(kkrol/casts|+3) venv:(3.7.7)
% grep -r "pan.deployed" src/                                                       
src/fsw/FCCode/MissionManager.cpp:    is_deployed_f("pan.deployed", Serializer<bool>()),
src/eeprom_configs.cpp:const std::vector<std::string> PAN::statefields = {"pan.state", "pan.deployed", "pan.sat_designation", "pan.cycle_no"};
src/flow_data.csv:Core state fields,true,pan.state,pan.deployed,pan.sat_designation
```

kkrol27:
Just going to make an issue to track this rather than dumping everything in the PR.

kkrol27:
I had to update the casting scheme slightly due to the error discussed [here](https://stackoverflow.com/questions/43897750/cannot-convert-from-pointer-to-base-class-to-pointer-to-derived-class). Not 100% sure why it didn't work before but does now, but, nonetheless, it works now haha

PR #345, opened 2020-03-29
fix downlink parser

### LINKED ISSUE #343: Downlink Event Parser does not temporarily save old values
Consider the following sequence of items in telemetry:
- A state field called `x`, which has a value of 5
- An event called `y` which has `x` as a data field, which occurred at time `t`, at which point `x had a value of 2

Rather than the final reported telemetry from the downlink parser for `x` having a value of 5, it will have a value of `2` because calling deserialize() for the event causes the value of `x` to be overwritten.

To solve this problem, the data fields of `y` should be saved prior to deserializing `y`, and then their values should be re-applied.

fatimayousuf:
# Fix Downlink Parser Bug

Fixes #343 

### Summary of changes
Record the original values of the event data fields before deserializing the event, and then reapplying the original values to the fields after the downlink is parsed

### Testing
In the test file, I changed the event data field (data1_f) from false to true after the event was signaled. I then tested that data1_f remained true even after the downlink was parsed.

### DIFF
--- a/src/gsw/parsers/src/DownlinkParser.cpp
+++ b/src/gsw/parsers/src/DownlinkParser.cpp
@@ -146,20 +146,34 @@ std::string DownlinkParser::process_downlink_packet(const std::vector<char>& pac
             for(ReadableStateFieldBase* field : flow->field_list) {
                 Event* event = registry.find_event(field->name());
                 if (event) {
+                    // Store the original values of the control cycle count and data fields
+                    unsigned int current_ccno = event->ccno->get();
+                    std::vector<bit_array> field_bits_original;
+                    for (ReadableStateFieldBase* data_field : event->_data_fields()) {
+                        data_field->serialize();
+                        field_bits_original.push_back(data_field->get_bit_array());
+                    }
+
                     const std::vector<bool>::iterator event_end_it =
                         frame_bits.begin() + event->get_bit_array().size();
                     
                     const std::vector<bool> event_bits(frame_bits.begin(), event_end_it);
                     event->set_bit_array(event_bits);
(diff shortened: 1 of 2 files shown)


src/gsw/parsers/src/DownlinkParser.cpp
```diff
@@ -146,21 +146,35 @@ std::string DownlinkParser::process_downlink_packet(const std::vector<char>& pac
             for(ReadableStateFieldBase* field : flow->field_list) {
                 Event* event = registry.find_event(field->name());
                 if (event) {
+                    // Store the original values of the control cycle count and data fields
+                    unsigned int current_ccno = event->ccno->get();
```
tanishqaggarwal: I don't think we need to restore the ccno stored inside the event, since it never gets overridden by anything.

fatimayousuf: The event ccno is a pointer to the control cycle count in the TimedControlTask class. Even though deserialize is only called from the ground, couldn't setting the control cycle count to value stored in the downlink (the last time the event was signaled) affect things? Maybe with ptest as we try sending uplinks and downlinks without a radio in the loop?

fatimayousuf: Also if I don't restore the control cycle count, then `TEST_ASSERT_EQUAL(tf.cycle_count_fp->get(), cycle_no);` in the test file won't pass

tanishqaggarwal: Ah, you're right, they're the same pointer. I see your point.

Issue #351, opened 2020-03-30
ptest documentation


afoarce:
Document updated functionality with parsetelem and telemetry file saving.

Issue #273, opened 2020-03-12
Ptest: store processed downlink somewhere


tanishqaggarwal:
Depends on #244.

Let's add an option into StateCmdPrompt that can actually process the returned telemetry file. The command could be called `parsetelem`: pass the most recent telemetry file into the DownlinkParser.

The DownlinkParser returns a JSON string with the available telemetry, which can then be displayed to the user. Some helpful tips on implementing this ability:
   - The downlink parser can be generated by running `pio run -e gsw_downlink_parser` in FlightSoftware. The binary ends up in `FlightSoftware/.pio/build/gsw_downlink_parser/program`
   - Integrating the DownlinkParser into the StateCmdPrompt requires starting the downlink parser as a subprocess. The state cmd prompt also needs to know where the Downlink Parser binary is located--you can do this by making the Downlink Parser a location specified in the config files (like the way Flight Software's location is specified in ptest/configs/fc_only_native.json)
   - Starting up the downlink parser will require creating a [subprocess](https://docs.python.org/2/library/subprocess.html) and maintaining it for the duration of the ptest testing. It feels wrong to have StateCmdPrompt own this subprocess since the class is purely meant to contain a user-facing frontend. Try to fit it into `RadioSession` instead, and have StateCmdPrompt's `parsetelem` call a similarly named function inside  `RadioSession` so that StateCmdPrompt acts as a client of RadioSession.

PR #352, opened 2020-03-30
Telemetry saving and parsing

### LINKED ISSUE #244: Ptest: save telemetry passed over USB into a file
If you currently send the command "ws telem.dump true" through ptest, ptest returns a hex representation of the produced telemetry snapshot of that control cycle, as follows:

    > telem
    Succeeded                                               (Completed in 567 us)
    > 
    [2020-03-11 22:14:41.980196] Received requested telemetry from spacecraft.
    \xc0\x00\x00\x00\x5a\x28

In order to do telemetry-in-the-loop testing without radio-in-the-loop, we need to be able to transact uplinks and downlinks over ptest. Let's add a feature to ptest that allows us to write this telemetry into a file instead of printing to console.

The relevant place to do this is [here](https://github.com/pathfinder-for-autonomous-navigation/FlightSoftware/blob/master/ptest/state_session.py#L99-L103):

    elif 'telem' in data:
        logline = f"[{data['time']}] Received requested telemetry from spacecraft.\n"
        logline += data['telem']
        print("\n" + logline)
        self.logger.put(logline, add_time = False)

We can augment this code to store the received telemetry in a timestamped file in the constructor argument `simulation_run_dir` (where simulation logs are stored.) Be sure that the telemetry is saved in a binary format, and not in the escaped ASCII format like the printed text above.

afoarce:
--------------------------

# Saving telemetry and processing downlinks

Fixes #244, #273  

### Summary of changes
- added feature to ptest that writes the telemetry data snapshot into a file when calling telem 
- added parsetelem option to ptest that parses the most recently written telemetry file and also outputs it to the user using RadioSession and a downlink parser subrocess
- made config file have location of downlink parser 

### Testing
Tested functionality through ptest and checking the written telemetry files for accuracy. Logs are posted in the Onedrive. 

### Documentation Evidence
- #351


ptest/radio_session.py
```diff
@@ -108,6 +118,21 @@ def write_state(self, field, val, timeout=None):
...
+        #get newest file
+        telem_files = glob.iglob(os.path.join(self.telem_save_dir, 'telem*'))
+        try:
+            newest_telem_file = max(telem_files, key=os.path.basename)
+        except ValueError:
+            return "No telemetry to parse."
+        self.console.write((newest_telem_file+"\n").encode())
+        telem_json_data = json.loads(self.console.readline().rstrip())
+        self.console.write((newest_telem_file+"\n").encode()) #twice to push response from downlinker parser
+        telem_json_data = json.loads(self.console.readline().rstrip())
```
tanishqaggarwal: I'm not sure if this behavior is correct. We don't want to force a response out of the downlink parser by sending the packet twice; the idea is that the downlink parser will send telemetry for the most recent _complete_ downlink frame once you send it a downlink packet from a new frame. The reason you're having to send it twice is because the example downlink frame that we have consists only of one packet.

tanishqaggarwal: Let me know if this comment doesn't make sense

ptest/configs/fc_only_native.json
```diff
@@ -1,18 +1,19 @@
 {
     "seed" : 2,
     "single_sat_sim" : true,
+    "downlink_parser_filepath" : "./.pio/build/gsw_downlink_parser/program",
```
tanishqaggarwal: You'll also need to modify ci.json

PR #361, opened 2020-04-01
Gswintegration

knk38:
--------------------------

# PR Title GSWintegration

### Summary of changes
- added JUnit testing with tutorial server
- connected flask server using HTTP Get and Post commands to send telemetry
- fixed some Readme stuff and other small things with package.json

### Testing

I tested the feature by sending example data that was generated by decrementing a state value. I also tested it by sending an example telemetry value to the ground software server. The test files are shown in the /test folder with example telemetry points defined in dictionary.json and given values in spacecraft,js's telemetry function.

### Documentation Evidence

I have in line documentation on methods that I added in spacecraft.js and the test/server.js


src/gsw/MCT/test/example-server/spacecraft.js
```diff
@@ -0,0 +1,85 @@
+/*
```
tanishqaggarwal, line 1: Do you need a separate spacecraft for testing? I thought the idea was that the underlying telemetry service would change, but everything else remains nearly the same

tanishqaggarwal: Same for the other files, like index.html, http.js, etc.

fatimayousuf: Yeah I have the same question. It seems like a lot of the files across the test folder and the public folder are duplicates or nearly exactly the same, such as the dictionary-plugin, historical telemetry plugin, http.js, etc. Why did you make a new test folder instead of editing the public folder?

src/gsw/MCT/test/mctTest.js
```diff
@@ -0,0 +1,29 @@
+/**
+ * Basic implementation of a history and realtime server.
+ */
+
+var Spacecraft = require('./example-server/spacecraft');
+var RealtimeServer = require('./example-server/realtime-server');
+var HistoryServer = require('./example-server/history-server');
+var StaticServer = require('./example-server/static-server');
```
fatimayousuf, line 8: I'm a bit confused. Are you using the server files in MCT/test/example-server or are you using the server files in MCT/server-files? You use MCT/server-files is package.json but MCT/test/example-server here. They're both really similar and some are exact duplicates of each other.

src/gsw/MCT/test/example-server/spacecraft.js
fatimayousuf: Yeah I have the same question. It seems like there are a lot of duplicate or near duplicate files across the test folder and the public folder. Why not edit the public folder instead of creating a new test folder?

knk38: the problem is that that spacecraft file has a different generate telemetry function than the other one, what i could do is put both methods in one file, but im not sure it would be accessible

src/gsw/MCT/test/mctTest.js
knk38: oh yeah ur right i will fix this

src/gsw/MCT/test/example-server/spacecraft.js
tanishqaggarwal: You could probably use inheritance to your benefit. Maybe create one common Spacecraft classes, and have two different child classes that implement "generate telemetry" differently?

PR #366, opened 2020-04-06
Remove const_cast

### LINKED ISSUE #325: const_cast can break things if compiler optimizes too much
This might be what caused the flaky Quake issue we saw before.
Consider removing const_cast and instead making fields mutable.

@athena255

tanishqaggarwal:
Fixes #325

### Summary of changes
- Removed const_cast from repository

### Testing
Existing unit tests should be sufficient.

### Constants
There just seems to be a line number change.

### Documentation Evidence
Having no documentation for this is sufficient since this is an implementation-only change and does not affect code requirements.


src/common/StateFieldTypes.inl
```diff
@@ -32,7 +32,7 @@ class SerializableStateFieldBase : virtual public StateFieldBase {
     virtual bool deserialize(const char *val) = 0;
     virtual const char *print() const = 0;
     virtual size_t bitsize() const = 0;
-    virtual const bit_array& get_bit_array() const = 0;
+    virtual bit_array& get_bit_array() = 0;
```
kkrol27, line 35: Why not make a `const` override here? You could implement one of them with `const_cast` to avoid duplicate code.

tanishqaggarwal:
I'm going to refactor this with the infrastructural change and then open later.

PR #370, opened 2020-04-07
Properly fix const cast

### LINKED ISSUE #325: const_cast can break things if compiler optimizes too much

tanishqaggarwal:
Fixes #325

### Summary of changes
Removed const_cast from repository

### Testing
Existing unit tests should be sufficient.

### Constants
There just seems to be a line number change.

### Documentation Evidence
Having no documentation for this is sufficient since this is an implementation-only change and does not affect code requirements.


src/fsw/FCCode/ADCSBoxController.cpp
```diff
@@ -125,7 +125,7 @@ void ADCSBoxController::execute(){
             send_cmd_table = true;
 
             // clear the state field now that it's loaded into temp_cmd_table
-            const_cast<WritableStateField<bool>*>(havt_cmd_disable_vector_fp[idx])->set(false);
+            havt_cmd_disable_vector_fp[idx]->set(false);
```
shihaocao, line 128: Was this change purely so that we don't have `const_cast` in our code?

tanishqaggarwal: Yep

src/common/Serializer.hpp
```diff
@@ -184,6 +184,7 @@ class SerializerBase : public SerializerType {
      * @return const bit_array& Reference to internal bit array.
      */
     const bit_array& get_bit_array() const { return serialized_val; }
+    bit_array& get_bit_array() { return serialized_val; }
```
kkrol27, line 187: I'm not familiar enough with the inheritance hierarchy, but to save a couple lines of code you could define the `const` version in the parent class once with `const_cast<...>` like this:

    const bit_array& get_bit_array() const { return const_cast<type &>(*this).get_bit_array(); }

You then only need to define the non-constant version in the child classes.

src/common/StateFieldTypes.inl
```diff
@@ -33,6 +33,7 @@ class SerializableStateFieldBase : virtual public StateFieldBase {
     virtual const char *print() const = 0;
     virtual size_t bitsize() const = 0;
     virtual const bit_array& get_bit_array() const = 0;
+    virtual bit_array& get_bit_array() = 0;
```
kkrol27, line 36: I think the general definition of `get_bit_array() const` I mentioned earlier would go here.

src/common/Serializer.hpp
tanishqaggarwal: Serializer.hpp is the top-level class. StateFieldTypes is a separate hierarchy.

src/common/StateFieldTypes.inl
tanishqaggarwal: Fixed

Issue #373, opened 2020-04-07
Long Orbit Propagator.


nhz2:
Long Orbit propagator that can run "asynchronously" (aka take more than one control cycle to finish).

nhz2:
I have a version of this in the `Orbit` class in psim.

Issue #372, opened 2020-04-07
Orbit validity checks.


nhz2:
Update `Orbit::_check_validity()` to ensure 
A valid orbit has finite and real position and velocity, is in low
earth orbit, and has a reasonable time stamp (within 20 years of pan epoch). The validity check should not reject gps readings due to reasonable noise.

Issue #339, opened 2020-03-29
Common Software Unit Tests failing on Teensy


tanishqaggarwal:
See attached build log from running `pio test -e teensy35_common &> build.txt`. GPSTime is failing, as are some serializers.

Can someone corroborate if the same issues are seen on a Teensy 3.6?

Note: to get all of the tests to run in sequence, increase the serial delays at the beginning of each test to 10000 ms.

[build.txt](https://github.com/pathfinder-for-autonomous-navigation/FlightSoftware/files/4399561/build.txt)

tanishqaggarwal:
Shihao corroborated that the same issues are seen on a Teensy 3.6

PR #374, opened 2020-04-07
Orbit class

nhz2:
# Orbit class

### Interface with flight software
I plan on using Orbit class as a state field in the orbit estimation control tasks. Also the best estimate of the satellites current orbit will have to be downlinked and sent to the other satellite. This hasn't been fully implemented yet, currently the Orbit class just has a position velocity and time, and can be propagated for short updates. Right now it isn't used in the rest of flight software.

### Summary of changes
The goal is to create a class to handle most of the logic related to storing, serializing, propagating, and checking validity of an orbit. This class will be used by the next version of the orbit estimator. Another goal is to specify, control, and minimize the number of calls to the gravity function, because one call could take about 2ms on a teensy.

- Added an Orbit class
- Added `-Wno-missing-braces` on native so I can compile on my mac.

The next updates to Orbit will add: 
- More validity checks. #372 
- Long Orbit propagator that can run "asynchronously" (aka take more than one control cycle to finish). #373
- Python or Matlab wrapper.
- Bit compression/ Orbit statefield.

### Testing
I have unit testing in `test/test_common_orbit/test_orbit.cpp`

### Constants
I add some Orbit related constants.

### Documentation Evidence
Inline documentation is sufficient.


nhz2:
> I haven't done a full read through of all the code, but I'm just curious why this is getting written in flight software as opposed to PSim?
> 
> I agree that the state field, serialization, and compression logic should live in flight software obviously, but it would seem to make more sense to me if we consolidated the core, GNC functionality like orbit propagation and state estimation in PSim so that it can be easily integrated with the MATLAB sim. Then do something like extend the orbit class in flight software to implement the serialization, include the flight software definition of GPS time, etc. PSim shouldn't have to care about all the bit fiddling technicalities and FSW shouldn't need to care about mathy implementation details as much.
> 
> Taking some approach like this would prevent odd circular dependencies in our repositories, I think help make things clearer, and conform to the development paradigm we sort of have going right now.

I 

> I haven't done a full read through of all the code, but I'm just curious why this is getting written in flight software as opposed to PSim?
> 
> I agree that the state field, serialization, and compression logic should live in flight software obviously, but it would seem to make more sense to me if we consolidated the core, GNC functionality like orbit propagation and state estimation in PSim so that it can be easily integrated with the MATLAB sim. Then do something like extend the orbit class in flight software to implement the serialization, include the flight software definition of GPS time, etc. PSim shouldn't have to care about all the bit fiddling technicalities and FSW shouldn't need to care about mathy implementation details as much.
> 
> Taking some approach like this would prevent odd circular dependencies in our repositories, I think help make things clearer, and conform to the development paradigm we sort of have going right now.

I put this in FlightSoftware because FlightSoftware and psim have slightly different testing infrastructure and constant tracking infrastructure, and I didn't want to have to deal with learning two different testing and constant tracking infrastructures when FlightSoftware's infrastructure is working well and does very similar things. Also the Orbit class depends on code in flight software, specifically `constant_tracker.hpp`. Eventually I would like to make a Orbit state field, like GPSTime. Also, the final orbit estimator will probably be a control task, which doesn't exist in psim. For MATLAB integration I am fine with moving the MATLAB sim directory into the FlightSoftware repo, or just giving the MATLAB sim a path to FightSoftware like in old ptest.

src/fsw/FCCode/Orbital/Orbit.h
```diff
@@ -0,0 +1,483 @@
+/*
```
tanishqaggarwal, line 1: Can you separate interface from implementation? There are a lot of functions and values, it would be easier to read and understand this file if they were listed separately.

test/test_fsw_orbit/test_orbit.cpp
```diff
@@ -0,0 +1,419 @@
+#include <unity.h>
```
tanishqaggarwal, line 1: Did you get to run this test file on Teensy to verify that it passes? If it doesn't we should add it to #339

tools/jacobian_autocoder.py
```diff
@@ -0,0 +1,60 @@
+# Nathan Zimmerberg (nhz2@cornell.edu)
```
tanishqaggarwal, line 1: What is the use case for this script?

test/test_fsw_orbit/test_orbit.cpp
```diff
@@ -0,0 +1,419 @@
...
+    
+    // 0.1 dt 100 sec test vs grace data
+    double junk;
+    for(int i=0; i<1000;i++){
+        y1.shortupdate(100'000'000,earth_rate_ecef,junk);
+    }
+    lin::Vector3d vfinal = y1.vecef();
+    lin::Vector3d rfinal = y1.recef();
+    //lin::Vector3d rfinal = y1.recef();
+    TEST_ASSERT_TRUE(y1.nsgpstime()==t1+100'000'000'000LL);
```
tanishqaggarwal, line 242: Apostrophes for separation is pretty cool

test/test_fsw_orbit/test_orbit.cpp
```diff
@@ -0,0 +1,419 @@
...
+
+    TEST_ASSERT_FLOAT_WITHIN(10.0, 0.0,(e2-e1));
+
+    //Test invalid orbit returns NAN
+    Orbit x;
+    TEST_ASSERT_FALSE(x.valid());
+    TEST_ASSERT_FLOAT_IS_NAN(x.specificenergy({0.0,0.0,gnc::constant::earth_rate_ecef_z}));
+}
+
+void test_shortupdate() {
```
tanishqaggarwal, line 215: This is a very long test function and I don't really get a clear idea of what it's testing. It would be helpful to have a series of long-form comments explaining the test steps and/or separation into several functions. Code readers should be able to make a mental model very quickly about code from the comments, and not have to read through every line to understand the test's functionality.

test/test_fsw_orbit/test_orbit.cpp
```diff
@@ -0,0 +1,419 @@
...
+                TEST_MESSAGE(errormessage);\
+                sprintf(errormessage, "    expected-actual: %.10e, %.10e, %.10e",expected(0)-actual(0),expected(1)-actual(1),expected(2)-actual(2));\
+                TEST_MESSAGE(errormessage);\
+                sprintf(errormessage, "    norm(expected-actual): %.10e",lin::norm(expected-actual));\
+                TEST_MESSAGE(errormessage);\
+                sprintf(errormessage, "    max error: %.10e",delta);\
+                TEST_MESSAGE(errormessage);\
+                TEST_FAIL();\
+            }\
+           } while(0)
```
tanishqaggarwal, line 89: Could you put this in test/custom_assertions.hpp?

test/test_fsw_orbit/test_orbit.cpp
```diff
@@ -0,0 +1,419 @@
...
+#else
+#include <Arduino.h>
+#define notcompiletime (millis()%2)
+#endif
+
+/**
+ * Lin Tensor Printing function
+ * */
+template<typename T>
+void printtensor(T x){
```
tanishqaggarwal, line 25: We should consider making these two functions part of the lin library? Especially determinant

src/fsw/FCCode/Orbital/Orbit.h
```diff
@@ -0,0 +1,483 @@
...
+#include <limits>
+#include <lin/core.hpp>
+#include <lin/generators.hpp>
+#include <common/constant_tracker.hpp>
+
+#include "geograv.hpp"
+#include "GGM05S.hpp"
+
+
+TRACKED_CONSTANT(constexpr static int, PANGRAVORDER,40);
```
tanishqaggarwal, line 44: There is `TRACKED_CONSTANT_C` that automatically adds the constexpr designation. Up to you if you want to use it.

I recommend making these constants static members of the class. Having floating global variables when not absolutely necessary is not a good idea

src/fsw/FCCode/Orbital/Orbit.h
```diff
@@ -0,0 +1,483 @@
...
+        //note if position is NAN, these checks will be false.
+        if (r2<MAXORBITRADIUS*MAXORBITRADIUS && r2>MINORBITRADIUS*MINORBITRADIUS){
+            _valid= true;
+        } else {
+            _valid= false;
+            _recef= lin::nans<lin::Vector3d>();
+            _vecef= lin::nans<lin::Vector3d>();
+            _ns_gps_time= 0;
+        }
+    }
```
tanishqaggarwal, line 127: Why not just combine these two functions?

src/fsw/FCCode/Orbital/Orbit.h
```diff
@@ -0,0 +1,483 @@
...
+    lin::Vector3d vecef() const{
+        return _vecef;
+    }
+
+    /** Return true if the Orbit is valid.
+     * A valid orbit has finite and real position and velocity, is in low
+     * earth orbit, and has a reasonable time stamp (within 20 years of pan epoch).
+     * The validity check should not reject 
+     * gps readings due to reasonable noise of:
+     * TODO add max expected gps error
```
tanishqaggarwal, line 103: We should try to track TODOs in issue tickets

src/fsw/FCCode/Orbital/Orbit.h
```diff
@@ -0,0 +1,483 @@
...
+     * grav calls: 0
+     * @param[in] r_half: The sats position in ecef0 at the half step (m).
+     * @param[in] dt: Time step used (s). 
+     * @param[out] jac: 
+     * The jacobian of y=f(x) where x and y are vectors
+           [r_ecef;
+            v_ecef;] in m and m/s.
+       and f is the shortupdate function.
+     */
+    static void _jacobian_helper(const lin::Vector3d& r_half, const  double& dt, lin::Matrix<double, 6, 6>& jac){
```
tanishqaggarwal, line 241: The implementation of a function like this should definitely go in a separate file. Having it be inline with a bunch of other function definitions detracts from the documentative purpose of a header file.

tanishqaggarwal:
We should figure out the best location for GNC code in FSW over our chat during the weekend, before merging this in

test/test_fsw_orbit/test_orbit.cpp
nhz2: Yeah, I think @kkrol27  is looking into the determinant, this recursive implementation is very computationally inefficient for large matrices though( I think O(n!)).

test/test_fsw_orbit/test_orbit.cpp
nhz2: Yes the test passes on teensy 3.6 

```
(venv) nathan@Nathans-MBP FlightSoftware (Orbit-class) $ pio test -e fsw_teensy36_hitl_leader -f "test_fsw_orbit"
Verbose mode can be enabled via `-v, --verbose` option
Collected 38 items

Processing test_fsw_orbit in fsw_teensy36_hitl_leader environment
----------------------------------------------------------------------------------------
Building...
CorePackageManager: Installing tool-scons @ ~3.30102.0
tool-scons @ 3.30102.0 has been successfully installed!
Uninstalling tool-scons @ 3.30101.0: 	[OK]
Uploading...
Programming............................................................................
Testing...
If you don't see any output for the first 10 secs, please reset board (press reset button)

test/test_fsw_orbit/test_orbit.cpp:398:test_basic_constructors	[PASSED]
test/test_fsw_orbit/test_orbit.cpp:399:test_applydeltav	[PASSED]
test/test_fsw_orbit/test_orbit.cpp:400:test_calc_geograv	[PASSED]
test/test_fsw_orbit/test_orbit.cpp:401:test_specificenergy	[PASSED]
test/test_fsw_orbit/test_orbit.cpp:402:test_shortupdate	[PASSED]
-----------------------
5 Tests 0 Failures 0 Ignored
========================= [PASSED] Took 263.74 seconds =================================
```

tools/jacobian_autocoder.py
nhz2: I added more about the use to the comments

src/fsw/FCCode/Orbital/Orbit.h
nhz2: The validity needs to be checked every time the orbit gets made or a deltaV is applied, but the validity needs to be read all the time, so _check_validity stores the result so it doesn't have to recompute the validity all the time. It's an implementation detail.

src/fsw/FCCode/Orbital/Orbit.h
nhz2: Yeah, though its tricky to get the compiler to do the memory optimizations on the gravity model (ignore higher order terms) at compile time. I'll try and make it a constexpr static member and see what happens.

nhz2: I can't seem to get it to compile with constexpr static members, I'm not familiar with how those work.

src/fsw/FCCode/Orbital/Orbit.h
nhz2: If you want the interface run `doxygen` in the docs directory and open one of the html files. The Orbit class interface should be summarized there.

tools/jacobian_autocoder.py
nhz2: I also moved this script into the `JacobianHelpers` sub directory

src/fsw/FCCode/Orbital/Orbit.h
nhz2: I moved the implementation into the `JacobianHelpers` sub directory

src/fsw/FCCode/Orbital/Orbit.h
tanishqaggarwal: Code should be as self-documenting as possible. Having clean interface files accomplishes that

test/test_fsw_orbit/test_orbit.cpp
nhz2: I refactored this test into lots of smaller sub tests.

nhz2:
This is moving over to psim

PR #390, opened 2020-04-13
Send ptest parsed telemetry to gsw database

### LINKED ISSUE #274: Ptest: allow processed downlink to be sent to ground software
Depends on #273.

Instead of just printing  the JSON-processed downlink to the ptest console, implement an option called `dbtelem` (Store Telemetry in Database) that sends the most recently DownlinkParser-parsed telemetry file from Flight Software into the ground software database. 

The ground software server's hostname and port is specified in `server_keys.json`, and has an HTTP endpoint called `send-telemetry` that accepts (time,field,value) tuples. Start up src/gsw/main.py and go to `http:://localhost:5000/swagger` to see how this endpoint works.

afoarce:
# Send ptest parsed telemetry to gsw database

Fixes #274 

### Summary of changes
- Added dbtelem command to ptest which will store the most recent telemetry frame into the ground software database.
- Added ".base" to fault names in order for Elasticsearch to successfully map the statefields

### Testing
Tested through ptest using ground software's server and Kibana for viewing data. In ptest, running telem and dbtelem (twice to get output from downlink parser) will show the statefield json data and then put each statefield into the database. Using either the SwaggerUI API or the Kibana interface, the data can be viewed.

### Constants
No constants changes.

### Telemetry
As summarized above, statefield names were changed to ensure they worked with Elasticsearch (e.g. creating gomspace.get_hk becomes gomspace.get_hk.base in the registry due to existence of gomspace.get_hk.override).  This change is necessary because Elasticsearch will map the base statefield as an object and then try to overwrite it as a nested type when it sees it has subfields, which gives an error.

### Documentation Evidence
Documented command in cmdprompt.py


ptest/cmdprompt.py
```diff
@@ -148,6 +148,17 @@ def do_telem(self, args):
         write_succeeded = "Succeeded" if write_succeeded else "Failed"
         print(f"{write_succeeded} \t\t\t\t\t\t(Completed in {elapsed_time} us)")
     
+    def do_dbtelem(self, args):
+        '''
+        Store Telemetry in Database
+        '''
+        args = args.split()
+        if len(args) != 0:
+            print('parsetelem takes no args')
```
tanishqaggarwal, line 157: 'dbtelem takes no args'

ptest/radio_session.py
```diff
@@ -130,6 +150,24 @@ def parsetelem(self):
...
                 telem_json_data = telem_json_data["data"]
         return telem_json_data
+    
+    def dbtelem(self):
+        jsonObj = self.parsetelem()
+        if not isinstance(jsonObj, dict):
+            print(jsonObj)
+            return False
+        failed = False
+        print(jsonObj)
```
tanishqaggarwal, line 160: Let's remove this print

ptest/cmdprompt.py
```diff
@@ -159,6 +159,22 @@ def do_parsetelem(self, args):
...
+        if len(args) != 0:
+            print('dbtelem takes no args')
+            return
+        else:
+            start_time = timeit.default_timer()
+            successful_upload = self.devices['FlightControllerRadio'].dbtelem()
+            elapsed_time = int((timeit.default_timer() - start_time) * 1E6)
+
+            write_succeeded = "Succeeded" if successful_upload else "Failed"
+            print(f"{write_succeeded} \t\t\t\t\t\t(Completed in {elapsed_time} us)")
```
tanishqaggarwal, line 176: Weird that this doesn't print when running.

ptest/cmdprompt.py
```diff
@@ -159,6 +159,22 @@ def do_parsetelem(self, args):
...
+        '''
+        Store Telemetry in Database
+        '''
+        args = args.split()
+        if len(args) != 0:
+            print('dbtelem takes no args')
+            return
+        else:
+            start_time = timeit.default_timer()
+            successful_upload = self.devices['FlightControllerRadio'].dbtelem()
```
tanishqaggarwal, line 172: Same thing here, re: hardcoding the device name

ptest/radio_session.py
```diff
@@ -153,13 +151,20 @@ def parsetelem(self):
     
     def dbtelem(self):
         jsonObj = self.parsetelem()
+        if not isinstance(jsonObj, dict):
+            print(jsonObj)
```
tanishqaggarwal, line 155: There's a print here that should probably be removed

afoarce: Should we print something else if theres not telemetry to parse or if the downlink producer returns null? Or not print at all?

tanishqaggarwal: Hm yeah some kind of descriptive error message would be nice

ptest/radio_session.py
```diff
@@ -130,6 +148,23 @@ def parsetelem(self):
...
                 telem_json_data = telem_json_data["data"]
         return telem_json_data
+    
+    def dbtelem(self):
+        jsonObj = self.parsetelem()
+        if not isinstance(jsonObj, dict):
+            print(jsonObj)
+            return False
+        failed = False
+        for field in jsonObj:
```
tanishqaggarwal, line 158: We can further simplify the configs structure in #413 if we could have this make an HTTP call to the telemetry webservice rather than directly modifying the elasticsearch database.

tanishqaggarwal: Alternatively, we could hardcode the server and port number for elasticsearch into this function, since it's unlikely to change in local testing.

tanishqaggarwal:
```
switchcomp FlightController
telem
switchcomp FlightControllerFollowerRadio
dbtelem
```

PR #392, opened 2020-04-14
Prop fixes

athena255:
# Propulsion Driver minor changes

### Summary of changes
- is_functional now returns true only if DCDC SpikeAndHold is enabled
- made tracked constants for numeric constants from PropulsionSystem.cpp and moved them to PropulsionSystem.hpp
- changed the constructor of Tank to not take any parameter
- the can_enter() function for PropState_Idle in PropController.cpp has been changed to return true when compiled for DESKTOP; otherwise, it returns is_functional
- added tests for is_functional and to debug HOOTL problems

### Testing
- added a test for is_functional in test_prop.cpp
- added tests for setup in test_prop.cpp  that check that pinMode has been called on the relevant pins and that valve pins are high when prop is firing. This should help debug HOOTL tests. 

### Constants
- Added constants for Tank2 valve pins (valve1_pin, valve2_pin, etc)
- Added constant for Tank1 valve pins (valve_primary_pin, valve_backup_pin)
- Added constants  for temperature and pressure sensor pins

### Telemetry
Nothing

### Documentation Evidence
N/A


src/fsw/FCCode/Drivers/PropulsionSystem.hpp
```diff
@@ -234,6 +233,9 @@ class _Tank1 : public Tank {
         static _Tank1 Instance;
         return  Instance;
     }
+    TRACKED_CONSTANT_SC(uint8_t, valve_primary_pin, 27);
+    TRACKED_CONSTANT_SC(uint8_t, valve_backup_pin, 28);
+    TRACKED_CONSTANT_SC(uint8_t, tank1_temp_sensor_pin, 21);
```
tanishqaggarwal, line 238: I'm surprised that the constants checker isn't complaining about these constants not getting added to the `constants` file. Could you run `tools/constants_reporter.py` to add them?

test/test_prop/test_prop.cpp
```diff
@@ -12,9 +12,81 @@ using namespace Devices;
...
+uint8_t const p5 = 27; 
+uint8_t const p6 = 28;
+
+// DCDC enable
+uint8_t const dcdc = 25;
+
+void test_setup()
+{
+    // Valve pins should all be set to OUTPUT
+    TEST_ASSERT_TRUE(is_output(p1));
```
tanishqaggarwal, line 38: Could you describe how this test works on desktop? Seems like the digitalWrite and read functions would require arduino. Or is it not meant to run on desktop?

PR #394, opened 2020-04-14
Piksi fault management

### LINKED ISSUE #210: Piksi fault management.

fatimayousuf:
# Piksi Fault Management

Fixes #210 

### Summary of changes
1) Finished Piksi fault handler. Moves the satellite to standby if:
- the Piksi has not sent data in a long time (Piksi's state is dead)
- we entered a close approach state but haven't gotten CDGPS in X time (X is a ground-configurable wait initially set to 24 hours)
- we got CDGPS during close approach, but are still in close approach and haven't received CDGPS in the last Y hours (Y=3 initially but is ground-configurable)

2) Added two methods to clock manager to convert between control cycle counts and system time

### Testing
I tested each of the cases listed above in the Piksi Fault Handler test file. 

### Documentation Evidence
I'll do that in a separate PR when this finalized and is merged in


src/fsw/FCCode/PiksiFaultHandler.hpp
```diff
@@ -4,6 +4,9 @@
 #include "FaultHandlerMachine.hpp"
 #include "piksi_mode_t.enum"
 
+TRACKED_CONSTANT_SC(unsigned int, default_no_cdgps_max_wait, PAN::one_day_ccno);
+TRACKED_CONSTANT_SC(unsigned int, default_cdgps_delay_max_wait, PAN::one_day_ccno/8);
```
tanishqaggarwal, line 8: Can you make these part of the class? Having global variables floating around is not great

src/fsw/FCCode/PiksiFaultHandler.cpp
```diff
@@ -44,14 +50,14 @@ fault_response_t PiksiFaultHandler::check_cdgps() {
 
     // Recommend moving to standby if we haven't recieved any readings in X time since 
     // moving to close approach state
-    if (close_appr_time > last_fix_time && duration > no_cdgps_max_wait_f.get()) {
+    if (close_appr_time >= last_fix_time && duration > no_cdgps_max_wait_f.get()) {
```
tanishqaggarwal, line 53: If both times are equal, I think we should test if the fault response still gets triggered?

src/fsw/FCCode/PiksiFaultHandler.cpp
```diff
@@ -5,23 +5,29 @@
...
 
-        piksi_state_fp = find_readable_field<unsigned int>("piksi.state", __FILE__, __LINE__);
+        piksi_state_fp = find_readable_field<unsigned char>("piksi.state", __FILE__, __LINE__);
         mission_state_fp = find_writable_field<unsigned char>("pan.state", __FILE__, __LINE__);
         last_fix_ccno_fp  = find_internal_field<unsigned int>("piksi.last_fix_ccno", __FILE__, __LINE__);
         enter_close_appr_time_fp = find_internal_field<unsigned int>("pan.enter_close_approach_ccno", __FILE__, __LINE__);
     }
 
 fault_response_t PiksiFaultHandler::execute() {
+    if (!fault_handler_enabled_f.get()) return fault_response_t::none;
```
tanishqaggarwal, line 29: Let's add a test to verify this flag does the right thing. I know the behavior is pretty simple and should work, but the idea behind testing everything is that the unit test serves as a "press all the levers" check.

src/fsw/FCCode/PiksiFaultHandler.cpp
```diff
@@ -0,0 +1,65 @@
...
+    if (mission_state == mission_state_t::follower_close_approach || 
+            mission_state == mission_state_t::leader_close_approach) {
+        return check_cdgps();
+    }
+    
+    return fault_response_t::none;
+}
+
+fault_response_t PiksiFaultHandler::check_cdgps() {
+    unsigned int close_appr_time = enter_close_appr_time_fp->get();
```
shihaocao, line 47: We should default `enter_close_appr_time_fp` to the maximum possible value of int to prevent unexpected behavior.

shihaocao: Nevermind, it's impossible to enter check_cdgps() without having first set an `enter_close_approach_ccno`.

src/fsw/FCCode/PiksiFaultHandler.cpp
```diff
@@ -0,0 +1,65 @@
...
+        enter_close_appr_time_fp = find_internal_field<unsigned int>("pan.enter_close_approach_ccno", __FILE__, __LINE__);
+    }
+
+fault_response_t PiksiFaultHandler::execute() {
+    if (!fault_handler_enabled_f.get()) return fault_response_t::none;
+
+    piksi_mode_t piksi_state = static_cast<piksi_mode_t>(piksi_state_fp->get());
+    mission_state_t mission_state = static_cast<mission_state_t>(mission_state_fp->get());
+
+    if (piksi_state == piksi_mode_t::dead) {
```
shihaocao, line 34: I was thinking about actually just removing the `dead` mode entirely and replacing it with a Fault that way its persistence can be adjusted, and we can also suppress/override it. @tanishqaggarwal @fatimayousuf I could quickly add that?

tanishqaggarwal: We could do that, but I don't think it wouldn't necessarily be quick...let's ticket it and do it later if we have time?

shihaocao: Sure

src/fsw/FCCode/PiksiControlTask.cpp
```diff
@@ -9,10 +9,11 @@ PiksiControlTask::PiksiControlTask(StateFieldRegistry &registry,
     pos_f("piksi.pos", Serializer<d_vector_t>(0,100000,100)),
     vel_f("piksi.vel", Serializer<d_vector_t>(0,100000,100)),
     baseline_pos_f("piksi.baseline_pos", Serializer<d_vector_t>(0,100000,100)),
-    current_state_f("piksi.state", Serializer<unsigned int>(4)),
+    current_state_f("piksi.state", Serializer<unsigned char>(4)),
```
shihaocao, line 12: could you update the serializer from a 4 to a 10? The maximum possible value of piksi state is 10.

src/fsw/FCCode/PiksiFaultHandler.cpp
```diff
@@ -0,0 +1,69 @@
...
+            mission_state == mission_state_t::leader_close_approach) {
+        return check_cdgps();
+    }
+    
+    return fault_response_t::none;
+}
+
+fault_response_t PiksiFaultHandler::check_cdgps() {
+    unsigned int close_appr_time = enter_close_appr_time_fp->get();
+    unsigned int last_fix_time = last_fix_ccno_fp->get();
```
shihaocao, line 52: Hm, last_fix_time is updated even if we have SPP in PiksiController. I think an additional field, of perhaps last_cdgps_fix_time, is necessary. Otherwise, the code below is "satisfied" even if we only have SPP

shihaocao: Perhaps we should clarify under which conditions we want to go to stand by? Is it on loss of FixedRTK, Any RTK, or all GPS entirely? @fatimayousuf @tanishqaggarwal

shihaocao: For what its worth, i think we should exit close approach (to standby) as soon as we lose FixedRTK

tanishqaggarwal: I agree

src/fsw/FCCode/PiksiControlTask.hpp
```diff
@@ -39,6 +39,9 @@ class PiksiControlTask : public TimedControlTask<void>
     // System time of last good Piksi reading
     InternalStateField<sys_time_t> last_fix_time_f;
 
+    // Control cycle of last good Piksi reading
+    InternalStateField<unsigned int> last_rtkfix_ccno_f;
```
shihaocao, line 43: Shouldn't these both be `sys_time_t`? Or both `unsigned int`?

src/fsw/FCCode/PiksiFaultHandler.cpp
shihaocao: I think `last_fix_time` still needs to be changed to `last_fixed_rtk_time`, or some name like that. And then it should also pull from the new `last_rtkfix_ccno_f("piksi.last_rtkfix_ccno")` state field so that the below logic is activated whenever we lose fixed RTK.

src/fsw/FCCode/PiksiControlTask.hpp
tanishqaggarwal: unsigned int since it's a ccno. Control cycle count is a much better way to track time across long durations since sys_time_t overflows after 1.5 hours

src/fsw/FCCode/PiksiFaultHandler.cpp
fatimayousuf: `last_fix_time` does refer to the `"piksi.last_rtkfix_ccno"` statefield btw, but I'll change the name of the variable too

PR #397, opened 2020-04-15
Added Fixed PPT Mode Guard

shihaocao:
# Fixed PPT Mode Guard

We should only set the PV voltages when we're in fixed PPT mode. Note this bug fix doesn't stop the not charging issue, it was merely discovered during testing.

### Summary of changes
- Added an if statement guard
- Modified `set_pv_volt()` to `set_ppt_mode()`, and updated radiation bit check

This PR
### Testing
Shown to work on EDU sat.

### Constants
Describe any important diffs to the `constants` file in the root level of the repository. Ensure that any constants you added to the code successfully made it into the `constants` file. If not, wrap the missing constants with `TRACKED_CONSTANT` macros and fix `constants_reporter.py` if necessary.

### Telemetry
N/A

### Documentation Evidence
Updated ReadTheDocs


shihaocao:
Closed because further testing is required.

test/test_fsw_gomspace/test_fsw_gomspace.cpp
```diff
@@ -308,7 +308,7 @@ void test_task_execute() {
     tf.pv2_output_cmd_fp->set(2000);
     tf.pv3_output_cmd_fp->set(3000);
 
-    tf.ppt_mode_cmd_fp->set(0); // 1 is MPPT, the hardware default
+    tf.ppt_mode_cmd_fp->set(0); // 1 is MPPT, 0 is the hardware default
```
fatimayousuf, line 311: Not that it really matters in the context of your pr, but from what I recall, whenever I did HITL testing on the GomSpace, the PPT mode was initially at 1. I'm not totally sure though, it's been a while XD

shihaocao: oh, the gomspace manual said "0" is the hardware default, but I think it can actually change if you write settings to it (I sometimes saw `2` come up initially). I think we should do further testing lol

shihaocao:
Ptested, Old ADCSCheckoutCase passed

PR #403, opened 2020-04-17
Prop sensors impl

athena255:
# Prop Sensors Implementation

### Summary of changes
- sensor implementation previously in the prop driver case PR
- prop sensors test
- prop sensors hardware build environment
- conditionally compiled tests in prop controller unit test for manually checking pressure and temperature conversion math
- ability to set fake values for the analogRead() result of all sensor pins. This should help functional tests

### Testing
- basic tests in prop_sensors that print the value of the sensors
- tests in prop controller unit tests file to test mathematical correctness

### Constants
- constants relating to sensors such as for regression calculation

### Telemetry
None

### Documentation Evidence
- All documentation from the sensor implementation that was in prop driver case PR is now here


src/fsw/FCCode/Drivers/PropulsionSystem.cpp
```diff
@@ -59,6 +66,11 @@ IntervalTimer _Tank2::thrust_valve_loop_timer = IntervalTimer();
 bool _PropulsionSystem::setup() {
     Tank1.setup();
     Tank2.setup();
+#ifndef DESKTOP
+    // Set 10-bit resolution since the regression for the pressure sensor
+    // calculations was computed using a 10-bit Teensy
```
tanishqaggarwal, line 71: Is this comment correct? I haven't used analogReadAveraging before but from what it seems online, this function is measuring the pin, not setting a resolution.

tanishqaggarwal: Did you mean to use the function `analogReadResolution`?

src/fsw/FCCode/Drivers/PropulsionSystem.hpp
```diff
@@ -197,7 +206,7 @@ class Tank {
      */
     bool is_valve_open(size_t valve_idx) const;
 
-protected:
+//protected: // yes, this is bad
```
tanishqaggarwal, line 209: Let's just remove the line, then.

src/fsw/FCCode/Drivers/PropulsionSystem.hpp
```diff
@@ -216,6 +225,23 @@ class Tank {
     // true if the valve is opened
     bool is_valve_opened[4];
 
+    // These constants are from the nonlinear regression for computing tank temperature
+    // T = A * ln(R)^EXP + B
+    // T is temperature, R is resistance
+    TRACKED_CONSTANT(double, temp_a, -35126.92396);
+    TRACKED_CONSTANT(double, temp_exp, 0.005);
+    TRACKED_CONSTANT(double, temp_b, 35493.23411);
```
tanishqaggarwal, line 233: Can these be TRACKED_CONSTANT_SC()? So that the variables become `static constexpr`

src/fsw/FCCode/radio_err_t.enum
```diff
@@ -1 +0,0 @@
-
```
tanishqaggarwal, line 1: Let's just delete this file

test/test_fsw_prop_controller/test_prop_controller.cpp
```diff
@@ -313,6 +314,37 @@ void test_firing_to_idle()
     tf.check_state(prop_state_t::idle);
 }
 
+// These two tests are manually checked, so the for loop is conditionally compiled
+// in order to not lag everything
+
+void test_temp_sensor_logic()
+{
+    Tank1.fake_tank1_temp_sensor_read = 1;
```
tanishqaggarwal, line 322: This line of code doesn't seem to do anything

test/test_prop/test_prop.cpp
```diff
@@ -321,10 +321,15 @@ void setup() {
     RUN_TEST(test_is_functional);
     scheduling_tests();
     RUN_TEST(test_reset);
+    TEST_ASSERT_TRUE(PropulsionSystem.is_functional());
```
tanishqaggarwal, line 324: Are these multiple is_functional checks in here so that we can verify the DCDC is on after every test, e.g. is not shutting down due to low power? If so, could you add a comment about this

src/fsw/FCCode/Drivers/PropulsionSystem.hpp
```diff
@@ -76,8 +76,17 @@ class Tank;
  *  - tank2 firing schedule
  * 
  * Only tank2 has a schedule since only tank2 uses the IntervalTimer to fire.
+ *
+ * ------
+ *
+ * To mock the sensors values, assigned
+ * - fake_tank1_temp_sensor_read to the desire value of Tank1.analogRead(temp_sensor_pin)
+ * - fake_tank2_temp_sensor_read to the desire value of Tank2.analogRead(temp_sensor_pin)
+ * - fake_tank2_pressure_low_read to the desire value of Tank2.analogRead(pressure_sensor_low_pin)
```
tanishqaggarwal, line 85: "desired"

src/fsw/FCCode/Drivers/PropulsionSystem.cpp
athena255: Yep, definitely meant analogReadResolution. Probably tab fail

src/fsw/FCCode/Drivers/PropulsionSystem.hpp
athena255: It is removed, but it is technically bad.

test/test_prop/test_prop.cpp
athena255: Yes

PR #421, opened 2020-04-22
Autocompiler for ptest

tanishqaggarwal:
### Summary of changes
- Autocompiles missing assets when running PTest.
- Some random updates to constants

### Testing
Verified that missing binaries are compiled when the PIO folder is removed. Verified that the binaries are not recompiled if they already exist.

### Constants
Some weird changes that have been left over on master.

### Telemetry
No changes.

### Documentation Evidence
No documentation required; this is a very silent feature.


tanishqaggarwal:
If the compilation fails it prints the error to the console (since errors aren't getting piped to /dev/null). But sure we can switch to subprocess. Could you merge it once the build passes?

kkrol27:
> If the compilation fails it prints the error to the console (since errors aren't getting piped to /dev/null). But sure we can switch to subprocess. Could you merge it once the build passes?

Will it exit the case if the build fails or no?

tanishqaggarwal:
It seems to exit fine if the build fails.

Issue #346, opened 2020-03-30
ADCSC Checkout Case Ptest Upgrades


shihaocao:
- Add further sensor checkouts
- Add wheel and ssa checkouts
- Update IMU tests after MAG breakout PR

shihaocao:
Going to leave this open because there's still more combinations to test, namely how magnetometers react to MTRS and wheels, also calibration?

PR #429, opened 2020-04-26
ADCSCheckoutCase - Wheels and SSA

shihaocao:
# ADCSCheckoutCase - Wheels and SSA

As of now, its just a diagnostic case.
Addresses part of #346 

### Summary of changes
- Runs batches of speed and torque tests on wheels
- Runs a basic magnitude and differential test on SSA sensors.

### Testing
Tested on EDUSat, performs as reasonably expected. Logs in HITL-Logs

Negative wheel speed commands do not work right now. Will be fixed in the future

### Constants
N/A

### Telemetry
N/A

### Documentation Evidence
Added more to the ADCSCheckoutCase blurb


constants
```diff
@@ -20,16 +20,18 @@ src/adcs/constants.hpp:144: "min_moment" = "-max_moment"
...
-src/adcs/constants.hpp:165: "max_speed_command" = "680.678f"
-src/adcs/constants.hpp:167: "min_speed_command" = "-max_speed_command"
-src/adcs/constants.hpp:174: "min_voltage_rd" = "0.0f"
-src/adcs/constants.hpp:176: "max_voltage_rd" = "3.3f"
-src/adcs/constants.hpp:178: "min_voltage_thresh" = "min_voltage_rd"
-src/adcs/constants.hpp:180: "max_voltage_thresh" = "max_voltage_rd"
-src/adcs/constants.hpp:182: "num_sun_sensors" = "20"
-src/adcs/constants.hpp:189: "max_devices" = "32"
+src/adcs/constants.hpp:160: "max_ramp_rd" = "max_torque"
+src/adcs/constants.hpp:162: "min_ramp_rd" = "0"
```
tanishqaggarwal, line 24: We need to create an issue ticket to modify ADCS software to make this "-max_torque", right? Correct me if I'm misremembering

ptest/cases/adcs_checkout_case.py
```diff
@@ -247,8 +376,15 @@ def run_case_singlesat(self):
         # Run checks on GYR if GYR is up.
         if self.rs("adcs_monitor.havt_device0"):
             self.gyr_checkout()
+        
+        # Run wheel checks if the RWAPOT is up.
+        rwa_pot_num = self.havt_devices.get_by_name("RWA_POT")
+        if self.rs(f"adcs_monitor.havt_device{rwa_pot_num}"):
+            self.wheel_checkout()
```
tanishqaggarwal, line 383: Should we throw some kind of failure if the pot isn't available?

ptest/cases/adcs_checkout_case.py
```diff
@@ -247,8 +376,15 @@ def run_case_singlesat(self):
...
 
+        self.ssa_checkout()
 
+        self.print_rs("gomspace.vbatt")
         # TODO FURTHER CHECKOUTS
```
tanishqaggarwal, line 388: What else is there left?

ptest/cases/adcs_checkout_case.py
```diff
@@ -150,13 +150,13 @@ def mag_checkout(self, mag_num):
         # earth's mag field is between 25 to 65 microteslas - Wikipedia
         self.logger.put(f"MAG{mag_num} readings: ")
         for i in range(10):
-            mag = mag_of(list_of_mag_rds[i])
+            mag = self.self.mag_of(list_of_mag_rds[i])
```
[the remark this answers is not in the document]
shihaocao: Oops, thats fixed now

constants
shihaocao: Actually, the readings are only from 0 to max_torque, but the commands can range from -max_torque to max_torque. So this PR is correct as is.

ptest/cases/adcs_checkout_case.py
shihaocao: As is, the ADCSCheckoutCase is intended to be multi-purposed for all kinds of test beds. The  case tells the user what test bed it thinks its running on earlier with the "HAVT TABLE MATCHES WITH" section. So device failures are cataloged and alerted there.

ptest/cases/adcs_checkout_case.py
shihaocao: Magnetometer Calibration, and how the magnetometers react to MTRs and Wheels, if that's still on the table?

constants
tanishqaggarwal: Right, but isn't it a bug that the readings only range from 0 to max_torque? Is there an issue ticket tracking this work?

kkrol27: This is the desired behavior. It's only supposed to be magnitude because that's how the hardware is actually commanded.

ptest/cases/adcs_checkout_case.py
```diff
@@ -213,11 +216,140 @@ def gyr_checkout(self):
...
+        list_of_voltages = []
+        for i in range(10):
+            voltages = []
+            for i in range(0, 20):
+                reading = self.print_rs(f"adcs_monitor.ssa_voltage{i}")
+                voltages += [reading]
+            list_of_voltages += [voltages]
+        # check readings changed over time
+        self.soft_assert(self.sum_of_differentials(list_of_voltages) > 0,
+            "SSA voltage readings did not vary across readings.")
```
kkrol27, line 351: Doesn't this soft assert need to be done on each sensor individually?

kkrol27:
Just have the one quick question about looking at the SSA voltage derivatives. Otherwise looks great.

ptest/cases/adcs_checkout_case.py
shihaocao: Yes, but I thought it'd be a bit cumbersome in this case for every single sensor (for the purposes of this checkout case). I talked with stewart about this, I think I'll actually split this into a separate manual diagnostic case later for flight satellite verification. To address that I'll add a quick ticket for that (to be done in the future).

PR #430, opened 2020-04-26
Adding the gomspace battery telemetry points and giving them initial values

knk38:
# Adding GomSpace telemetry points and initializing them

Fixes #. (Optional; sometimes PRs don't have an associated ticket)
#385 
### Summary of changes
- created a new object definition for the battery (battery.json) 
- created a plugin to take in battery telemetry type values
-created battery-telemetry.js to create telemetry points based off the object definition and sent it to openMCT
-as an aside fixed console output to make it less cluttery and more helpful
-moved MCT to the roof of FlightSoftware

-NOTE: spacecraft.js is still in this PR and can be run using "npm start" the battery is run by "npm start test"

### Testing
-Tested the points by running MCT with all of the files and checking if both object type telemetry and primitive type telemetry was displayed.
-also checked JS console in browser to ensure that there were not any underlying errors, (the only error that shows up is expected and occurs if the flask server is not online)

### Constants
-no changes to constants file

### Telemetry
-no changes to the telemetry file, however changes to battery.json were needed to allow subobjects to be called without the "gomspace." prefix
-some slight changes in naming to make points easier to access

### Documentation Evidence
-added substantial in line documentation to battery-telemetry.js
-also added documentation to new files that were added, describing their use (this is located at the top of the file)


tanishqaggarwal:
Fixes #385

MCT/server-files/battery-telemetry.js
```diff
@@ -0,0 +1,100 @@
+/*
+ battery-telemetry.js simulates the gomspace generating telemetry.
+*/
+const variables = require('./state-variables.js')
+
+const axios = require('axios');
+function Battery() {
```
fatimayousuf, line 7: functions in the Battery class, like updateState and generateTelemetry, look like they could be generalized to any system creating statefields, besides the gomspace battery. It seems to me that, when you add telemetry files for devices like Piksi, Quake, etc., there'll be a lot of copy+paste and overlap for these methods. Maybe you could put those methods in a superclass and then make Battery() a subclass?

fatimayousuf: It's definitely not imperative that this happen for this specific pr. I'm just wondering how all the pieces will fit together as you add more devices and telemetry generators

MCT/public/telemetry-plugin.js
```diff
@@ -0,0 +1,147 @@
...
+function getBattery() {
+    return http.get('/battery.json')
+        .then(function (result) {
+            return result.data;
+        });
+}
+
+var objectProvider1 = {
+    get: function (identifier) {
+        return getBattery().then(function (battery) {
```
tanishqaggarwal, line 84: Could you explain the logic in this function?

MCT/server-files/battery-telemetry.js
tanishqaggarwal: I agree. We need to make a much more streamlined architecture for the code here. For this PR let's merge this in so that we have the pieces for battery telemetry. I'll try to think of some good solutions for restructuring the code.

MCT/public/telemetry-plugin.js
```diff
@@ -0,0 +1,147 @@
+//This file contains all the plugins necessary for openMCT to handle our telemetry values
+
+function getDictionary() {
+    return http.get('/dictionary.json')
+        .then(function (result) {
+            return result.data;
+        });
+}
+
+var objectProvider = {
```
tanishqaggarwal, line 10: I don't think splitting between "spacecraft" and "everything else" is the best strategy for writing these functions. It makes adding new domain objects a very cumbersome project.

MCT/server-files/battery-telemetry.js
knk38: Yeah for sure I think we could make a telemetry object superclass that contains all of the functions required to send telemetry (the last 4 functions) and then just have each subclass modify the generateTelemetry and this.state field (eventually updateState I think could be made general as well)

MCT/public/telemetry-plugin.js
knk38: yeah i agree, i tried to combine them, but the problem is with naming and how they are designed now they have to use different objectProviders unless I redesign how the telemetry is installed to openMCT

MCT/public/telemetry-plugin.js
knk38: so the two BatteryPlugin and SpacecraftPlugin functions call their object providers which read in each field of the JSON and add them to the openMCT dictionary, then the composition provider links the root folder to every telemetry point that comes in and has the sat. prefix or the gomspace. prefix

MCT/public/index.html
```diff
@@ -20,10 +19,17 @@
         openmct.time.clock('local', {start: -15 * 60 * 1000, end: 0});
         openmct.time.timeSystem('utc');
         openmct.install(openmct.plugins.Espresso());
-
+        openmct.install(BatteryPlugin());
         openmct.install(DictionaryPlugin());
         openmct.install(HistoricalTelemetryPlugin());
         openmct.install(RealtimeTelemetryPlugin());
+        openmct.install(openmct.plugins.URLIndicator({
+            url: 'http://localhost:5000',
```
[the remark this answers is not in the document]
knk38: yeah i agree, but this url indicator has turned in to too much of a hassle, ill take it out for now and add it if I can figure it out later

tanishqaggarwal:
@knk38 and I agreed to close this until things are a little more stable.

PR #431, opened 2020-04-26
Refactor FSW enum to be dict-accessible

tanishqaggarwal:
# Summary of changes
- Remove `get_by_name` and `get_by_num` from PTest FSW enums, instead making the values dict-accessible.

### Testing
Existing tests are passing, so change is ok.

### Constants
No constants changed.

### Telemetry
No telemetry changed.

### Documentation Evidence
This is an implementation-only change; PTest testcase writing patterns serve as sufficient documentation.


ptest/runsim.py
```diff
@@ -157,7 +158,7 @@ def set_up_sim(self):
         if self.single_sat_sim:
             self.sim = SingleSatSimulation(self.is_interactive, self.devices, self.random_seed, testcase(self.simulation_run_dir))
         else:
-            self.sim = Simulation(self.is_interactive, self.devices, self.random_seed)
+            self.sim = Simulation(self.is_interactive, self.devices, self.random_seed, testcase(self.simulation_run_dir))
```
shihaocao, line 161: whats this for?

tanishqaggarwal: this is what constructs the testcase and allows it to run

ptest/cases/base.py
```diff
@@ -20,11 +20,13 @@ def __init__(self, arr):
...
-
-    def get_by_num(self, num):
-        return self._indexed_by_num[num]
+    def __getitem__(self, item):
+        if type(item) is str:
+            return self._indexed_by_name[item]
+        elif type(item) is int:
+            return self._indexed_by_num[item]
+        else:
+            raise AttributeError(f"Cannot access FSWEnum with key: {item}")
```
kkrol27, line 29: Just to be clear, you could also do something like `enum[1]` and it'd return the string representation of the enum value?

tanishqaggarwal: yep

PR #432, opened 2020-04-26
Prevent mission states that are not "docking" or "docked" from being saved to EEPROM

tanishqaggarwal:
### Summary of changes
- See title. The goal is to always force the satellite into a clean boot state, unless it's in a state where actuation of wheels or anything might be dangerous.

### Testing
Unit test passes on HITL and HOOTL.

### Constants and Telemetry
No changes.

### Documentation Evidence
Deferred to #433


ptest/cases/quake_fault_handler.py
```diff
@@ -0,0 +1,110 @@
+# Runs mission from startup state to standby state.
+from .base import SingleSatOnlyCase, TestCaseFailure
+from .mission import MissionStateMachineUtilities
+
+class QuakeFaultHandler_Fast(SingleSatOnlyCase, MissionStateMachineUtilities):
```
fatimayousuf, line 5: This file doesn't seem like it's related to this pr.

tanishqaggarwal: Oh shoot, I clicked merge too fast. It'll get overwritten soon anyway

Issue #295, opened 2020-03-12
Complete stub implementation of PropOverpressuredFaultHandler


tanishqaggarwal:
Or do whatever makes most sense to integrate overpressure handling.

Depends on #262.

PR #434, opened 2020-04-26
Main fault handler end-to-end testing

### LINKED ISSUE #296: End-to-end MainFaultHandler Unit Test
Write a unit test that uses real fault handlers instead of mocks.

tanishqaggarwal:
Fixes #296 

### Summary of changes
- Add C++-based end-to-end testing of MainFaultHandler.
- Move Piksi and Quake fault handler testing into same folder as MainFaultHandler.
- Added an additional build flag called `SPEEDUP` that artificially shortens certain timeouts in flight software. The flag is enabled by default in HOOTL unless one of the environments `fsw_native_leader_realtime` or `fsw_native_follower_realtime` is compiled.

A PR (or set of PRs) for MainFaultHandler end-to-end testing in PTest is coming soon. After #295 is completed, we'll create a PR for that too.

### Testing
The main fault handler tests that one or two simple faults in combination cause the correct response.

The Quake and Piksi fault handlers' combination-faulting is deferred to PTest testing since at a fault-handler level, the interface is identical to the simple faults. It makes more sense to do an end-to-end test in PTest for these fault handlers since  

### Constants
Added a SPEEDUP-ed version of 24 hours that only takes 1 minute in real time.

### Telemetry
Added `qfh.state` which enables tracking of the Quake Fault Handler's state from ptest.

### Documentation Evidence
This is a testing-only PR, so no documentation is necessary besides the internally contained documentation.


test/test_fsw_fault_handler/test_fixture_main_fh.cpp
```diff
@@ -0,0 +1,70 @@
+#include "test_fixture_main_fh.hpp"
+#include "FaultHandlerMachineMock.hpp"
+
+TestFixtureMainFHBase::TestFixtureMainFHBase() {
+    cc = 0;
+
+    // Prwepare inputs for main fault handler
```
fatimayousuf, line 7: you have a spelling error hehe 😜

fatimayousuf, line 7: You have a spelling error 😜

test/test_fsw_fault_handler/test_fixture_main_fh.hpp
```diff
@@ -0,0 +1,102 @@
...
+    std::shared_ptr<Fault> prop_overpressure_fault_fp;
+
+    WritableStateField<bool>* fault_handler_enabled_fp = nullptr;
+
+    /**
+     * @brief Construct a new Test Fixture.
+     * 
+     * @param mock If true, replaces sub-fault handlers with mocks.
+     */
+    TestFixtureMainFHBase();
```
fatimayousuf, line 42: There is no mock parameter

src/fsw/FCCode/MainFaultHandler.hpp
```diff
@@ -6,7 +6,9 @@
 
 class MainFaultHandler : public FaultHandlerMachine {
   #ifdef UNIT_TEST
-    friend class TestFixtureMainFH;
+    friend class TestFixtureMainFHBase;
+    friend class TestFixtureMainFHMocked;
+    friend class TestFixtureMainFHEndToEnd;
```
fatimayousuf, line 11: could you maybe write a sentence in the header files on what each of these classes do?

test/test_fsw_fault_handler/test_fixture_main_fh.cpp
```diff
@@ -0,0 +1,70 @@
...
+    fault_handler_enabled_fp->set(state);
+}
+
+TestFixtureMainFHMocked::TestFixtureMainFHMocked() : TestFixtureMainFHBase() {
+    num_fault_handler_machines = fault_handler->fault_handler_machines.size();
+
+    // Replace all fault handler submachines with mocks
+    for(size_t i = 0; i < num_fault_handler_machines; i++) {
+        fault_handler->fault_handler_machines[i] = std::make_unique<FaultHandlerMachineMock>(registry);
+    }
```
fatimayousuf, line 41: why is this necessary? If you're preparing all the input fields, then shouldn't all the submachine work fine?

tanishqaggarwal: The goal of this was to make the submachines have exactly the same commanding interface, so that I could exhaustively test all combinations of faults using for-loops in test_main_fault_handler.cpp.

The test where I don't substitute the submachines for mocks is test_main_fh_end_to_end, where I selectively tested certain combinations of real faults.

src/fsw/FCCode/MainFaultHandler.hpp
tanishqaggarwal: Added to test_fixture_main_fh.hpp

test/test_fsw_fault_handler/test_main_fh_end_to_end.cpp
```diff
@@ -0,0 +1,237 @@
...
+
+        tf.cc += 5;
+        tf.prop_failed_pressurize_fault_fp->signal(); tf.step();
+        tf.prop_failed_pressurize_fault_fp->signal();
+        response = tf.step();
+        TEST_ASSERT_EQUAL(fault_response_t::safehold, response);
+    }
+
+    /**
+     * Suppose the pressurization fault is currently signaled (so that the safehold
```
kkrol27, line 166:
```suggestion
     * Suppose the pressurization fault is currently signaled (so that the standby 
```

src/fsw/FCCode/MissionManager.cpp
```diff
@@ -275,7 +278,7 @@ void MissionManager::dispatch_docking() {
         set(sat_designation_t::undecided);
     }
     else if(too_long_in_docking() && !docked_fp->get()) {
```
shihaocao, line 280: Unrelated to changes, but isn't !docked_fp->get() always true by the time we get to line 280? Should probably simplify that

test/test_fsw_fault_handler/test_faults.cpp
```diff
@@ -9,9 +9,9 @@ void test_faults() {
 
     // TODO
     // RUN_TEST(test_prop_overpressure_fault_handler);
-    // RUN_TEST(test_quake_fault_handler);
+    RUN_TEST(test_quake_fault_handler);
     // RUN_TEST(test_piksi_fault_handler);
```
shihaocao, line 13: shouldn't this be uncommented?

test/test_fsw_fault_handler/test_fixture_main_fh.hpp
shihaocao: I think there is still no mock parameter, i am still confused

test/test_fsw_fault_handler/test_main_fh_end_to_end.cpp
```diff
@@ -0,0 +1,237 @@
...
+    fault_response_t response = tf.step();
+    TEST_ASSERT_EQUAL(fault_response_t::none, response);
+}
+
+void test_single_simple_faults(std::vector<std::shared_ptr<Fault>>& faults,
+    fault_response_t expected_response, TestFixtureMainFHEndToEnd& tf)
+{
+    fault_response_t response;
+    for(auto& fault_ptr : faults) {
+        fault_ptr->signal(); tf.step();
```
shihaocao, line 16: I think it doesn't hurt to assert that tf.step() = none here, as well as all occurances of tf.step() where the fault handling response should be none.

test/test_fsw_fault_handler/test_main_fh_end_to_end.cpp
```diff
@@ -0,0 +1,237 @@
...
+    TEST_ASSERT_EQUAL(fault_response_t::standby, response);
+
+    // Continue stepping through the Quake fault state machine until we reach safehold.
+    tf.cc = 2 * PAN::one_day_ccno + 1;
+    tf.step(); // Now in powercycle_1
+    tf.cc = 7 * PAN::one_day_ccno / 3 + 1;
+    tf.step(); // Now in powercycle_2
+    tf.cc = 8 * PAN::one_day_ccno / 3 + 1;
+    tf.step(); // Now in powercycle_3
+    tf.cc = 3 * PAN::one_day_ccno / 3 + 1;
```
shihaocao, line 91: I think all of these tf.steps() should assert that the fault_response is `standby`

test/test_fsw_fault_handler/test_faults.cpp
tanishqaggarwal: Yep good catch

src/fsw/FCCode/MissionManager.cpp
tanishqaggarwal: docked_fp() tracks if the physical docking switch is pressed. The point of the if...else if structure is that there may be cases where we've docked but haven't gotten confirmation from the docking switch. In that case, we rely on the ground to set the state to "docked"--and if we don't get comms from the ground within 24 hours about this, then we'll revert to standby.

src/fsw/FCCode/SimpleFaultHandler.cpp
```diff
@@ -9,10 +9,12 @@ SimpleFaultHandler::SimpleFaultHandler(StateFieldRegistry& r, Fault* f,
         recommended_state(rs)
 {
     assert(rs == mission_state_t::safehold || rs == mission_state_t::standby);
+    mission_state_fp = FIND_WRITABLE_FIELD(unsigned char, pan.state);
```
athena255, line 12: How is this `FIND_WRITABLE_FIELD` different than the other `find_writable_field<unsigned char>` ?

test/test_fsw_fault_handler/test_main_fh_end_to_end.cpp
```diff
@@ -0,0 +1,237 @@
...
+     * response (safehold) is the one that dominates. Then, these two tests check 
+     * that if one of the faults goes away, the one remaining is the one whose fault
+     * response is applied.
+     **/
+    {
+        TestFixtureMainFHEndToEnd tf;
+
+        // If both faults occur simultaneously, the stronger
+        // response is recommended.
+        tf.prop_failed_pressurize_fault_fp->signal();
```
athena255, line 108: Due to persistence, prop_failed_pressurize_fault will not be faulted upon the first signal.

test/test_fsw_fault_handler/test_main_fh_end_to_end.cpp
```diff
@@ -0,0 +1,237 @@
...
+        // cycles later, it should not affect the fault response recommendation.
+        // If the pressurization fault then disappears, the fault response recommendation
+        // is still safehold.
+        TestFixtureMainFHEndToEnd tf;
+        TEST_ASSERT_EQUAL(fault_response_t::none, tf.step());
+        tf.adcs_wheel1_adc_fault_fp->signal();
+        TEST_ASSERT_EQUAL(fault_response_t::safehold, tf.step());
+
+        tf.cc += 5;
+        tf.prop_failed_pressurize_fault_fp->signal();
```
athena255, line 153: This fault will not affect the response anyway since the pressurize_failed fault has a persistence of 10

test/test_fsw_fault_handler/test_main_fh_end_to_end.cpp
```diff
@@ -0,0 +1,237 @@
+#include "test_fault_handlers.hpp"
```
athena255, line 1: Suggestion: The persistence of faults is not set in stone. We may later decide to adjust the persistence values. When we do, some test cases here will fail because they may expect a fault to be faulted, but due to persistence, the fault has not been signaled for enough control cycles to be faulted. Perhaps use the persistence to determine how many cycles to step in order to get the fault state.

test/test_fsw_fault_handler/test_quake_fault_handler.cpp
```diff
@@ -138,7 +149,7 @@ void test_qfh_unfaulted() {
     // If the radio is enabled, ensure that there's a transition
     // to the forced standby state after 24 hours.
     {
-        TestFixture tf{qfh_state_t::unfaulted};
+        TestFixtureQFH tf{qfh_state_t::unfaulted};
```
athena255, line 152: The last_comms state field is set to the current control cycle count when QuakeManager is created. This is not updated until QuakeManager receives comms. If QuakeManager is in disabled state for an extended amount of time (like 24 hours), and is then turned on, QuakeFaultManager might believe that the radio has been unsuccessful in attempting comms for over 24 hours when actually, the radio has been disabled for 24 hours.

src/fsw/FCCode/MissionManager.cpp
```diff
@@ -263,11 +262,15 @@ void MissionManager::dispatch_leader_close_approach() {
     }
 }
 
+static bool have_set_docking_entry_ccno = false;
```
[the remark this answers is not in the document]
tanishqaggarwal: Ah yeah I should add a comment here. This flag checks if we've set the state field called `docking_entry_ccno`, which indicates the control cycle # at which we entered the docking state. This state field is used by `PiksiFaultHandler` to know if we've been lacking CDGPS for too long.

src/fsw/FCCode/SimpleFaultHandler.cpp
tanishqaggarwal: Kyle made this macro to prevent having a bunch of `__FILE__` and `__LINE__` statements everywhere. It's no different; it's just a wrapper around `find_writable_field<unsigned char>`.

test/test_fsw_fault_handler/test_main_fh_end_to_end.cpp
tanishqaggarwal: Ah thanks. I'll make a note of this so that we know why fault handler tests are failing if they fail in the prop faults PR

test/test_fsw_fault_handler/test_main_fh_end_to_end.cpp
tanishqaggarwal: Yep good idea. I'll implement that.

test/test_fsw_fault_handler/test_quake_fault_handler.cpp
tanishqaggarwal: I think there's a check in the unfaulted test (see QuakeFaultHandler.cpp) that ignores the last_comms if the radio has been disabled for a while.

In practice we would never disable the radio for 24 hours; it's only ever disabled once, in startup, and then enabled permanently (there aren't even any ground commands or autonomous transitions that could cause it to be disabled.) And in startup, it's disabled for a maximum of 30 minutes.

src/fsw/FCCode/QuakeFaultHandler.cpp
```diff
@@ -4,41 +4,49 @@
...
+
+    cur_state.set(static_cast<unsigned char>(qfh_state_t::unfaulted));
+    add_writable_field(cur_state);
 }
 
 fault_response_t QuakeFaultHandler::execute() {
-    switch(cur_state) {
+    qfh_state_t state = static_cast<qfh_state_t>(cur_state.get());
+
+    switch(state) {
```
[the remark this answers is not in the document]
tanishqaggarwal: Good point, this is redundant. I'll remove

test/test_fsw_fault_handler/test_main_fh_end_to_end.cpp
tanishqaggarwal: Actually, this is not going to affect the unit test because the faults are mocked. If the persistence is changed in the unit test, it'll be easy to trace back the failure to a corresponding change in the mocked fault.

src/fsw/FCCode/MissionManager.cpp
shihaocao: Wait uh, but I think I meant like, the only time you can get to the `else if` branch is like if the earlier `if(docked_fp->get())` is false (since state_fields aren't updated until uplink_consumer is called again)? Or am i actually high

test/test_fsw_fault_handler/test_main_fault_handler.cpp
```diff
@@ -195,7 +94,7 @@ void test_main_fh_safehold_fault() {
  * responses works.
  */
 void test_main_fh_toggle_handling() {
-    TestFixtureMainFH tf;
+    TestFixtureMainFHMocked tf;
     assert(tf.num_fault_handler_machines == 10);
 
     // This is a random combination that definitely causes a fault
```
shihaocao, line 100: Unrelated to changes, but below there's also another safehold_combo with only 9 out of 10 elements

shihaocao: (look below 201, wouldn't let me comment any further below)

src/fsw/FCCode/MissionManager.cpp
tanishqaggarwal: oh yeah true. thanks

test/test_fsw_fault_handler/test_main_fault_handler.cpp
tanishqaggarwal: Fixed, thanks

test/test_fsw_fault_handler/test_fixture_main_fh.hpp
tanishqaggarwal: I got rid of the comment, Im pretty sure

tanishqaggarwal: Oh huh the change disappeared, weird

Issue #426, opened 2020-04-24
Hardcode elasticsearch port and server information


tanishqaggarwal:
Alternatively, we could hardcode the server and port number for elasticsearch into this function, since it's unlikely to change in local testing. This would still allow us to delete the elasticsearch config information in #413.

_Originally posted by @tanishqaggarwal in https://github.com/pathfinder-for-autonomous-navigation/FlightSoftware/pull/390/threads/MDIzOlB1bGxSZXF1ZXN0UmV2aWV3VGhyZWFkMjU1ODg2NDk4OnYy/unresolve_

PR #444, opened 2020-04-30
Fix location of telem, dbtelem parsetelem functions

### LINKED ISSUE #425: Consolidate uplink producer and downlink producer binaries into the `PTest` class
So that we don't have a weird separation of duties/crossing across state sessions, radio sessions, and statecmdprompt.

tanishqaggarwal:
Fixes #425 and #426.

### Summary of changes
- Move location of telemetry parsing functions to StateSession, since downlink is obtained from there and should be processed there.
- Hardcode elasticsearch server and port, as well as binary filepaths--they won't change.
- Added `StateSessionOnly` decorator to StateCmdPrompt to prevent functions from being called on the wrong kind of `cmded_device`.

### Testing
I booted up a ptest case with both a StateSession and RadioSession. The functions worked inside StateSession as expected.

### Constants
No new constants.

### Telemetry
No new telemetry.

### Documentation Evidence
This does not change the outer-facing interface of PTest. It's purely an implementation change.


ptest/state_session.py
```diff
@@ -418,14 +435,44 @@ def uplink(self, fields, vals, timeout=None):
...
+        self.dp_console.write((newest_telem_file+"\n").encode())
+        telem_json_data = json.loads(self.dp_console.readline().rstrip())
+        if telem_json_data is not None:
+                telem_json_data = telem_json_data["data"]
+        return telem_json_data
+
+    def dbtelem(self):
+        jsonObj = self.parsetelem()
+        if not isinstance(jsonObj, dict):
+            print("Telemetry in obtained downlink files is not of JSON format.")
```
afoarce, line 463: I don't think this print statement is necessarily correct, you could just be getting the string from parsetelem that says "No telemetry to parse" or have a None object, which is different from having incorrect JSON.

tanishqaggarwal: I'll change it to "Error parsing telemetry."

Issue #287, opened 2020-03-12
Attitude Controller


shihaocao:
For the sake of clarity, we'll merge the `AttitudeComputer` and `ADCSCommander` control tasks into a single control task named `AttitudeController`. In this process, we'll also integrate the controllers that have been incorporated into PSim (the detumbler and pointer controllers to be specific).

The main reasons for merging the control tasks tasks is to centralize all of the control logic for a given ADCS mode. Currently, the pointing objectives are determined in the `ADCSComputer` while the actual pointing happens in the `ADCSCommander`. This by itself makes sense except for the fact that not all ADCS modes require pointing objectives so sometimes the `ADCSComputer` is useless. Furthermore, many different ADCS modes use the same controller. This futhre makes the decentralization of attitude control difficult to follow.

In addition to merging the control tasks and integrating the controllers, we'll also need to make a small update to the current `AttitudeEstimator`. The finite differencing currently used to determine our angular rate is incredibly unreliable and we'll be moving to leveraging our gyroscope measurement in the future.

kkrol27:
@shihaocao could you expand on which GNC functions we are referencing here?

shihaocao:
I've actually got no idea what the names of them are but,
They are the functions that convert the desired vectors to actual MTR and Wheel commands.

kkrol27:
Ahhh okay so stuff on top of the controller itself. We can talk about this this weekend and find the companion PSim tickets.

kkrol27:
Just assigning myself too so I don't forget.

kkrol27:
@shihaocao After talking with @tanishqaggarwal and @stewartaslan I'm just going to take this over for next week to take some stuff off your plate.

kkrol27:
Per @nhz2 request, we'll also need to update documentation here in accordance with implementation: https://pan-software.readthedocs.io/en/latest/flight_software/components.html

kkrol27:
The current pointing strategy defined in FSW needs to be reworked have the long edge of the satellite along the orbit normal and point the antenna phase outward at some angle `phi` specified in the hill frame.

PR #446, opened 2020-05-01
Update the Attitude Estimator to the UKF Interface

### LINKED ISSUE #348: Attitude Estimator
The updated attitude estimator interface has been merged in to the most up to date version of PSim and can now be integrated.

The tasks required are as follows:

* Add an `orbit.time` state field to the `FieldCreatorTask`. This will be a `double` holding time in seconds since the PAN epoch (closes #350).
* Update the `AttitudeEstimator` interface to take the gyroscope reading as an input and add a statefield to output our estimated angular rate (i.e. measured gyro minus the estimated bias).
* For now, only initialize/reset the filter when we a valid sun vector. position, time, and magnetometer reading (i.e. use the triad initialization function).
* We should have some threshold on the estimated state covariance which causes us to reset the filter - e.g. `if (lin::fro(estimate.P) > thresh)` - where thresh should be some tracked constant.
* Implement #445 as well.

### LINKED ISSUE #350: GPS/GNC Time Implementation
We need to add a state field to provide time in seconds/nanoseconds since the "PAN epoch". This will be fed to many of the GNC algorithms and environmental models.

### LINKED ISSUE #445: Rename fields inside AttitudeEstimator
Not applicable to this review, but we should adjust this to `w_body_f` and `w_wheels_f`.

_Originally posted by @kkrol27 in https://github.com/pathfinder-for-autonomous-navigation/FlightSoftware/pull/435_

kkrol27:
# Update the Attitude Estimator to the UKF Interface

Closes #348.
Closes #350.
Closes #445.

### Summary of changes
- Added the `orbit.time` state field in the `FieldCreatorTask` which holds the time in seconds since the PAN epoch as a double.
- Update the `AttitudeEstimator` interface:
  - Added the `attitude_estimator.mag_flag` state field to chose a default magnetometer sensor.
  - Added the `attitude_estimator.w_body` state field which is the current angular rate estimate.
  - Removed the `attitude_estimator.h_body` state field.
  - Added the `attitude_estimator.fro_P` state field to help monitor uncertainty.
- Updated mission manager to consume the angular rate state field as opposed to angular momentum - the detumble threshold doesn't need to take into account wheel rates because they won't be spinning.

Going forward we have tickets #448, #287, and then full attitude estimation+control in the loop PTest cases.

### Testing
See updated unit tests for the attitude estimator and mission manager. Note that currently we cannot write an attitude estimator initialization test because the triad based reset function is not yet implemented in PSim. Once this is complete, we can expand testing (#448).

### Constants
NA

### Telemetry
Removed the angular momentum field and added the estimated angular rate and state covariance frobenius norm field.

### Documentation Evidence
See inline documentation.


kkrol27:
@stewartaslan How about the serializer bounds on the forbenius norm of the state covariance?
@tanishqaggarwal Can a state field hold a value larger than it's serializer bounds? And then when it gets serialized is the value just clamped?

kkrol27:
@srutiv Also, do you remember which magnetometer was more reliable from your analysis?

tanishqaggarwal:
> @tanishqaggarwal Can a state field hold a value larger than it's serializer bounds? And then when it gets serialized is the value just clamped?

A state field can hold a larger value, but it'll get clamped when being downlinked/uplinked.

kkrol27:
> @stewartaslan How about the serializer bounds on the forbenius norm of the state covariance?

Fixed; setting the upper bound on the serializer to be on the order of the frobenius norm of the initial state covariance.

PR #452, opened 2020-05-03
Ptest reconfigurations (in preparation for fault handling testcases)

### LINKED ISSUE #297: Re-enable MainFaultHandler inside MainControlLoop by removing its "disable" flag

tanishqaggarwal:
Closes #297

### Summary of changes
This is a major PTest refactoring.
- There is now a boot sequence controller called `BootUtil` that can be used by testcases to control their boot trajectory. I'll make some documentation shortly on how to use this boot utility.
- Enums are moved out into a separate class called `Enums` to reduce cluttering of the base PTest case.
- `single_sat_sim` is now an option controlled by the testcase, and no longer a configuration item.

There are also some miscellaneous FSW changes. I've documented these via comments on this PR.

The remaining work is to integrate fault handler PTest cases. They currently live on the `fault_ptest_2` branch, but those changes depend on this PR being merged.

### Testing
HOOTLs pass on the existing HOOTL testcases, so the PTest reconfiguration didn't damage their correctness. Same with FSW unit tests.

### Constants
None.

### Telemetry
None.
### Documentation Evidence
The configuration options and available helper functions in the PTest base classes are getting increasingly more complex and worth documenting. I'm going to defer this to #453 but I'll finish the ticket soon so that everyone on FSW can get caught up to the changes quickly.


src/fsw/FCCode/MissionManager.cpp
```diff
@@ -161,6 +161,7 @@ void MissionManager::dispatch_startup() {
             prop_state_t::disabled);
     }
     else {
+        is_deployed_f.set(true);
```
tanishqaggarwal, line 164: This flag needs to be set so that we don't go through deployment wait on every single boot of the satellite.

src/fsw/FCCode/MainControlLoop.cpp
```diff
@@ -71,10 +71,12 @@ MainControlLoop::MainControlLoop(StateFieldRegistry& registry,
     // Since all telemetry fields have been added to the registry, initialize flows
     downlink_producer.init_flows(flow_data);
 
-    // Temporarily disable fault handling until it's better tested
```
tanishqaggarwal, line 74: MainFaultHandler is now enabled by default in FSW by removing this code. (Closes #297)

src/fsw/FCCode/MainControlLoop.cpp
```diff
@@ -71,10 +71,12 @@ MainControlLoop::MainControlLoop(StateFieldRegistry& registry,
...
-    WritableStateField<bool>* fault_handler_enabled_fp =
-        find_writable_field<bool>("fault_handler.enabled", __FILE__, __LINE__);
-    fault_handler_enabled_fp->set(false);
+    #ifndef FLIGHT
```
tanishqaggarwal, line 74: I think it's now a good idea to require HITLs to have a serial connection in order for flight software to run. Otherwise the boot sequence will run haphazardly.

src/fsw/FCCode/GomspaceController.cpp
```diff
@@ -181,7 +181,7 @@ void GomspaceController::execute() {
     get_hk_fault.evaluate(!gs.get_hk());
 
     // Check that the battery voltage is above the threshold
-    low_batt_fault.evaluate(gs.hk->vbatt < batt_threshold_f.get());
+    low_batt_fault.evaluate(vbatt_f.get() < batt_threshold_f.get());
```
tanishqaggarwal, line 184: This change allows PTest to control the low battery fault by adjusting the battery voltage, rather than clumsily suppressing or overriding the fault. While this option isn't currently used, it's a good idea to enable it.

src/fsw/FCCode/FieldCreatorTask.hpp
```diff
@@ -31,6 +33,7 @@ class FieldCreatorTask : public ControlTask<void> {
           add_readable_field(pos_f);
           add_readable_field(pos_baseline_f);
 
+          pos_baseline_f.set({nan_d, nan_d, nan_d});
```
tanishqaggarwal, line 36: Without this change, a boot sequence to the "leader" state quickly becomes a boot sequence to the "docking" state, since the leader will automatically transition to leader_close_approach and docking since it will think it has zero space between itself and the other spacecraft.

ptest/cases/base.py
```diff
@@ -209,85 +100,151 @@ def finish(self):
...
+        """
+        return "manual"
+
+    @property
+    def fast_boot(self):
+        """
+        If true, the boot utility will immediately jump to the initial_state
+        rather than stepping through the state machine.
+        """
         return True
```
shihaocao, line 122: This seems to say we're always going fo fast_boot?

tanishqaggarwal: This means the _default_ is to go to fast boot.

tanishqaggarwal:
Fast boot is ideal for checkouts when we boot to manual. I edited the checkout cases to align to this paradigm. They should be a pretty good example for checkout cases moving forward, i think.

Slow boot is more for mission testcases / fault handler testing where it is important to be in a mission state with the correct initial conditions.

tanishqaggarwal:
By default the SingleSatOnlyCases are configured for the right settings for hardware checkout (manual boot, fast boot), so no additional configuration should be necessary. I think we're done with writing checkout cases, anyway?

tanishqaggarwal:
Merging to allow for the fault handling PR to actually open.

PR #454, opened 2020-05-03
DCDC Control inside MM

### LINKED ISSUE #443: Set DCDC outputs appropriately in MissionManager
Otherwise prop and ADCS won't work!

tanishqaggarwal:
Closes #443.

### Summary of changes
- Add ADCS and SPH DCDC control to MissionManager. ADCS DCDC control is directly managed by the subsystem. SPH DCDC is managed by the MissionManager (since two subsystems, docking and propulsion, depend on it.)

### Testing
Additional checks were added in the MissionManager unit test to ensure that the SPH DCDC is actuated at the right times. It's difficult to write a unit test for ADCSBoxController but the DCDC control is directly coupled to the ADCS state, hence when the ADCS is disabled the DCDC should be as well.

### Constants
None added.

### Telemetry
None added.

### Documentation Evidence
This is increasing conformance of the MissionManager to documentation-specified behavior, so no additional documentation is necessary.


tanishqaggarwal:
Going to merge this in since it cannot _possibly_ break anything right now. We never yet approach the docking state during testing, and ADCS is almost always on and we want DCDC to be on then.

tanishqaggarwal:
Based on existing usages of the DCDC controller state fields, it should be ok to merge this in. The only uses are in hardware checkout cases, where the DCDC pins must be set to true manually anyway since the spacecraft is in the manual state. Everywhere else, it doesn't really matter if the DCDC pins are on or not.

Issue #234, opened 2020-03-01
Gomspace Stops Charging


shihaocao:
The gomspace stops charging once gomspace controller writes settings to the GomSpace. When you comment out the code that writes settings to the gomspace, meaning only read data requests are present, FSW no longer stops GomSpace from charging.

shihaocao:
Discussion from testing on 04/08:

In flight we should be defaulting to PPT auto.
Further, we should only be calling `gs.set_pv_volt()` when `ppt_mode.get() == 2` where 2 is fixed ppt.

Unanswed q's:
You can't charge gomspace with power supply in auto PPT mode or else whack stuff happens. How to handle this?
Gomspace default values how?

shihaocao:
Find default settings on gomspace

shihaocao:
Gomspace should default to Auto PPT in flight, but on the ground, fixed is fine, unsure how to let it charge tho

stewartaslan:
This will be fixed by #386, so it's redundant.

tanishqaggarwal:
Determined to be a symptom of #386: the PPT mode is not actually writable by the Gomspace driver, for some reason.

tanishqaggarwal:
Closing and tracking on #485

PR #456, opened 2020-05-04
Hardcode gs constants

### LINKED ISSUE #386: Set hardcoded constants in the Gomspace control task, rather than the initially read values
From GomspaceController.cpp
```
if (control_cycle_count==1){
        power_cycle_output1_cmd_f.set(false);
        power_cycle_output2_cmd_f.set(false);
        power_cycle_output3_cmd_f.set(false);
        power_cycle_output4_cmd_f.set(false);
        power_cycle_output5_cmd_f.set(false);
        power_cycle_output6_cmd_f.set(false);

        pv1_output_cmd_f.set(gs.hk->vboost[0]);
        pv2_output_cmd_f.set(gs.hk->vboost[1]);
        pv3_output_cmd_f.set(gs.hk->vboost[2]);

        ppt_mode_cmd_f.set(gs.hk->pptmode);

        heater_cmd_f.set(gs.get_heater());

        counter_reset_cmd_f.set(false);
        gs_reset_cmd_f.set(false);
        gs_reboot_cmd_f.set(false);
    }
```

This is a problem, since it's dynamically setting some pretty important values. Instead, we should change these values to be tracked constants so that the effect of flight software on hardware is well-defined.

fatimayousuf:
# Hardcode default values of Gomspace commands

Fixes #386 

### Summary of changes
- Pretty self-explanatory, sets the command state fields of the gomspace controller to tracked constants, instead of reading the default values from the hk struct.

### Testing
I used the existing gomspace controller test. I did change the test a little bit. Before, the command statefields would read off of the hk struct in the first control cycle so that there wouldn't be any unwanted writes. I changed the test to check that the statefield commands are set to the tracked constant values instead.

### Constants
As described above, the default values for the command statefields are all tracked constants and are in the constants.hpp file. I somewhat arbitrarily chose the default constant values for the power volt commands, the ppt mode command, and the heater command, so please do let me know if I should change them.


tanishqaggarwal:
Closes #234

Issue #419, opened 2020-04-21
Creating Command Client


tanishqaggarwal:
Offshoot of #398

PR #457, opened 2020-05-04
Command client skeleton

tanishqaggarwal:
Related to #419 

### Summary of changes
- Skeleton for command client

### Testing
The client starts up fine. Currently there's no functionality inside it so there's really nothing to test.

### Constants and Telemetry
None; this is a ground-side change.

### Documentation Evidence
In principle it's already addressed by the Swagger HTTP endpoint documentation. Perhaps sometime in the future it'll be appropriate to add command client documentation to some kind of "mission manual."


cmdclient/config.json
```diff
@@ -0,0 +1,7 @@
+{
```
tanishqaggarwal, line 1: This is an example config and it's okay to keep it in the repository. Real configs for flight must be hidden

PR #468, opened 2020-05-09
Orbit controller skeleton

tanishqaggarwal:
# Summary of changes
- Orbit controller skeleton

Need to continue adding actual content into this control task.

### Testing
A unit test has been added that just adds initialization. Passing of the CI tests implies that nothing will break in HOOTL or HITL, since this control task currently has no impacts on behavior. (The state fields it sets are never used.)

### Constants
Added a constant for the orbit controller's offset from the control cycle.

### Telemetry
Added inputs and outputs for the orbit controller. Currently, they are unassigned from any flows.

### Documentation Evidence
I'll open a ticket soon to document this in RTD


kkrol27:
Some double array serializer assertion is failing?

PR #475, opened 2020-05-15
Radio testing fixes may

athena255:
# Radio Testing Fixes

Fixes #. 265
### Summary of changes
- radio.state changed from internal state field to readable state field
- minor bug fix for radio_session.py
- minor bug fixes for send_uplink.go
- rest of the changes are due to auto-reformatter

### Testing
- End-to-end radio test: we were able to read state fields on elastic search and were also able to update state fields by sending telemetry packets

### Constants
- No new constants

### Telemetry
- radio.last_comms_ccno 
- radio.state

### Documentation Evidence
- N/A


ptest/send_uplink.go
```diff
@@ -196,13 +196,13 @@ func main() {
...
 	var radioKeys RadioKeys
 	json.Unmarshal([]byte(radioKeyJson), &radioKeys)
 	user := radioKeys.Email_username
 
 	// process the credential file
-	credential, err := ioutil.ReadFile("configs/credentials.json")
+	credential, err := ioutil.ReadFile("ptest/configs/credentials.json")
 	if err != nil {
 		log.Fatalf("Unable to read client secret file: %v", err)
 	}
```
fatimayousuf, line 208: This file can be deleted since the Go script isn't used in radio session anymore

ptest/radio_session.py
```diff
@@ -40,11 +40,12 @@ def __init__(self, device_name, imei, uplink_console, port, send_queue_duration,
         self.flask_app=create_radio_session_endpoint(self)
 
         try:
-            self.http_thread = Process(name=f"{self.device_name} HTTP Command Endpoint", target=self.flask_app.run, kwargs={"port": self.port})
+            self.http_thread = Process(name=f"{self.device_name} HTTP Command Endpoint", target=self.flask_app.run, kwargs={"port": str(self.port)})
```
fatimayousuf, line 43: I don't think you need to cast the port to string haha, my mistake. I recall importing multiprocessing fixed the error you were getting in starting the http endpoints

ptest/radio_session.py
```diff
@@ -11,7 +11,7 @@
 import glob
```
fatimayousuf, line 11: I just merged a pr that made a lot of changes to radio session. Consequentially, I can see that there are a few merge conflicts in this file. So I suggest we could undo of all the changes made in this file, except adding `traceback.print_exc()` in the try/except block?

athena255: Sure, how about I undo all the ptest changes?

fatimayousuf: Uhhh I think the only files in ptest that you changed were radio session and `send_uplink.go` haha

fatimayousuf: I think the only files in ptest that you changed were `radio_session.py` and `send_uplink.go`. It seems that those two files have a bunch of merge conflicts now lol, so I just thought those should be resolved before this pr gets approved and merged in. I also thought that, as you resolved this merge conflicts, it would be good to keep `traceback.print_exc()` in the try/except block in case we have trouble starting up the http endpoints again.

configs/flight_and_test_envs.ini
```diff
@@ -77,7 +77,7 @@ lib_deps = ${fsw_common.lib_deps}
 lib_archive = false
 lib_compat_mode = off
 test_build_project_src = true
-test_filter = test_fsw_*
+test_filter = test_fsw_quake*
```
fatimayousuf, line 80: I think this should be changed back to `test_fsw_*`

Issue #292, opened 2020-03-12
Test Piksi fault response in ptest


tanishqaggarwal:
Depends on #210 
Depends on #460

tanishqaggarwal:
We should make this a lower-priority item, since we could nix this fault handler entirely and monitor the Piksi fault conditions on the ground manually. There are other controls (such as safehold low-batt) that would initiate a stop of the mission anyway.

Issue #290, opened 2020-03-12
Test Quake fault response in ptest


tanishqaggarwal:
Ensure that the Quake fault machine steps through its paces in ptest, i.e. that mission states change and that the quake gets forced into powercycling states.

PR #479, opened 2020-06-13
Fault testing in ptest

### LINKED ISSUE #299: Test "super simple" fault handlers in ptest

tanishqaggarwal:
Fixes #299, #290.

### Summary of changes
Add fault response handling in mission manager (+ unit test)
Simplify quake fault handler logic (+ unit test)
PTest fault handling cases for:
- Wheel faults
- Low battery fault
- Quake fault handling

There is additional work to be done  for the Piksi fault handler testing, tracked in #292.

### Testing
I verified that the QuakeFaultHandler_Fast, ADCSWheelFaultHandler, and LowBattFaultHandler ptest cases pass (see CI for the latter two, and the [HOOTL logs](https://cornellprod-my.sharepoint.com/:f:/g/personal/saa243_cornell_edu/EhXyXVrpMTJKkf1tXGyNit4BobNxT7913W2pOnKIxtfhdA?e=maFIRy) for the first case.) I've also posted HOOTL logs for the testcases that require sim and therefore can't be checked in CI, since a base feature of the code changed.

Unit test changes work since they pass.

More complete testing of ADCSWheelFaultHandler and LowBattFaultHandler will require HITL. **This PR should not be merged without this testing.**

### Constants
No significant changes, just line number edits.

### Telemetry
No changes. 

### Documentation Evidence
No existing functionality was changed; only implementation details were changed, and tests were added. I don't think this PR necessitates additional documentation.


tanishqaggarwal:
PR testing still needs to be fixed up a bit.

ptest/cases/prop_state_machine_case.py
```diff
@@ -178,17 +178,17 @@ def print_object(self):
 
     @property
     def min_num_cycles(self):
-        return  (int(self.ctrl_cycles_per_filling) + int(self.ctrl_cycles_per_cooling))*int(self.max_pressurizing_cycles)
+        return (int(self.ctrl_cycles_per_filling) + int(self.ctrl_cycles_per_cooling))*int(self.max_pressurizing_cycles) + 4
```
kkrol27, line 181: What's up with this `+4`?

ptest/cases/quake_fault_handler.py
```diff
@@ -1,110 +1,126 @@
...
         elif self.test_stage == "powercycles":
             if not hasattr(self, "powercycles_count"):
                 self.powercycles_count = 0
 
-            if not self.mission_state == "standby":
-                raise TestCaseFailure("QuakeFaultHandler did not force satellite into standby after 24 hours of no comms.")
-
-            if self.comms_blackout_timer > self.one_day_ccno / 3:
-                # TODO check if Quake radio has been powercycled.
+            self.cycles_since_blackout_start += 1
```
kkrol27, line 91: I think we can move `self.cycles_since_blackout_start += 1` outside the `if` statements? It seems to be in every branch.

kkrol27: Scratch that - read it wrong.

ptest/runsim.py
```diff
@@ -247,6 +248,10 @@ def main(args):
                 For the current run, a subdirectory of DATA_DIR is created in which the actual data is stored.''', default=log_dir)
     args = parser.parse_args(args)
 
+    if args.clean:
+        print("Removing EEPROM file due to user request.")
+        os.system("rm -f eeprom.json")
```
kkrol27, line 253: I'd look at `os.remove` here for compatibility reasons - even tho we really only run on unix.

src/fsw/FCCode/MissionManager.cpp
```diff
@@ -95,17 +95,23 @@ void MissionManager::execute()
...
+
+    if (!(state == mission_state_t::startup || state == mission_state_t::manual))
     {
-        transition_to(mission_state_t::standby, adcs_state_t::point_standby, prop_state_t::idle);
-        return;
+        if (fault_response == fault_response_t::safehold)
+        {
+            transition_to(mission_state_t::safehold, adcs_state_t::startup, prop_state_t::disabled);
+            return;
+        }
```
kkrol27, line 105: Transitioning from safehold to safehold is alright?

src/fsw/FCCode/QuakeFaultHandler.cpp
```diff
@@ -95,50 +94,26 @@ fault_response_t QuakeFaultHandler::dispatch_powercycle_1()
...
-    {
-        transition_to(qfh_state_t::unfaulted);
-        return fault_response_t::none;
-    }
-    else
-    {
-        return fault_response_t::safehold;
-    }
+fault_response_t QuakeFaultHandler::dispatch_safehold() {
+     if (radio_is_disabled() || less_than_one_day_since_successful_comms()) {
```
kkrol27, line 110: Is the radio ever disabled...? Not sure if that should be allowed to happen.

test/test_fsw_mission_manager/test_mission_manager.cpp
```diff
@@ -295,6 +295,68 @@ void test_dispatch_undefined() {
...
+                                         mission_state_t::detumble})
+    {
+        TestFixture tf{initial_state};
+        tf.set(fault_response_t::standby);
+        tf.step();
+        tf.check(initial_state);
+        tf.set(fault_response_t::safehold);
+        tf.step();
+        tf.check(mission_state_t::safehold);
+    }
```
kkrol27, line 336: Just want to be clear that we would return to initialization hold if we exited there for safehold? Perhaps add a clarifying comment here.

ptest/cases/prop_state_machine_case.py
tanishqaggarwal: This is due to a +4 in PropController. @athena255 could you explain?

src/fsw/FCCode/QuakeFaultHandler.cpp
tanishqaggarwal: It will be disabled in the startup state; the Quake Fault Handler should not be running at that time.

test/test_fsw_mission_manager/test_mission_manager.cpp
tanishqaggarwal: Not sure what you mean

src/fsw/FCCode/MissionManager.cpp
tanishqaggarwal: There's no special logic in the transition_to function for safehold, so it should be fine.

.github/workflows/hootl.yml
```diff
@@ -32,11 +41,17 @@ jobs:
...
     - name: Hardware Checkout Cases
       run: |
-        python -m ptest runsim -c ptest/configs/ci.json -t DCDCCheckoutCase -ni
-        python -m ptest runsim -c ptest/configs/ci.json -t GomspaceCheckoutCase -ni
-        python -m ptest runsim -c ptest/configs/ci.json -t SpinMotorsCase -ni
-        python -m ptest runsim -c ptest/configs/ci.json -t ADCSCheckoutCase -ni
+        python -m ptest runsim -c ptest/configs/ci.json -t DCDCCheckoutCase -ni 
+        python -m ptest runsim -c ptest/configs/ci.json -t GomspaceCheckoutCase -ni 
+        python -m ptest runsim -c ptest/configs/ci.json -t SpinMotorsCase -ni 
+        python -m ptest runsim -c ptest/configs/ci.json -t ADCSCheckoutCase -ni
```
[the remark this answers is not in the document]
tanishqaggarwal: These are needless.
```suggestion
        python -m ptest runsim -c ptest/configs/ci.json -t DCDCCheckoutCase -ni
        python -m ptest runsim -c ptest/configs/ci.json -t GomspaceCheckoutCase -ni
        python -m ptest runsim -c ptest/configs/ci.json -t SpinMotorsCase -ni
        python -m ptest runsim -c ptest/configs/ci.json -t ADCSCheckoutCase -ni
```

ptest/runsim.py
tanishqaggarwal: Agreed.
```suggestion
        try: 
            os.remove("eeprom.json")
        except OSError:
            continue
```

tanishqaggarwal:
There's no PRs dependent on this one, so you're not blocking me @kkrol27 . This testing is extremely important and I'd prefer as many eyes as possible, so I'll leave this up for as long as it takes for a thorough review.

test/test_fsw_mission_manager/test_mission_manager.cpp
kkrol27: Say we deployed. Prior to exiting initialization hold we get kicked to safehold. When we exit safehold what state do we enter?

src/fsw/FCCode/QuakeFaultHandler.cpp
kkrol27: Ah got it - sounds good.

ptest/cases/safehold_reboot.py
```diff
@@ -0,0 +1,39 @@
...
+            else:
+                self.logger.put("Satellite now in in safehold state.")
+                self.test_stage = "reboot"
+
+        elif self.test_stage == "reboot":
+            self.logger.put("Expecting satellite to reboot.")
+            self.test_stage = "await_reboot"
+
+        elif self.test_stage == "await_reboot":
+            self.finish()
```
kkrol27, line 39: One other question/comment is whether or not it'd be useful to abstract away this testcase structure in some fashion. It seems to be fairly pervasive in the codebase.

ptest/cases/piksi_fault_handler.py
```diff
@@ -0,0 +1,57 @@
...
+    def initial_state(self):
+        return "leader"
+
+    @property
+    def fast_boot(self):
+        return False
+
+    @property
+    def one_day_ccno(self):
+        return 60 * 1000 // 170
```
shihaocao, line 19: what is the 170 supposed to denote?

tanishqaggarwal: The control cycle time in HOOTL.

ptest/cases/safehold_reboot.py
tanishqaggarwal: Yeah I agree, it's also only going to become more pervasive. Not sure what a good mechanism for this would be, though, and I'm worried that we'd end up sacrificing readability

ptest/cases/prop_state_machine_case.py
```diff
@@ -212,60 +211,73 @@ def cycle_until_change(self, verbose=False, max_cycles=0):
...
                     print("Executing Cycle #: " + str(i))
-                    self.print_object()
         return str(-1)
-            
+
     def test_disabled_to_idle(self):
         print("[TESTCASE] test_disabled_to_idle: manually setting prop to idle")
-        assert self.read_state("dcdc.SpikeDock_cmd") == "true"
-        self.state = Enums.prop_states.get_by_name("idle")
+        if not self.read_state("dcdc.SpikeDock_cmd") == "true":
```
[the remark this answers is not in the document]
tanishqaggarwal: Thanks for the catch! Going to leave this unresolved since this will require retesting the case in HITL.

test/test_fsw_mission_manager/test_mission_manager.cpp
tanishqaggarwal: There's no autonomous commanding out of safehold, so it would be up to us on the ground to command back into init_hold.

Of course if we didn't know the previous state was init_hold this will be pretty risky...let me open a ticket to track this issue.

ptest/cases/prop_state_machine_case.py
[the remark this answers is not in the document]
tanishqaggarwal: Seems like this causes a failure in HOOTL, likely because the flags aren't coupled together in HOOTL the way they are in real hardware.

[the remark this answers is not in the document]
tanishqaggarwal: Hm nvm, looks like they are coupled...I'll investigate further.

ptest/cases/prop_state_machine_case.py
athena255: There are four state transitions from idle to firing.

ptest/cases/prop_state_machine_case.py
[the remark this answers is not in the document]
shihaocao: Was there ever a resolution to this?

ptest/cases/piksi_fault_handler.py
```diff
@@ -0,0 +1,57 @@
...
+        return \
+        """
+        NOTE: This test requires Flight Software to be compiled without the
+        \"SPEEDUP\" flag so that the timeouts for the Piksi Fault
+        Handler are realistic.\n
+        """
+
+    @property
+    def one_day_ccno(self):
+        return 24 * 60 * 1000 // 170
```
shihaocao, line 57: one day = 24 hrs = 24 * 60 minutes = 24 * 60 * 60 seconds = 24 * 60 * 60 * 1000 milliseconds; We're missing a * 60 right?

ptest/cases/quake_fault_handler.py
```diff
@@ -1,110 +1,126 @@
...
+        self.powercycle_happening = None
 
+    def setup_post_bootsetup(self):
         self.ws("fault_handler.enabled", True)
 
-        self.test_stage = "waiting_for_boot"
+    def check_quake_powercycled(self):
+        if not self.powercycle_happening:
+            print("Quake radio was not powercycled.")
+        self.logger.put("Comms blackout caused a powercycle of Quake.")
```
shihaocao, line 42: this should be indented once more right?

ptest/cases/quake_fault_handler.py
```diff
@@ -1,110 +1,126 @@
...
-        if self.test_stage == "waiting_for_boot" and self.run_boot_to_leader():
-            self.logger.put("[TESTCASE] Satellite has booted to leader. Starting a 24-hours comms blackout (in satellite time; real duration of blackout is 1 min).")
-            self.test_stage = "first_no_comms" 
-
-        elif self.test_stage == "first_no_comms":
-            if self.comms_blackout_timer > self.one_day_ccno:
+    def run_case_singlesat(self):
+        self.collect_diagnostic_data()
+
+        if not hasattr(self, "test_stage"):
```
shihaocao, line 53: is this `if not hasattr(self, "test_stage"):` line necessary? it seems like it would always be true on the first pass.

I guess my greater question is how this entire case steps forward a cycle without a `self.cycle()` command?

ptest/cases/quake_fault_handler.py
```diff
@@ -1,110 +1,126 @@
...
-
-        self.one_day_ccno = 24 * 60 * 60 * 1000 / 170
-        self.logger.put("""[TESTCASE] 
-NOTE: This test requires Flight Software to be compiled with the
-\"SHORTEN_DAYS\" flag so that the timeouts for the Quake Fault
-Handler only take minutes, not hours.""")
-        self.logger.put("[TESTCASE] Waiting for the satellite to boot to standby.")
-
+    def one_day_ccno(self):
+        return 24 * 60 * 1000 // 170
```
shihaocao, line 126: * 60?

ptest/cases/utils.py
```diff
@@ -229,7 +227,9 @@ def run_boot_sequence(self):
                     raise TestCaseFailure("Satellite went to initialization hold instead of detumble.")
                 else:
                     raise TestCaseFailure(f"Satellite failed to exit deployment wait period. \
-                        Elapsed deployment period was {true_elapsed}.")
+                        Satellite state is {satellite_state}. Elapsed deployment period was {true_elapsed}.")
+            else:
+                self.elapsed_deployment += 1
 
         elif self.boot_stage == 'detumble_wait':
             self.num_detumble_cycles += 1
```
shihaocao, line 235: is there a reason why the `self.num_detumble_cycles += 1` was moved in the above block, but not down here?

test/test_fsw_fault_handler/test_main_fh_end_to_end.cpp
```diff
@@ -37,7 +37,6 @@ void test_single_simple_safehold_faults() {
         tf.adcs_wheel3_adc_fault_fp,
         tf.adcs_wheel_pot_fault_fp,
         tf.low_batt_fault_fp,
-        tf.prop_overpressure_fault_fp
```
shihaocao, line 40: what's happening with this fault?
Just checking if this was removed intentionally

src/fsw/FCCode/QuakeManager.cpp
```diff
@@ -268,8 +268,10 @@ void QuakeManager::dispatch_transceive()
             return handle_no_comms();
         }
 
-        // We have comms, so update the last_checkin time
-        last_checkin_cycle_f.set(control_cycle_count);
+        #if !(defined(DESKTOP)) || defined(UNIT_TEST)
+            // We have comms, so update the last_checkin time
+            last_checkin_cycle_f.set(control_cycle_count);
```
shihaocao, line 273: This looks like an important line critical to fsw, why would we only set this during UNIT_TEST?

ptest/cases/quake_fault_handler.py
shihaocao: I think this is my answer to my second question:

# Step 3.4. Step the flight computer forward.
            if self.is_single_sat_sim:
                self.flight_controller.write_state("cycle.start", "true")
            else:
                self.flight_controller_follower.write_state("cycle.start", "true")
                self.flight_controller_leader.write_state("cycle.start", "true")

Which also answers why the hasattr is there

ptest/cases/piksi_fault_handler.py
tanishqaggarwal: Good catch! Good thing this fault handler isn't implemented yet.

ptest/cases/quake_fault_handler.py
tanishqaggarwal: Should be, yes.

Btw: to make these iterations faster, there's a way to suggest code changes inline.

src/fsw/FCCode/QuakeManager.cpp
tanishqaggarwal: This also happens in flight. The condition is "if (not desktop) or (unit_test)" which is equivalent to "if (flight or HITL or unit test)".

ptest/cases/utils.py
tanishqaggarwal: Could you clarify what you mean by this comment?

test/test_fsw_fault_handler/test_main_fh_end_to_end.cpp
tanishqaggarwal: The overpressure fault handler is not yet implemented, so I removed it from this list.

ptest/cases/prop_state_machine_case.py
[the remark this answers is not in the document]
tanishqaggarwal: Resolved now! I had to add an extra "cycle" in the post-boot setup function.

src/fsw/FCCode/QuakeManager.cpp
shihaocao: Gotcha

ptest/cases/utils.py
shihaocao: In lines 231 and 232, you moved the `self.elapsed_deployment += 1` into an `else` block, but not the `self.elapsed_deployment += 1` on line 235

ptest/cases/quake_fault_handler.py
```diff
@@ -1,110 +1,127 @@
...
 
+            if self.cycles_since_blackout_start > self.one_day_ccno // 3:
+                self.check_quake_powercycled()
                 self.powercycles_count += 1
-                self.logger.put(f"[TESTCASE] Powercycled {self.powercycles_count} time(s).")
-                self.comms_blackout_timer = 0
-                if self.powercycles_count == 3:
-                    self.logger.put("[TESTCASE] Transitioning to safehold.")
+                self.cycles_since_blackout_start = 0
+                if self.powercycles_count >= 3:
```
shihaocao, line 98: Why 3?

src/fsw/FCCode/MissionManager.cpp
```diff
@@ -95,17 +95,23 @@ void MissionManager::execute()
...
-    else if (fault_response == fault_response_t::standby &&
-             state != mission_state_t::safehold && state != mission_state_t::standby)
+
+    if (!(state == mission_state_t::startup || state == mission_state_t::manual))
     {
-        transition_to(mission_state_t::standby, adcs_state_t::point_standby, prop_state_t::idle);
-        return;
+        if (fault_response == fault_response_t::safehold)
+        {
+            transition_to(mission_state_t::safehold, adcs_state_t::startup, prop_state_t::disabled);
```
shihaocao, line 103: If we transition adcs to startup, that sends the ADCS into passive and turns off the ADCS DCDC.

I recall that we had a lot of problems with ADCS motors and them not turning if the DCDC's were not on before sending commands.

If we come back out of startup, the DCDC's will basically be turned on only a millisecond before the ADCS could potentially be getting motor commands. (Where as in my ADCS testing, the DCDC was always on a full control cycle before motor commands were sent.)

@kkrol27 Have we ever done any testing where we leave the ADCS box on, but turn the DCDC off and then back on again? I'm worried we could possibly lock up the ADCS box if we cycle DCDC.

Alternatively, do we need to turn the DCDC off?

test/test_fsw_mission_manager/test_mission_manager.cpp
shihaocao: `Init_hold` is only the 30 mins post deployment right? By the time we get to dealing with the errors/faults 30 mins would probably have passed by then. So I don't think there will be a scenario we go back to init_hold

ptest/cases/quake_fault_handler.py
tanishqaggarwal: This is an arbitrary decision. We've decided to try to powercycle the quake 3 times before going to safehold.

ptest/cases/utils.py
tanishqaggarwal: I think I should be able to fix this to be more consistent.

src/fsw/FCCode/MissionManager.cpp
tanishqaggarwal: Good point here. Waiting for Kyle's response.

ptest/cases/quake_fault_handler.py
shihaocao: Just for documentation: I'm worried that this code (as is) is dependent on an assumption that a powercycle takes only 1 control cycle to occur. 

Per @tanishqaggarwal, the gompace slowly ramps down voltage, so it could take around 30 seconds to reboot quake.

test/test_fsw_mission_manager/test_mission_manager.cpp
kkrol27: Good with me then.

src/fsw/FCCode/MissionManager.cpp
kkrol27: Hmm this is a good point - two things:

* I don't recall the situation exactly but I thought the issue we encountered arose because the DCDC control task was running after the ADCS had been turned on or did we fix that and still have to introduce a control cycle slip? Did we ever explicitly write this down somewhere?
* It may be worth introducing a little extra logic to let the DCDC's be in "bootup" to a cycle before being considered powered up? This could easily alleviate this issue.

ptest/cases/quake_fault_handler.py
kkrol27: Do we have a way of reading the voltage levels off the GOMSpace?

src/fsw/FCCode/MissionManager.cpp
tanishqaggarwal: I second the bootup idea. @athena255 thoughts?

tanishqaggarwal: Also @fatimayousuf since you implemented DCDC logic

shihaocao: As it stands, I think it might be okay to just push this current section of the code being pushed since I think transitioning to `adcs::startup` is probably the right move (and turning off the DCDC).

Maybe we can create a ticket to do further testing and confirm if a specific pre-booting of DCDC is necessary?

tanishqaggarwal: Ah, sorry, I confused myself and consequently everyone.

I think DCDCs have negligible startup time, so a bootup might not be necessary there. The issue is that gomspace output cycling takes non-negligible time. So the "bootup" thing should happen for Gomspace cycles.

Can you create a ticket to track the ADCS/DCDC stuff, Shihao?

tanishqaggarwal: (Deleted a previous comment because I confused myself twice)

A ticket sounds like a great idea. Could you create it and schedule it?

ptest/cases/quake_fault_handler.py
tanishqaggarwal: @kkrol27 Unfortunately no, but I think we should create some additional logic to handle the 30-second cycling time.

src/fsw/FCCode/MissionManager.cpp
shihaocao: Created #487

ptest/cases/quake_fault_handler.py
kkrol27: @shihaocao do you want to open a ticket to track GOMSpace bootup changes?

tanishqaggarwal: I can do it.

tanishqaggarwal: #489

tanishqaggarwal:
I'm going to merge this PR and track the HITL testing on a separate ticket (#490). Adding this faults stuff into master is pretty important since the semester's starting up and there will need to be more development on this front.

PR #483, opened 2020-08-30
Remove quake control task

tanishqaggarwal:
### Summary of changes
- Remove Quake Control Task since it's obsoleted by changes to QuakeManager.

### Testing
Existing unit tests are sufficient to catch regressions. This component is completely unused in software, so it's OK to remove.

### Constants
No changes.

### Telemetry
No changes.

### Documentation Evidence
We might have to shift around the structure on ReadTheDocs a bit since QuakeControlTask is no longer there. @athena255 could you look into this?


tanishqaggarwal:
Seems like this component is actually being used still by QuakeManager. Oops.

Issue #491, opened 2020-09-08
GomspaceCheckoutCase is broken in HOOTL


tanishqaggarwal:
There is an error message 

```
Power cycled outputs could not turn on after 600 cycles (1 minute)
Could not update power_cycle_output0
Could not update power_cycle_output1
Could not update power_cycle_output2
Could not update power_cycle_output3
Could not update power_cycle_output4
Could not update power_cycle_output5
```

even when using `fsw_native_leader_realtime`. Need to check why this is, and if it's happening in HITL...

PR #492, opened 2020-09-08
Fix testcase timings

tanishqaggarwal:
### Summary of changes
- Add control cycle time and "24 hour" time durations as state fields so that PTest can read them and change its behavior accordingly.
  - This correspondingly allows destroying the distinction between "Realtime" and "Fast" Ptest cases, and instead lets the code logic determine how long the test will take.
- Edit out comms blackout fault-handling code from MissionManager; it's no longer needed.
- Make various improvements to comments and functionality within PTest code.

### Testing
No additional functionality was added, so existing testcases are sufficient. Here I'm including a [log of the QuakeFaultHandler run](https://cornellprod-my.sharepoint.com/:f:/g/personal/saa243_cornell_edu/EjuTKsKYCsFMv8eX2PrCV30BhthRV5Yc3rxx-ug53Ixelg?e=5OOkXH), since this testcase can't run in CI.

During the course of this implementation I discovered the bug #491 .

### Constants
- Removed radio silence duration constant from MissionManager since it should no longer care about radio silence; QuakeFaultHandler manages that.
- Changed the deployment wait to be a function of `PAN::one_day_ccno` so that all long-duration timers in the code are based on a consistent baseline.
- Added the Gomspace powercycling period as a constant (previously it was hardcoded.)
- Reduced the "one_day_ccno" duration in SPEEDUP operations, so that testcases run faster.

### Telemetry
As mentioned above, I added some fields so that PTest can ingest data about the flight software's timing setup. These fields do not need to be exported to flows since they're PTest-only.

I also removed max_radio_silence from the telemetry since the field is no longer used. Should we consider adding the Quake Fault Handler timings as configurable telemetry points?

### Documentation Evidence
No new functionality was added. Existing tickets related to documenting PTest capture a lot of the code that was touched.


athena255 (approved):

fatimayousuf (approved):
:)

PR #503, opened 2020-09-23
Cmdclient

knk38:
# Adding A Command Client to interact with RadioSession endpoints

### Summary of changes
- implemented methods to pause/unpause the uplink timer
- implemented methods to send/remove uplinks
- implemented the method to list all queued uplinks

### Testing
-When testing, I added print statements in each command that printed the content of the http GET/POST as well as the HTTP error code. (An code of 200 signified that the connection to the endpoint was successful) I made sure that each command printed 200.
-For the send and remove commands, I added a print statement in radiosession.py that printed "uplink sent" or "uplink removed" instead of actually sending the uplink to Iridium.

### Documentation Evidence
-there wasn't much documentation to be added, as I was implementing already documented methods


cmdclient/config.json
```diff
@@ -1,7 +1,7 @@
 {
     "FlightControllerFollower" : {
         "type" : "RadioSession",
-        "server" : "127.0.0.1",
-        "port" : 8000
+        "server" : "",
+        "port" :
```
knk38, line 5: I wanted to keep this change to hide the server/port?

fatimayousuf: well it's just localhost, so it's not anything confidential

cmdclient/__main__.py
```diff
@@ -48,33 +53,58 @@ def do_list(self, args):
...
         """
-        raise NotImplementedError
+        args = str(args).split()
+        if len(args) > 2:
+            print('You have specified too many arguments')
+        elif len(args)< 2:
+            print("You have not specified enough arguments use 2 arguments: <field> <value>")
+        else:
+            endpoint = self.url+"/remove"
+            r = requests.post(url = endpoint, data = {"field":args[0],"value":args[1]})
```
tanishqaggarwal, line 89: This will only handle the first field in a set of fields.

cmdclient/__main__.py
```diff
@@ -48,33 +53,58 @@ def do_list(self, args):
...
+        args = str(args).split() #splits the string args into an array of elements
+        if len(args) > 2:
+            print('You have specified too many arguments')
+        elif len(args)< 2:
+            print("You have not specified enough arguments use 2 arguments: <field> <value>")
+            #checks that the proper formatting has been followed, add <field> <value>
+
+        else:
+            endpoint = self.url+"/send-telem"
+            r = requests.post(url = endpoint, data = {"field":args[0],"value":args[1]})
```
tanishqaggarwal, line 75: This will only handle the first specified field.

ptest/http_cmd.py
```diff
@@ -128,10 +128,10 @@ def send_telem():
         to = "data@sbd.iridium.com"
         sender = "pan.ssds.qlocate@gmail.com"
         subject = imei
-        SendMessage(sender, to, subject, "", "", 'uplink.sbd')
-
+        #SendMessage(sender, to, subject, "", "", 'uplink.sbd')
+        print("sent uplink")
```
fatimayousuf: I think we should keep SendMessage, since that is what will send the actual email to iridium

cmdclient/config.json
tanishqaggarwal: Makes sense, although hiding this from the rest of the world isn't super critical since this is a general server configuration. Up to you if you want to leave it in

tanishqaggarwal (changes requested):
Just a quick little suggested change

cmdclient/__main__.py
```diff
@@ -48,33 +53,58 @@ def do_list(self, args):
...
 
     def do_add(self, args):
         """
         Adds a set of state fields to the currently queued uplink. The command may fail if
         the packet size after adding the state fields would exceed the maximum allowable uplink
         size (currently 70 bytes.)
         """
-        raise NotImplementedError
+        args = str(args).split() #splits the string args into an array of elements
+        if len(args) > 2:
```
fatimayousuf, line 67: So if you go to ptest/endpoint_configs/radio_session/send_telem.yml, you'll see that this endpoint accepts a list of son objects. The list can be as long as you want. However, each of the json objects should only have two kay-value pairs, field and value. So an example of an args input might be:

[ {
          "field": dahkj
          "value": adkld
     },
     {
          "field":12daji34,
          "value: 380196
     },
     {
          "field": kodkwo
          "value": 8307
} ]

you should go through each json element of the list and check that there are only two key-value pairs: the field and the value.

cmdclient/__main__.py
```diff
@@ -48,33 +53,58 @@ def do_list(self, args):
...
+            r = requests.post(url = endpoint, data = {"field":args[0],"value":args[1]})
+            print(r.content)
 
     def do_remove(self, args):
         """
         Removes a set of state fields from the currently queued uplink.
         """
-        raise NotImplementedError
+        args = str(args).split()
+        if len(args) > 2:
```
fatimayousuf, line 83: Same comment as send_telem. look at ptest/endpoint_configs/radio_session/remove.yml to see the format of the input

fatimayousuf (changes requested):
Sorry it took me a while. I truly do not remember what I did for radio session, so I reviewed my old work  before looking at this pr 😅

cmdclient/__main__.py
knk38: Okay thanks for this clarification, i fixed my helper method to follow this format for converting args

cmdclient/config.json
knk38: imma keep it secret just in case 😎

tanishqaggarwal (approved):

fatimayousuf (approved):
😄

PR #508, opened 2020-10-03
MTR flight software binary

tanishqaggarwal:
### Summary of changes
- Creates a simplified flight software binary that can be used for MTR Ptesting. It runs with a 50hz control loop including the debug task.
- Make build system changes
- Consolidate ptest code for automatically compiling correct binaries.
   - **Note that this change affects all flight software developers** since the `binary_filepath` config argument has been renamed to `pio_target`. The reason for this is that nobody needs a full system-wide path anymore so a) it shortens the configuration, b) it reduces the amount of string-mangling that would be required in the `get_pio_asset` function.

### Testing
The `fsw_native_mtr_ptest` and `fsw_teensy36_test_mtr` environments works fine using ptest's EmptyCase in HITL and HOOTL. This should be sufficient for the sake of the MTR ptesting.

[Testing logs link](https://cornellprod-my.sharepoint.com/:f:/g/personal/saa243_cornell_edu/Er3Gt-77lcxIvFphgsFYnS4Bc4owJW9rIvtxv9x1QhDN3w?e=zHkaaA)

### Documentation Evidence
We should make this environment part of the procedure for MTR Ptesting. After we write that procedure we should add its OneDrive link to the PIO config sections defining its environment so that the purpose for the environment does not get lost.


kkrol27 (approved):
Looks good to me - just a couple suggestions for the future. Also curious, what's the purpose for making a completely separate binary just for MTR testing rather than disabling most stuff within the standard flight binary?

platformio.ini
```diff
@@ -95,6 +95,8 @@ build_flags =
   -DLIN_RANDOM_SEED=6828031
   -DUNITY_INCLUDE_DOUBLE
   -DNDEBUG
+  -D SERIAL3_RX_BUFFER_SIZE=512
+  -D SERIAL4_RX_BUFFER_SIZE=1024
```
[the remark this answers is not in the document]
tanishqaggarwal: Oh yeah I guess they would be. I forgot that ADCS is affected by this platform. Shouldn't matter tho because ADCS doesn't use these ports

ptest/__init__.py
```diff
@@ -0,0 +1,13 @@
...
+def get_pio_asset(asset):
+    """
+    Searches for a PIO-compiled utility and compiles it
+    if it is not available.
+    """
+
+    asset_filepath = f".pio/build/{asset}/program" 
+    if not os.path.exists(asset_filepath):
+        print(f"Compiling PlatformIO target {asset}.")
+        os.system(f"pio run -e {asset} > /dev/null")
```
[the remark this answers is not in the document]
tanishqaggarwal: We could but it would take me extra overhead to figure out how the PIO python module works. This is a perfectly fine workaround atm

tanishqaggarwal:
We have to create a new binary because of flight software control cycles being ~5 Hz. We want something about 10x faster, so stripping down flight software and modifying control cycle offsets is essential.

src/fsw/targets/mtr_ptest.cpp
```diff
@@ -0,0 +1,164 @@
...
+    class FieldCreator {
+      private:
+        WritableStateField<unsigned char> adcs_state_f;
+        ReadableStateField<double> time_f;
+        ReadableStateField<lin::Vector3d> pos_f;
+        ReadableStateField<lin::Vector3d> pos_baseline_f;
+
+      public:
+        FieldCreator(StateFieldRegistry& r) :
+            adcs_state_f("adcs.state", Serializer<unsigned char>()),
```
knk38, line 51: Should I be reading these state fields instead of mtr_mode or should I also read these in when the test case starts and ends?

tanishqaggarwal: I feel like you probably only have to touch `adcs.state` once, maybe to make it manual. Perhaps never, since there's no logic driving the ADCS's actuators at the moment, so `adcs.state` is effectively not used.

knk38 (approved):
Overall I have no burning questions, all the changes make sense to me, just wondering about those additional state fields in mtr_ptest.cpp (i left a comment there)

src/fsw/FCCode/ADCSBoxMonitor.cpp
```diff
@@ -124,7 +124,6 @@ static lin::Vector<float, N> to_linvector(const std::array<float, N>& src) {
 }
 
 void ADCSBoxMonitor::execute(){
-
     //define nan
     const float nan = std::numeric_limits<float>::quiet_NaN();
```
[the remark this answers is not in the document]
tanishqaggarwal: Fixing

src/fsw/targets/mtr_ptest.cpp
tanishqaggarwal: Probably best to minimize state field interactions to the bare minimum, otherwise the control loop will get slowed down.

src/fsw/targets/mtr_ptest.cpp
```diff
@@ -0,0 +1,164 @@
...
+            time_f("orbit.time", Serializer<double>(0.0, 18'446'744'073'709'551'616.0, 64)),
```
athena255, line 52: what is the meaning of these constants?

src/fsw/targets/mtr_ptest.cpp
```diff
@@ -0,0 +1,164 @@
...
+     * that they can instantiate state fields expected by
+     * the tasks required by the control loop. The unnecessary
+     * control tasks are marked as such.
+     */
+    FieldCreator field_creator_mtr;
+    ClockManager clock_manager;
+    ADCSBoxMonitor adcs_monitor;
+    AttitudeEstimator attitude_estimator; // This control task is only used to initialize state fields.
+    AttitudeComputer attitude_computer; // This control task is only used to initialize state fields.
+    ADCSCommander adcs_commander; // This control task is only used to initialize state fields.
```
athena255, line 87: Aren't we getting rid of ADCSCommander soon? When we do, will this not work anymore?

src/fsw/targets/mtr_ptest.cpp
```diff
@@ -0,0 +1,164 @@
...
+    static constexpr unsigned int dcdc_controller_offset = 0;
+    static constexpr unsigned int adcs_monitor_offset = 500;
+    static constexpr unsigned int debug_task_offset = 5000;
+    static constexpr unsigned int adcs_box_controller_offset = 10000;
+
+  public:
+    MTRTestControlLoop(StateFieldRegistry& registry)
+    : ControlTask<void>(registry),
+      ADCS_INITIALIZATION,
+      dcdc("dcdc"),
```
athena255, line 113: Could you instantiate this in the order that they were declared above? That would be helpful if we ever need to add to this file.

src/fsw/targets/mtr_ptest.cpp
```diff
@@ -0,0 +1,164 @@
...
+        one_day_ccno_f.set(24 * 60 * 60 * 1000 / control_cycle_ms_f.get());
+
+        #ifndef DESKTOP
+            // Allow for PTest testing to occur in a controlled way.
+            while(!Serial) {}
+        #endif
+    }
+
+    void execute() {
+        clock_manager.execute();
```
athena255, line 147: Do you anticipate having to add more control tasks to this file in the future? If so, have you consider putting the control tasks into a vector of some sort and then loop through the vector calling execute/execute_on_time?

src/fsw/targets/mtr_ptest.cpp
tanishqaggarwal: Yeah...but all the state fields in ADCSCommander will just get moved to other control tasks, so fixing this binary should just require a one-line removal of the ADCSCommander.

In any case this binary doesn't serve much utility once we've actually done the MTR testing

src/fsw/targets/mtr_ptest.cpp
tanishqaggarwal: This is just copied over from `FieldCreatorTask` and I'm actually not sure why the limit on time is so weird. @kkrol27 ?

src/fsw/targets/mtr_ptest.cpp
tanishqaggarwal: We shouldn't need to add more tasks. I suppose we could probably do this for MainControlLoop but there's also value in being declarative about which tasks are being run. Refactoring to use a vector still means there is some list of control tasks somewhere.

src/fsw/targets/mtr_ptest.cpp
tanishqaggarwal: Hmm the instantiation order should be compiler-enforced, I think. I think it's in the right order.

athena255 (approved):

PR #511, opened 2020-10-04
Add bootcount to EEPROM

### LINKED ISSUE #248: Add bootcount to EEPROM
The EEPROM is a 2 KB permanent storage device for the Teensy; it can be used to retain state values if the Teensy unexpectedly reboots.

One of the items we should save to EEPROM is the number of times the satellite has booted. In order to do this we want to create a state field with the boot count (maybe call it "pan.bootcount") and add it to the list of fields saved by EEPROM (located in `eeprom_configs.cpp`). A good place to create this field would probably be the `MainControlLoop.cpp` file, in a similar way to how the system memory use field was created.

Once this field is created, we should modify MissionManager to skip its deployment waiting period if it's already booted once previously. We only need to do the deployment wait at the start of the mission, so storing the boot count and using it to drive logic is a good way to prevent wasting time in that state.

madeleinemc:
Add bootcount to EEPROM
Fixes #248

### Summary of changes
- Add "pan.bootcount" to eeprom_saved_fields
- Create bootcount field in FieldCreatorTask
- Increment bootcount in MissionManager constructor

### Testing
- Add test_bootcount in test_mission_manager
- Test that bootcount is in eeprom
- Test that bootcount initializes and increments correctly
- Test that deployment wait is 0 when bootcount is 2

### Telemetry
"pan.bootcount" is now in "eeprom_saved_fields"

### Documentation Evidence
There is associated commenting for every test line and hpp file entry.


src/fsw/FCCode/MissionManager.cpp
```diff
@@ -84,7 +89,7 @@ MissionManager::MissionManager(StateFieldRegistry &registry, unsigned int offset
                   prop_state_t::disabled); // "Starting" transition
     docking_config_cmd_f.set(true);
     enter_docking_cycle_f.set(0);
-    is_deployed_f.set(false);
+    is_deployed_f.set(false); // not necessarily true
```
tanishqaggarwal, line 92: This field is actually redundant now, since `is_deployed = bootcount > 1`. Consider removing the field

test/test_fsw_EEPROM/test_fsw_EEPROM.cpp
```diff
@@ -19,10 +19,11 @@ class TestFixture {
     std::shared_ptr<ReadableStateField<unsigned char>> mission_mode_fp;
     std::shared_ptr<ReadableStateField<bool>> is_deployed_fp;
     std::shared_ptr<ReadableStateField<unsigned char>> sat_designation_fp;
-    std::shared_ptr<ReadableStateField<unsigned int>> control_cycle_count_fp;
+    td::shared_ptr<ReadableStateField<unsigned int>> control_cycle_count_fp;
```
tanishqaggarwal, line 22: Syntax error here

telemetry
```diff
@@ -1590,6 +1591,14 @@
             "type": "lin double vector",
             "writable": false
         },
+        "pan.bootcount": {
+            "bitsize": 7,
+            "flow_id": "undefined",
```
tanishqaggarwal, line 1596: Since we added this field to the telemetry we should consider putting it in a flow so that we don't have to do more work later to organize the flows

telemetry
```diff
@@ -1590,6 +1591,14 @@
             "type": "lin double vector",
             "writable": false
         },
+        "pan.bootcount": {
+            "bitsize": 7,
```
tanishqaggarwal, line 1595: I think we should keep the bitsize 32, because it's possible this value might become large.

test/test_fsw_mission_manager/test_fixture.cpp
```diff
@@ -21,6 +21,10 @@ TestFixture::TestFixture(mission_state_t initial_state) : registry()
 
     docked_fp = registry.create_readable_field<bool>("docksys.docked");
 
+    bootcount_shr_fp = registry.create_readable_field<unsigned int, 1000>("pan.bootcount");
```
tanishqaggarwal, line 24: What does shr stand for?

test/test_fsw_mission_manager/test_mission_manager.cpp
```diff
@@ -233,7 +233,7 @@ void test_dispatch_docking() {
     // Check that mission manager moves to standby
     tf2.check(mission_state_t::standby);
 
-    // Even if a significant amount of time passes, mission managaer should still move to 
+    // Even if a significant amount of time passes, mission manager should still move to
```
tanishqaggarwal, line 236: thanks lol

src/fsw/FCCode/MissionManager.cpp
```diff
@@ -169,11 +174,13 @@ void MissionManager::dispatch_startup()
 {
     set(radio_state_t::disabled);
 
-    // Step 1. Wait for the deployment timer length.
-    if (deployment_wait_elapsed_f.get() < deployment_wait)
-    {
-        deployment_wait_elapsed_f.set(deployment_wait_elapsed_f.get() + 1);
-        return;
+    // Step 1. Wait for the deployment timer length. Skip if bootcount > 1
+    if (bootcount_fp->get() == 1) { //Can a serialized unsigned int be compared in this way?
```
tanishqaggarwal, line 178: Serialization does not affect the value of the field that can be accessed via `get` and `set`. It only affects the way the value is written to a bitstream for downlink/uplink.

telemetry
tanishqaggarwal: I think the "core state fields" flow is a good place to put it. You can modify `src/flow_data.csv` to add the field to this telemetry flow.

tanishqaggarwal (changes requested):

test/test_fsw_mission_manager/test_fixture.cpp
madeleinemc: shared pointer, since there's another pointer below it

tanishqaggarwal: Ah I should've shared this earlier: https://github.com/pathfinder-for-autonomous-navigation/FlightSoftware/wiki/Naming-Convention

madeleinemc: There are two bootcount pointers in the text fixture. One is a pointer and one is a shared pointer, so that's why I named them that way. I will change the shared pointer to _fp instead of _shr_fp, but the other one is a regular (non-shared) pointer to a state field- what should I call it? The other regular pointers are also called _fp

tanishqaggarwal: You should only really need 1 pointer I think, try removing uses of the non-shared pointer since it's secondary

tanishqaggarwal (approved):

src/fsw/FCCode/FieldCreatorTask.hpp
```diff
@@ -29,7 +30,8 @@ class FieldCreatorTask : public ControlTask<void> {
         vel_f("orbit.vel", Serializer<lin::Vector3d>(0,100000,100)),
         pos_baseline_f("orbit.baseline_pos", Serializer<lin::Vector3d>(0,100000,100)),
         vel_baseline_f("orbit.baseline_vel", Serializer<lin::Vector3d>(0,100000,100)),
-        prop_planner_state_f("prop.planner.state", Serializer<unsigned char>(2))
+        prop_planner_state_f("prop.planner.state", Serializer<unsigned char>(2)),
+        bootcount_f("pan.bootcount",Serializer<unsigned int>(4294967295), 1000)
```
athena255, line 34: consider adding this magic number (4294967295) to TRACKED_CONSTANTS

src/fsw/FCCode/MissionManager.cpp
```diff
@@ -43,6 +45,9 @@ MissionManager::MissionManager(StateFieldRegistry &registry, unsigned int offset
     add_writable_field(sat_designation_f);
     add_internal_field(enter_close_approach_ccno_f);
 
+    bootcount_fp = find_readable_field<unsigned int>("pan.bootcount", __FILE__, __LINE__);
```
athena255, line 48: @tanishqaggarwal  
Not sure if this is an issue, but should we account for the case where find_readable_field fails? I believe this method returns nullptr

athena255 (approved):

src/fsw/FCCode/FieldCreatorTask.hpp
madeleinemc: ah yes that's the max value for a 32-bit unsigned int. Is it worth adding to tracked constants or should I just add an inline comment?

athena255: Oh in that case, it's fine. max values for 32-bit is typically represented in hex form 0xffffffff, but I think this is fine

src/fsw/FCCode/MissionManager.cpp
tanishqaggarwal: Why does this method return nullptr?

tanishqaggarwal: If find_readable_field fails then all of flight software stops and won't even get to the first control cycle. We'd be able to detect this on HOOTL pretty much immediately.

src/fsw/FCCode/FieldCreatorTask.hpp
tanishqaggarwal: Might be worth changing to 0xfffffff just to be extra clear.

madeleinemc: Sounds good, will do

Issue #501, opened 2020-09-23
Fix HOOTL slowdown


tanishqaggarwal:
Fix the issues seen in #378. This is becoming more important since we're approaching the need to write mission-level testcases.

Suggestion: use a list instead of TinyDB. It has O(1) insertion time.

Issue #378, opened 2020-04-08
HOOTL slows down over long periods of time


tanishqaggarwal:
See the attached graph of pan.cycle_no, which increments almost logarithmically. Need to investigate this.
<img width="677" alt="Screen Shot 2020-04-05 at 11 49 53 AM" src="https://user-images.githubusercontent.com/5239500/78836233-b1632280-79bf-11ea-85c5-11ec3bc038f2.png">

shihaocao:
Do you have the data points for this? Was thinking about checking it on log log plot

kkrol27:
I'd look for some operations that `O(log n)` or `O(n log n)` where somehow `n` is directly related to number of cycle lmao This is pretty funny.

tanishqaggarwal:
I think it came from here:
https://cornellprod-my.sharepoint.com/personal/saa243_cornell_edu/_layouts/15/onedrive.aspx?id=%2Fpersonal%2Fsaa243%5Fcornell%5Fedu%2FDocuments%2FOAAN%20Team%20Folder%2FSubsystems%2FSoftware%2FPull%20Request%20HITL%20Logs%2F%23359%20%2D%20Docking%20ptest%20checkout%20and%20DockingSystem%2BDockingController%20fixes%2FDockingCheckoutCase%5F20200331%2D155335

kkrol27:
As in the log writing? How exactly is that handled?

tanishqaggarwal:
I think I discovered the issue. TinyDB, the database that backs the incoming telemetry, is slow:

<img width="886" alt="Screen Shot 2020-04-20 at 11 21 27 AM" src="https://user-images.githubusercontent.com/5239500/79769304-f0677100-82f9-11ea-8ff9-35140bf7fb62.png">

This test runs 100,000 iterations of writing 100 values to the database.

Over time, as the database gets large, the insertion operation to the database has time complexity <img src="https://render.githubusercontent.com/render/math?math=O(n)">. Hence, the plot of time per control cycle vs control cycle # should be proportional to <img src="https://render.githubusercontent.com/render/math?math=n^2">, which is exactly the behavior we see.

I'll open a PR to add this stress test to the database, but I think we can consider this issue closed (unless we want to fix this issue somehow.)

PR #512, opened 2020-10-06
Fixing HOOTL slowdown

knk38:
--------------------------

# Fixing HOOTL slowdown

#501 #378 

### Summary of changes
- Changed the implementation of data_consumers.py to use a basic list instead of tinyDB
- Changed plotter.py to support these new changes and be able to plot the data
-Changed the command prompt to use the new list

### Testing
Below is a plot for the pan.cycle_no field over a period of 2.38 hours. It shows a linear vs a logarithmic curve which is an improvement from before.
Screen Shot 2020-10-05 at 12 44 44 AM

I also have a plot for the duration to write 100 values which is pretty constant, especially compared to before.
Screen Shot 2020-09-27 at 5 11 43 PM

### Documentation Evidence
I believe that my inline documentation is sufficient as I didn't create any new methods or drastically change code structure.


tanishqaggarwal (approved):
Grreat :D

Issue #476, opened 2020-05-16
Autonomous Mission Control PTest case


tanishqaggarwal:
Now that #473 is merged in, after #419 is completed we can start doing testing with a PTest case that periodically sends up position and velocity data. This ticket tracks the creation of that testcase + the completion of an end-to-end testcase for the command client and autonomous uplinking.

tanishqaggarwal:
Some ideas on requirements:
- Passes GPS data back and forth between spacecraft
    - Propagates orbits of spacecrafts to high precision, based on last known data
- If leader goes into standby or safehold, ends the mission on the follower
- If there is an extended comms blackout, 
- Ability to switch roles of leader and follower
- Doesn’t do anything if spacecraft roles are not both leader and/or follower
- Notifies operators of any key issues: - One of the spacecraft faulted out to leader or standby
       - Spacecraft are about to dock

Issue #453, opened 2020-05-03
Document ptest testcase writing options


tanishqaggarwal:

tanishqaggarwal:
Closing all documentation tickets because we're sufficiently all onboarded and at this point if cases are messy, who cares

PR #513, opened 2020-10-08
Restructure ptest

tanishqaggarwal:
### Summary of changes
- Made simulation an object of the ptest case rather than the overall ptest run. Enables testcase to control what it's doing with the simulation (critical for #476 )
- Some ptest simplifications here and there

These changes don't make a substantial difference in the everyday way people interact with/write testcases for PTest. But it enables more flexibility with how we use the simulation, which is critical for #476 .

### Testing
Existing Ptest cases are sufficient to catch changes. I added a new "failing" test case to CI to ensure that failures run smoothly.

Here are [HOOTL logs](https://cornellprod-my.sharepoint.com/:f:/g/personal/saa243_cornell_edu/ErQs_TIftzRCpOFP2zBjOaUBjhOxOEuY837hp2WVUk_qQg?e=skFQnW) for demonstrating that the sim-based cases still work correctly (since they can't run in CI.) QuakeFaultHandler is broken for unrelated reasons; this is being tracked in #514.

### Constants
None.

### Telemetry
None.

### Documentation Evidence
Going to make this a part of #453 since what's changing here is how ptest cases are written. The best form of communicating these changes is to actually have people writing new ptest cases with these changes.


ptest/cases/__init__.py
```diff
@@ -2,8 +2,14 @@
 # - tools/alltest.sh
```
tanishqaggarwal, line 2: Restructured this file so that it's explicitly clear which testcases should run in CI and which should not.

ptest/psim/simulation.py
```diff
@@ -216,8 +182,62 @@ def stop(self, data_dir):
...
+    def send_actuations_to_simmed_satellites(self):
+        self.main_state = self.main_state_promise.result()
+        self.main_state['follower'] = self.eng.actuator_command(self.actuator_commands_follower,self.main_state['follower'], nargout=1)
+        self.main_state['leader'] = self.eng.actuator_command(self.actuator_commands_leader,self.main_state['leader'], nargout=1)
+
+    def save_sim_data(self):
+        self.main_state_trajectory.append(
+            json.loads(self.eng.jsonencode(self.main_state, nargout=1)))
+
+class CppSimulation(Simulation):
```
tanishqaggarwal, line 241: @kkrol27 fyi: this is now the entrypoint for C++ based simulation.

ptest/usb_session.py
```diff
@@ -56,9 +56,6 @@ def __init__(self, device_name, uplink_console, port, is_teensy, simulation_run_
         # Open a connection to elasticsearch
```
tanishqaggarwal, line 56: Removing needless functions in this file.

tanishqaggarwal:
Including everyone as a reviewer who regularly interacts with ptest.

ptest/cases/base.py
```diff
@@ -1,43 +1,34 @@
...
+class PTestCase(object):
     """
     Base class for all HITL/HOOTL testcases.
     """
 
-    def __init__(self, data_dir):
+    def __init__(self, is_interactive, random_seed, data_dir):
         self._finished = False
+        self.is_interactive = is_interactive
+        self.random_seed = random_seed
```
athena255, line 19: What is the purpose of self.random_seed?
Does interactive mean that the case does use MATLAB?

ptest/cases/base.py
```diff
@@ -160,11 +175,11 @@ def _setup_case(self):
         self.setup_pre_bootsetup()
 
         # Prevent faults from mucking up the state machine.
-        self.sim.flight_controller.write_state("gomspace.low_batt.suppress", "true")
-        self.sim.flight_controller.write_state("fault_handler.enabled", "false")
-        self.one_day_ccno = self.sim.flight_controller.smart_read("pan.one_day_ccno")
+        self.flight_controller.write_state("gomspace.low_batt.suppress", "true")
+        self.flight_controller.write_state("fault_handler.enabled", "false")
```
athena255, line 179: What does it mean when fault_handler.enabled is true vs if it is false?

ptest/cases/base.py
```diff
@@ -160,11 +175,11 @@ def _setup_case(self):
...
+        self.one_day_ccno = self.flight_controller.smart_read("pan.one_day_ccno")
```
athena255, line 180: What is a smart_read and why do smart_read instead of regular read?

ptest/runsim.py
```diff
@@ -209,7 +194,7 @@ def stop_all(self, reason_for_stop, is_error = True):
             os.close(binary['pty_master_fd'])
             os.close(binary['pty_slave_fd'])
 
-        sys.exit(1 if is_error else 0)
+        sys.exit(1 if (is_error or self.testcase.errored) else 0)
 
 def check_system_dependencies():
```
athena255, line 199: Don't we not require go anymore?

ptest/runsim.py
```diff
@@ -49,19 +48,11 @@ def start(self):
...
-            tb = traceback.format_exc()
-            self.sim.testcase.logger.put(tb)
-            if not self.is_interactive:
-                time.sleep(1) # Allow time for the exception to be handled by the logger.
-                self.sim.testcase.logger.stop()
-                time.sleep(1.5) # Allow time for the logger to stop
-                self.stop_all("Exiting due to testcase failure.")
-        
         if self.is_interactive:
             self.set_up_cmd_prompt()
```
athena255, line 57: Is the cmd_prompt running in its own thread?

ptest/psim/simulation.py
```diff
@@ -114,52 +116,27 @@ def run(self):
...
-            # Step 2. Update dynamics
-            main_state_promise = self.eng.main_state_update(self.main_state, nargout=1, background=True)
-
-            # Step 3. Simulate flight computers
-            # Step 3.1. Use MATLAB simulation as a base
-            self.computer_state_follower, self.actuator_commands_follower = \
-                self.eng.update_FC_state(self.computer_state_follower,self.sensor_readings_follower, nargout=2)
-            self.computer_state_leader, self.actuator_commands_leader = \
-                self.eng.update_FC_state(self.computer_state_leader,self.sensor_readings_leader, nargout=2)
+        while step < num_steps and self.running and not self.testcase.finished:
```
athena255, line 122: This looks nice

ptest/cases/base.py
tanishqaggarwal: PTest still relies on MATLAB but this restructuring makes it possible to use the C++-based psim instead.
self.random_seed is a randomness parameter taken by MATLAB.
The `-ni` option sets self.is_interactive to False, and `-i` option sets self.is_interactive to True. Interactivity enables/disables the PTest command line.

ptest/cases/base.py
tanishqaggarwal: The main fault handler can be completely disabled, so that no fault response occurs on the spacecraft.

ptest/runsim.py
tanishqaggarwal: Good point, I can add a Go check here.

ptest/runsim.py
tanishqaggarwal: Yep. Otherwise the testcase and command line can't run in parallel.

ptest/cases/base.py
tanishqaggarwal: Shihao wrote smart_read which automatically performs type inference on the state field. There's no great convention for using one function over the other

ptest/runsim.py
athena255: I thought Fatima changed it to a python script

tanishqaggarwal: Oh you're right! Haha. Changed.

ptest/psim/simulation.py
```diff
@@ -216,8 +189,55 @@ def stop(self, data_dir):
...
+        self.sensor_readings_leader = self.eng.sensor_reading(self.main_state['leader'],self.main_state['follower'], nargout=1)
+
+    def update_dynamics(self):
+        self.main_state_promise = self.eng.main_state_update(self.main_state, nargout=1, background=True)
+
+    def simulate_flight_computers(self):
+        self.computer_state_follower, self.actuator_commands_follower = \
+            self.eng.update_FC_state(self.computer_state_follower,self.sensor_readings_follower, nargout=2)
+        self.computer_state_leader, self.actuator_commands_leader = \
+            self.eng.update_FC_state(self.computer_state_leader,self.sensor_readings_leader, nargout=2)
```
kkrol27, line 234: This is something that we'll need for MATLAB but not really otherwise. Basically, sims without a second satellite won't fully sim the dynamics of another satellite if they don't need to.

kkrol27 (approved):
Looks pretty good to me. When we integrate cpp psim back in we may need to do a little bit of refactoring (differentiating between single sat and dual sat simulations etc) but I don't think it's worth worrying about that right now.

athena255 (approved):
Looks good

ptest/data_consumers.py
```diff
@@ -1,7 +1,4 @@
-import queue, os, json, threading, time, datetime
-from tinydb import TinyDB
-from tinydb.middlewares import CachingMiddleware
-from tinydb.storages import JSONStorage
```
knk38, line 4: i think tinyDB needs to also be taken out of requirements.txt, i can do it in my pr tho if you want

knk38 (approved):
Looks good, thanks for getting rid of the tinyDB import / requirements for me

tanishqaggarwal:
Going to merge this so that we can check if any existing PRs are impacted by this change. They should not, but after merging I'll check that the existing testcases are not impacted.

PR #515, opened 2020-10-09
Prop fault response ptest

### LINKED ISSUE #502: HOOTL/HITL testing of prop fault handler
We need to write a ptest case for this and run it in HOOTL and HITL.

athena255:
# PropFaultHandler PTEST

The purpose of this ptest case is to check that the mission state correctly goes into standby when a fault is signaled by the PropFaultHandler

Fixes  #502 

### Summary of changes
- Modified MainFaultHandler.cpp and added the PropFaultHandler to it
- Created a PropFaultHandler ptest case

### Testing
Tested only natively

### Constants
N/A

### Telemetry
N/A

### Documentation Evidence
- ptest file is documented
- test assertion functions are documented with the reason why the assert is expected


tanishqaggarwal (changes requested):
Looks good, although I don't think we should merge this until at least some kind of HITL test that takes sensors into account--it would be good to have an end-to-end test of the prop fault handling (so that we see valves actually firing in the appropriate ways at the appropriate times.)

src/fsw/FCCode/MainFaultHandler.cpp
```diff
@@ -28,6 +28,9 @@ void MainFaultHandler::init()
 
     std::vector<Fault*> active_list_2_standby_super_simple_faults {
         find_fault("prop.pressurize_fail.base", __FILE__, __LINE__),
+        find_fault("prop.overpressured.base", __FILE__, __LINE__),
```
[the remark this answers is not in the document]
tanishqaggarwal: Outdated comment

ptest/cases/prop_fault_handler.py
```diff
@@ -0,0 +1,189 @@
...
+        if self.fault_name == "prop.pressurize_fail":
+            self.test_pressurize_fail()
+        elif self.fault_name == "prop.overpressured":
+            self.test_overpressured()
+        elif self.fault_name == "prop.tank2_temp_high":
+            self.test_tank2_temp_high()
+        elif self.fault_name == "prop.tank1_temp_high":
+            self.test_tank1_temp_high()
+        elif self.fault_name == "finished":
+            self.finish()
```
tanishqaggarwal, line 121: (Disclaimer: this is just a personal preference)

This doesn't read naturally to me because the names of the functions `test_*` have nothing to do with the body of the function, which is executing the test and then transitioning to the next test. But I also recognize there's no easy way to appealingly structure this. Maybe something like:
```
def dispatch_test():
   faults_to_test = [list, of, faults]
   next_fault_to_test = {"fault" : "next fault"}

   for fault in faults_to_test:
        if self.fault_name = fault:
            self.test_fault()
            if self.test_stage == "reset": self.fault_name = next_fault_to_test[fault]
```
and then `self.next_test_stage` is some kind of dictionary that indicates the order in which to test.

Btw: does the testcase break if this ordering is changed? If it is, would it be worth restructuring to be less fragile to fault order?

ptest/cases/prop_fault_handler.py
```diff
@@ -0,0 +1,189 @@
...
+        self.ws("fault_handler.enabled", True)
+        self.flight_controller.write_state("dcdc.SpikeDock_cmd", True)
+
+# --------------------------------------------------------------------------------------
+# Helper methods for tests
+# --------------------------------------------------------------------------------------
+
+    def force_fault(self, fault_name):
+        self.ws("{}.suppress".format(fault_name), False)
+        self.ws("{}.override".format(fault_name), True)
```
tanishqaggarwal, line 33: Cycle after forcing fault?

ptest/cases/prop_fault_handler.py
```diff
@@ -0,0 +1,189 @@
...
+            self.check_prop_state("handling_fault", "prop state should be in handling fault since the fault is forced")
+            self.logger.put("Prop now in handling_fault.")
+            # pressurize_fail has different behavior than the other faults. It stays in handling_fault until suppressed by the ground
+            if FAKE_PRESSURE:
+                self.ws("prop.tank2.pressure", 12)
+            if self.fault_name == "prop.pressurize_fail":
+                self.test_stage = "handle_pressurize_fail"
+            else:
+                self.test_stage = "venting"
+            self.cycle()
```
tanishqaggarwal, line 148: I wonder if adding cycles in the force_fault and reset_fault functions will eliminate the need for these random cyclings

ptest/cases/prop_fault_handler.py
```diff
@@ -0,0 +1,189 @@
...
+            self.check_mission_state("leader", "mission state should be in leader")
+            self.logger.put("Overriding the {} fault.".format(self.fault_name))
+            self.force_fault(self.fault_name)
+            self.test_stage = "handling_fault"
+
+        elif self.test_stage == "handling_fault":
+            self.check_prop_state("handling_fault", "prop state should be in handling fault since the fault is forced")
+            self.logger.put("Prop now in handling_fault.")
+            # pressurize_fail has different behavior than the other faults. It stays in handling_fault until suppressed by the ground
+            if FAKE_PRESSURE:
```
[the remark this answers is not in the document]
tanishqaggarwal: If the tank pressure is getting overwritten because of calls to cycle(), consider creating a custom cycle() function for this testcase that applies the fake tank pressure and then cycles.

ptest/cases/prop_fault_handler.py
```diff
@@ -0,0 +1,189 @@
...
+            self.logger.put("Prop now in venting.")
+            self.check_mission_state("standby", "mission state should be in standby when prop is venting")
+            self.logger.put("Satellite now in in standby state.")
+            self.reset_fault(self.fault_name)
+            self.test_stage = "reset"
+
+        elif self.test_stage == "reset":
+            self.logger.put("Resetting satellite state")
+            self.check_prop_state("idle", "prop state should be in idle when fault is suppressed")
+            self.mission_state = "leader"
```
tanishqaggarwal, line 176: Need to cycle after changing the mission state

ptest/cases/prop_fault_handler.py
```diff
@@ -0,0 +1,189 @@
...
+            self.reset_fault(self.fault_name)
```
tanishqaggarwal, line 177: Isn't the fault getting reset a second time here, since it also got reset in line 170? Sorry if I'm reiterating something we've discussed before

ptest/cases/prop_fault_handler.py
tanishqaggarwal: Can the resetting of mission state be put into the reset_fault function?

tanishqaggarwal (commented):
I think I might have some suggestions on how to address the ad-hoc cycling fixes, but not sure if they work.

PR #516, opened 2020-10-11
Hardware stress test + Long Duration Gomspace Test

### LINKED ISSUE #401: Hardware stress test as a ptest case
Spin up all wheels, fire a bunch of valves, turn docking motors...just do a bunch of things and see if the satellite can hold up.

afoarce:
# PR Hardware stress test + Long Duration Gomspace Test

Fixes #401 

### Summary of changes
- Made a PTest checkout case for hardware stress testing (fire valves, turn docking, max speed on wheels)
- Made a PTest checkout case for Gomspace long duration testing (occasionally spin wheels/motor, fire valves)

### Testing

- HOOTL tested
- Needs to be HITL tested

### Documentation Evidence
The test case has in line documentation/logging that explains what it is doing.


tanishqaggarwal (approved):
The test structure looks good! I'd also loop @athena255, @fatimayousuf , and @shihaocao into this PR because it touches prop, Gomspace and ADCS. Maybe you guys don't need to review it thoroughly but please just stay abreast of what this PTest case entails.

ptest/cases/gomspace_long_duration_case.py
```diff
@@ -0,0 +1,89 @@
+from .base import SingleSatOnlyCase, TestCaseFailure
```
tanishqaggarwal, line 1: Shouldn't need testcasefailure

ptest/cases/hardware_stress_test_case.py
```diff
@@ -0,0 +1,63 @@
...
+
+    def prop_valves_setup(self):
+        self.logger.put("Begin Valve Setup")
+        self.ws("dcdc.SpikeDock_cmd", True) 
+        self.sim.flight_controller.write_state(
+            "prop.state", Enums.prop_states["disabled"])
+
+    def docking_spin_motor_setup(self):
+        self.logger.put("Begin Docking Motor Setup")
+        self.ws("dcdc.SpikeDock_cmd", "true") #should this be "true" or True ?
```
tanishqaggarwal, line 28: Should be True because `ws` is a smart function.

ptest/cases/hardware_stress_test_case.py
```diff
@@ -0,0 +1,63 @@
...
+
+    def spin_motors(self, speed):
+        assert(speed in range(-680, 681))
+        self.ws("adcs_cmd.rwa_speed_cmd", [speed, speed, speed])
+
+    def fire_valves(self):
+        val = "0"
+        if self.rs("pan.cycle_no")%2 == 0:
+            val = "1"
```
tanishqaggarwal, line 37: This val might be wrong, the value should be in milliseconds and greater than 3 or so in order for the valves to actually fire. Might be a good idea to set them to their max value of 1000

ptest/cases/gomspace_long_duration_case.py
```diff
@@ -0,0 +1,89 @@
...
+        self.logger.put("Begin Docking Motor Setup")
+        self.ws("dcdc.SpikeDock_cmd", "true") #should this be "true" or True ?
+
+    def spin_motors(self, speed):
+        assert(speed in range(-680, 681))
+        self.ws("adcs_cmd.rwa_speed_cmd", [speed, speed, speed])
+
+    def fire_valves(self):
+        val = "0"
+        if self.rs("pan.cycle_no")%2 == 0:
```
tanishqaggarwal, line 36: Same thing here, re: prop valve firing times

ptest/cases/gomspace_long_duration_case.py
```diff
@@ -0,0 +1,89 @@
+from .base import SingleSatOnlyCase, TestCaseFailure
+from .utils import Enums
+
+class GomspaceLongDurationCheckoutCase(SingleSatOnlyCase):
```
tanishqaggarwal, line 4: Since a lot of the utility functions are shared, consider splitting them out into a shared class and combining these two files. Python supports multiple inheritance and therefore allows "mixing in" functionality from another class.

tanishqaggarwal (changes requested):
Oops I had more comments, for some reason they didn't come through.

ptest/cases/hardware_stress_test_case.py
tanishqaggarwal: Nevermind

ptest/cases/gomspace_long_duration_case.py
```diff
@@ -0,0 +1,89 @@
...
+        self.docking_spin_motor_setup()
+        self.prop_valves_setup()
+        self.adcs_spin_wheels_setup()
+        self.cycle()
+
+    def adcs_spin_wheels_setup(self):
+        self.logger.put("Begin ADCS Motor Setup")
+        self.ws("dcdc.ADCSMotor_cmd", True)
+        self.cycle()
+        self.ws("adcs.state", Enums.adcs_states["point_manual"])
```
shihaocao, line 16: heads up, this will need a change of `point_manual` to `manual` pending newest PR

ptest/cases/gomspace_long_duration_case.py
```diff
@@ -0,0 +1,47 @@
+from .base import SingleSatOnlyCase, TestCaseFailure
+from .utils import Enums
+from .hardware_stress_test_case import ActuateHardwareCase
+
+class GomspaceLongDurationCheckoutCase(ActuateHardwareCase):
+    def setup_post_bootsetup(self):
+        self.print_header("Begin Gomspace Long Duration Checkout Case")
+        self.docking_spin_motor_setup()
+        self.prop_valves_setup()
+        self.adcs_spin_wheels_setup()
```
tanishqaggarwal, line 10: This setup stuff is common to both cases and could be abstracted away into the parent class.

tanishqaggarwal (approved):
Looks good!

PR #517, opened 2020-10-11
Attitude Controller

### LINKED ISSUE #287: Attitude Controller

shihaocao:
# Attitude Controller

Fixes #287 

### Summary of changes
- Removed AttitudeComputer from MCL
- Added AttitudeController (does not need temporary data structures for mtr and rwa_torques)
- Added AttitudeController pio unit test
- Bump psim and lin submodule pointers
- ADCSCommander no longer needs to read pointing vectors, instead reads suggested mtr and rwa_torques
- AttitudeComputer replaced by AttitudeController in FSW, MCL
- ADCSCommander is restructured to hopefully be simpler
- ADCSCommander has upgraded pio unit test
- New differentialation of adcs_state's:
- `point_manual` is used by the flight computer/gs in the case where we set the pointing vectors from the ground
- `manual` is what should be used by ptest cases that directly set the actuators!
### Testing
- Deferring exhaustive attitude controller testing to later intergated ptest + psim testing
- Basic function pio unit test to confirm that AttitudeController won't brick FSW
- pio unit tested the new ADCSCommander

### Constants
N/A

### Telemetry
- AttitudeComputer replaced by AttitudeController in relevant fields
- renamed pointing objective fields
- corrected the proper critical commands in Core ADCS telemetry

Also, take the opportunity to add the telemetry into the appropriate flows in src/flow_data.csv.
- Yes I did the above, it hurt me though lol

### Documentation Evidence
- On the way...


src/.~lock.flow_data.csv#
```diff
@@ -0,0 +1 @@
+,shihao,shihao-T490V2,18.10.2020 02:03,file:///home/shihao/.config/libreoffice/4;
```
kkrol27, line 1: Not sure what this is or if it's supposed to be in here?

kkrol27, line 1: Not sure what this is or if it needs to be in here?

src/fsw/FCCode/AttitudeController.hpp
```diff
@@ -0,0 +1,120 @@
...
+class AttitudeController : public TimedControlTask<void> {
+   public:
+    /**
+     * @brief Construct a new attitude controller.
+     *
+     * @param registry
+     * @param offset
+     */
+    AttitudeController(StateFieldRegistry &registry, unsigned int offset);
+    ~AttitudeController() = default; //is this needed?
```
kkrol27, line 22: There only really a point in specifying this if we mark it as `virtual`. However, given we never deconstruct control tasks it doesn't really matter lmao.
```suggestion
```

src/fsw/FCCode/AttitudeController.hpp
```diff
@@ -0,0 +1,120 @@
...
+
+    /**
+     * @brief Update the attitude controllers suggestion for ADCS actuations.
+     */
+    void execute() override;
+
+   protected:
+    // Inputs from the adcs box monitor
+    ReadableStateField<lin::Vector3f> *const w_wheels_rd_fp;
```
kkrol27, line 31: Don't we need to read in the magnetic field from here as well? For the detumbler that is?

src/fsw/FCCode/AttitudeController.hpp
```diff
@@ -0,0 +1,120 @@
...
+    ReadableStateField<lin::Vector3d> const *const pos_baseline_ecef_fp;
+
+    // Inputs/intermediate pointing objective states
+    ReadableStateField<lin::Vector3f> pointer_vec1_current_f;
+    ReadableStateField<lin::Vector3f> pointer_vec2_current_f;
+    // writable to allow ground override
+    WritableStateField<lin::Vector3f> pointer_vec1_desired_f;
+    WritableStateField<lin::Vector3f> pointer_vec2_desired_f;
+
+    // Internal to this statefield, data containers
```
kkrol27, line 54:
```suggestion
    // Internal to this control task, data containers
```

src/fsw/FCCode/AttitudeController.cpp
```diff
@@ -0,0 +1,236 @@
...
+    pointer_data = gnc::PointingControllerData();
+    pointer_data.primary_desired   = pointer_vec1_desired_f.get();
+    pointer_data.primary_current   = pointer_vec1_current_f.get();
+    pointer_data.secondary_desired = pointer_vec2_desired_f.get();
+    pointer_data.secondary_current = pointer_vec2_current_f.get();
+    pointer_data.w_wheels          = w_wheels_rd_fp->get();
+    pointer_data.w_sat             = w_body_est_fp->get();
+    pointer_data.b                 = b_body_rd_fp->get();
+
+    if (lin::any(!(lin::isfinite(pointer_data.primary_desired) && lin::isfinite(pointer_data.primary_current))))
```
kkrol27, line 216: I know I wrote these initially but reading through them again they're a little hard to follow... This may be cleaner:
```suggestion
    if ( !lin::all(lin::isfinite(pointer_data.primary_desired)) || !lin::all(lin::isfinite(pointer_data.primary_current)))
```
totally up to you though.

src/fsw/FCCode/AttitudeController.cpp
```diff
@@ -0,0 +1,236 @@
...
+        return;
+
+    if (lin::any(!(lin::isfinite(pointer_data.secondary_desired) && lin::isfinite(pointer_data.secondary_current))))
```
kkrol27, line 219: Same here.

src/fsw/FCCode/AttitudeController.cpp
```diff
@@ -0,0 +1,236 @@
...
+        /*
+         * We'll autonomously calculate a pointing objective and then control to
+         * it when in standby or docking.
+         */
+        case adcs_state_t::point_standby:
+        case adcs_state_t::point_docking:
+            default_actuator_commands();
+            default_pointing_objectives();
+            transfer_internal_to_output_vectors();
+            calculate_pointing_objectives();
```
kkrol27, line 79: This is a little confusing here that sometimes we copy the values into the output vectors from the switch statement and other times we leave it up to the controller. Not sure if there is a cleaner way of handling that.

kkrol27 (commented):
All in all looks super clean. I have one concern though about how we're handling the outputs from the control task. I was under the impression the attitude controller would always output actuation suggestions and the `ADCSCommander` would read those in and conditionally copy them into the pre-existing `adcs_cmd.*` fields. That way the controller could signal with NaNs that it doesn't have a suggestion and then the commander would be responsible for actually controlling zero torque or whatever we think is best.

In short I guess what I'm proposing is keep the `adcs_cmd.*` fields completely in the commander and then adding two fields for MTR and wheel commands (`attitude.*`) that the commander would read in.

test/test_fsw_attitude_controller/test_attitude_controller.cpp
```diff
@@ -0,0 +1,297 @@
...
+    TestFixture tf;
+    tf.adcs_state_fp->set(static_cast<unsigned char>(adcs_state_t::point_standby));
+    load_good_data(tf);
+
+    tf.step();
+
+    std::cout << lin::transpose(tf.pointer_vec1_desired_fp->get());
+    std::cout << lin::transpose(tf.pointer_vec2_desired_fp->get());
+
+    // all things related to the pointing objectives should be NaN
```
tanishqaggarwal, line 162: This comment seems false

tanishqaggarwal (approved):
Looks solid.

kkrol27:
@shihaocao has the scope of this PR changed to getting a detumble HOOTL working as well?

shihaocao:
Main changes since last review:

- In order to remove the temporary data structures in controller, I made the controller publish its own suggested commands
- Then ADCSCommander will read these commands, and conditionally forward them to the controller if desired
- I also restuctured ADCSCommander to hopefully be easier to understand

- Updated unit tests as required for both ADCSCommander and AttitudeController

ptest/cases/utils.py
```diff
@@ -227,7 +227,7 @@ def setup_boot(self):
             # For now, set this to 1 since the value doesn't matter (we force the satellite
```
tanishqaggarwal, line 227: Does this stuff work now?

src/fsw/FCCode/ADCSCommander.cpp
```diff
@@ -94,64 +106,65 @@ void ADCSCommander::execute() {
...
-}
 
-void ADCSCommander::dispatch_startup(){
-    // don't apply any commands
-    // ADCSBoxController automatically sets ADCS to adcs::ADCSMode::ADCS_PASSIVE
-    // When MissionManager is in safehold, adcs_state = startup (here)
-}
-void ADCSCommander::dispatch_limited(){
-    // TODO: set to desired mag_moment for pointing strat
+        /** We do nothing here since ptest/ground will write into the adcs_cmd fields
```
tanishqaggarwal, line 116: In my head, the name point_manual implies the _pointing strategy_ is being set manually, not the individual wheel/MTR commands. In that case it should have the same logic as point_standby. I feel like the name "manual" is better

src/fsw/FCCode/ADCSCommander.hpp
```diff
@@ -53,15 +59,14 @@ class ADCSCommander : public TimedControlTask<void> {
...
-    /** Internal specific dispatch call to calculate commands */
-    void dispatch_startup();    
-    void dispatch_limited();    
-    void dispatch_zero_torque();
-    void dispatch_zero_L();     
-    void dispatch_detumble();   
-    void dispatch_manual();     
-    void dispatch_standby();    
-    void dispatch_docking();    
+    template<typename T, size_t N>
```
tanishqaggarwal, line 62: This is hacky, wondering if we can move to some place better.

telemetry
```diff
@@ -1713,6 +1713,22 @@
             "type": "unsigned int",
             "writable": true
         },
+        "pointer.mtr_cmd": {
```
tanishqaggarwal, line 1716: It's weird to me that the maximum torque of the magnetorquers is greater than the maximum torque of the reaction wheels. Is this correct?

src/fsw/FCCode/ADCSCommander.cpp
tanishqaggarwal: If you do change this, please also change the corresponding enums table in `ptest/cases/utils.py`

tanishqaggarwal (approved):

telemetry
tanishqaggarwal: Resolution: the number here is the commanded magnetic moment, not the torque. The torque is found via u x B and B ~ 10^-5, so the torque is lower than the reaction wheel torque.

PR #521, opened 2020-10-18
Adding Magnetorquer Test case

### LINKED ISSUE #424: Write Ptest case for cycling magnetorquers and collecting magnetometer data
To characterize the magnetometer's sensing of MTRs.

Either make this part of the existing "science" ADCS ptest case that does data logging, or make it a separate case. Implementation details TBD.

knk38:
# Adding Magnetorquer Test case

Fixes #424 

### Summary of changes
- Created new test case for cycling magnetorquers and collecting magnetometer data
- Added test case to __init__.py
- I think this has some changes to logger in it too 

Need to confirm that the magnetometers are functioning correctly on the hardware side.

### Testing
I used HITL and HOOTL testing. Some charts for the mtr_cmd and mag1/mag2 vector state fields are below.

This is the graph of the magnetometer readings (which are constant for some reason)
Screen Shot 2020-10-18 at 2 03 59 PM

This is the graph of the magnetorquer command state field which confirms that we are commanding the mtr correctly.
Screen Shot 2020-10-18 at 3 44 44 PM

### Documentation Evidence

- added quite a bit of inline documentation
- each function has a specification under it


ptest/cases/mtr_checkout_case.py
```diff
@@ -0,0 +1,186 @@
+from .base import SingleSatOnlyCase, TestCaseFailure
```
tanishqaggarwal: This file should be removed right?

ptest/cases/torquer_rotate_case.py
```diff
@@ -0,0 +1,153 @@
...
+            waiting and then checking for white noise
+        """
+        #enables MTorquers
+        self.cycle()
+        self.ws( "adcs_cmd.mtr_mode", 1 )# 1 == MTR_ENABLED
+
+        self.log_white_noise()
+
+    def torque_test(self, value):
+        """
```
tanishqaggarwal, line 78: Thanks for these super detailed comments per function.

tanishqaggarwal (approved):
Looks great.

@srutiv might need some additional utilities to be able to do stuff like regressions on the data. Maybe some way to export CSV/Excel data from its JSON description. Could you support this, Kirin?

ptest/cases/mtr_checkout_case.py
knk38: yeah sorry just removed it

shihaocao:
I think we might actually be seeing no interference because adcs_state was never set to manual, also do we need to see wheel interference, as well as mag2 data?

PR #527, opened 2020-10-30
MTR Diagnosis + Hotfix

shihaocao:
# MTR Diagnosis + Hotfix

Supports MTR testing and characterization

As of this PR, the MTR pins indeed do have their voltages set when commanded from flight software.
Tested using manual mode, and negative MTR commands do work too.

### Summary of changes
- Fixes ordering of Enum

### Testing
Tested as necessary on HITL Barebones + Sensors

### Constants
N/A

### Telemetry
N/A

### Documentation Evidence
N/A


tanishqaggarwal (approved):

ptest/cases/utils.py
```diff
@@ -135,8 +135,8 @@ class Enums(object):
     ])
 
     mtr_modes = FSWEnum([
-        "MTR_ENABLED",
-        "MTR_DISABLED"
+        "MTR_DISABLED",
+        "MTR_ENABLED"
```
kkrol27, line 139: Ooof wait so we had this backwards here? As in we thought they were enabled by there were actually disabled?

kkrol27 (commented):
What exactly do you mean negative MTR commands don't work?

ptest/cases/utils.py
shihaocao: ya fat f

PR #529, opened 2020-11-01
Teensy common test failures

### LINKED ISSUE #339: Common Software Unit Tests failing on Teensy

athena255:
#412 Common Software Test Failures Fix

Fixes #339 

### Summary of changes
- GPSTime types changes from unsigned long to unsigned long long because an unsigned long in Teensy is only 4 bytes as oppose to the 8 byte unsigned long long
- Common tests failed on Teensy because Teensy requires that a constant double be defined with an "L", so 3.14159265L instead of 3.14159265. This fixes the "assert double" statements. 

TODO:
- we'd want to check out the SerializerTypes.inl constants, in particular, dqsz, dvcsz, dqcsz because 29, 9, 9 passes, and 40, 10, 10 passes, but 30, 10, 10 does not pass. 

### Testing
Common Software Tests on Teensy36 now all pass

### Constants
- GPSTime.hpp: NANOSECONDS_IN_WEEK changed the type from unsigned long to unsigned long long

### Telemetry
No changes


tanishqaggarwal (approved):
Approving, though could you check to see if the -fsingle_precision flag thing that Kyle was talking about fixes the problem?

PR #531, opened 2020-11-02
509/powerlosscase

18kudlugim:
# Quake Fault Handler/Radio Power Loss PTest Case
Absorbs quake_fault_handler.py PTest case

### Summary
- Understands and tracks exactly when QFH must power cycle, fails if this is done inappropriately 
- Would like someone to look over logic for when we are expecting power cycling
- Will add prop power loss case soon

### Constants
Just a note, for HOOTL, we changed the TRACKED_CONSTANT_SC max_transceive_cycles to:

PAN::one_day_ccno / (24 * 60) > 1 ? PAN::one_day_ccno / (24 * 60) : 1


ptest/cases/power_loss_conditions_case.py
```diff
@@ -0,0 +1,117 @@
...
+            while True:
+                # Check that radio was in wait before cycling.
+                radioWasInWait = self.radio_in_wait()
+                self.cycle()
+                self.diagnostics()
+                self.cycles_since_blackout_start += 1
+                self.check_faulty_powercycle(self.one_day_ccno/3,radioWasInWait)
+
+                if (self.cycles_since_blackout_start >= self.one_day_ccno/3) and self.radio_in_wait() and radioWasInWait:
+                    break
```
tanishqaggarwal, line 111: This block of code is identical to the while-block above, and is a good candidate for breaking out into a function.

tanishqaggarwal (changes requested):
Great job with this! This particular fault handler's testing is an extremely tricky thing to get right and you did this beautifully.

I made a lot of comments here that are arguably just personal style preferences. It's your code and you own it, so if you don't like a suggestion I made, just say no to it

I also think we should figure out some way to test this behavior in HITL. I haven't checked yet if I'm able to get radio communications to the EDU from my house, but I'll check that soon. Assuming that's a go, it would be great if you could start thinking about what modifications are needed to get this working for HITL and write an issue ticket to encapsulate what needs to get done. One important thing to look out for would be all of the unrealistic timing speedups.

tanishqaggarwal:
Closing due to #590

PR #534, opened 2020-11-06
PTest Hot Fixes

shihaocao:
# PTest Hot Fixes
--------------------------

### Fixes:
- Observed behavior: sometimes long lines get cut off when writing command to HITL through ptest. Observed when auto cycle not active. This is likely because when auto cycle is inactive, the debug console loops rapidly often attempting to read a command before it is finished, causing the command to not be completed.
- MTR pins still weren't working, changed some setup in MTorquerCase

### Summary of changes
- Minor changes to MTorquerCase
- debug_console now waits for the `\n` character

### Testing
- HITL tested with ADCSCheckoutCase and MTorquerCase
- Used fsw_teensy36_hitl_leader, and fsw_teensy36_test_mtr envs

### Constants
Uh debug_console constant?

### Telemetry
N/A

### Documentation Evidence
Todo


ptest/cases/torquer_rotate_case.py
```diff
@@ -112,13 +112,12 @@ def mtr_test(self):
             and also for 0.11
         """
         self.print_header( "TORQUE TESTS: " )
-
-        for val in range( 25, 100, 25 ): #start at 0.025 and ramp up by 0.025 until at 0.1
-            val = val / 1000
+        for val in range( 125, 501, 125 ): #start at 0.025 and ramp up by 0.025 until at 0.1
```
knk38, line 115: maybe change the comment or delete it altogether so it makes sense next to the code

knk38 (approved):
yeah looks good, the changes all make sense

ptest/cases/torquer_rotate_case.py
```diff
@@ -112,13 +112,12 @@ def mtr_test(self):
...
+            val = val / 10000
             self.torque_test( val )
 
-        #Test for max value = 0.11
-        self.torque_test( 0.11 )
+        #Test for max value: from adcs constants
+        self.torque_test( 0.056668)
```
tanishqaggarwal, line 120: @kkrol27 could you confirm if this value change is ok?

Shihao, could you add a link to source code here so that its 100% clear where the number is from?

src/common/debug_console.cpp
```diff
@@ -181,8 +181,11 @@ void debug_console::process_commands(const StateFieldRegistry& registry) {
     if (!found_input) return;
     input.copy(buf, sizeof(buf));
 #else
-    for (size_t i = 0; i < SERIAL_BUF_SIZE && Serial.available(); i++) {
-        buf[i] = Serial.read();
+    char lastchar = '\n';
```
tanishqaggarwal, line 184: Can you add some more comments on why this code is structured the way it is?

tanishqaggarwal (changes requested):

ptest/cases/torquer_rotate_case.py
shihaocao: @kkrol27

src/common/debug_console.cpp
```diff
@@ -181,8 +181,13 @@ void debug_console::process_commands(const StateFieldRegistry& registry) {
...
+    for (size_t i = 0; 
+    // looping condition is once there are bytes available, 
+    // keep looping while there are bytes available or the terminating char \n hasn't appeared yet
+    i < SERIAL_BUF_SIZE && (Serial.available() || lastchar != '\n'); i++) {
```
tanishqaggarwal, line 188: I don't like the loop structure here--I personally prefer to let the second condition be a breaking if-statement in the loop body--but it works.

tanishqaggarwal (approved):

src/common/debug_console.cpp
shihaocao: Yeah in hindsight probably could've been a bit more clear

PR #535, opened 2020-11-07
Prop fill env

### LINKED ISSUE #524: Integrate prop filling code into build system

madeleinemc:
# Prop filling environment 

Fixes #524 

### Summary of changes
- Created the prop_fill environment in PlatformIO, which runs the prop filling code with gomspace pings (in src/prep).

After this gets approved, I'll add it to the Pre-Integration Procedures documentation.

### Testing
Builds successfully (pio run -e prop_fill)


src/prep/FillingSequence.ino
```diff
@@ -0,0 +1,269 @@
+
```
tanishqaggarwal, line 1: I would just get rid of this file to reduce confusion

configs/prop_fill_env.ini
```diff
@@ -0,0 +1,8 @@
+; prop_fill_env
+;
+; Environment for prop filling/topoff
+[env:prop_fill]
+extends = teensy36
```
tanishqaggarwal, line 5: I would actually not extend teensy bc I don't know how all the additional flags in that environment affect this file. Could you use the base teensy framework that's specified in the platformIO documentation instead?

tanishqaggarwal (changes requested):
The file compiles for me, I just want to be careful with what options we use for compilation.

madeleinemc:
note that the most recent commit has only been tested on a teensy35 (on modified code) and not a teensy36 (but the code is for a teensy36)

tanishqaggarwal (approved):
🇺🇸 🚀 🇺🇸💥 🇺🇸

PR #544, opened 2020-11-09
Prep For Charging PTest Case

18kudlugim:
Sets gomspace ppt value to 2 (fixed value for charging).


ptest/cases/prep_for_charging_case.py
```diff
@@ -0,0 +1,20 @@
+# Pre-integration procedure for battery charging.
```
tanishqaggarwal, line 1: Can you add this to the CI checks? The scripts live under .github/workflows/hootl.yml

ptest/cases/prep_for_charging_case.py
```diff
@@ -0,0 +1,20 @@
...
+
+  def run_case_singlesat(self):
+    # Set PPT mode to 2 while charging the battery.
+    self.ws("gomspace.pptmode",2)
+    pptmode = self.rs("gomspace.pptmode")
+    
+    self.logger.put("                                 ")
+    self.logger.put("=================================")
+    self.logger.put("=================================")
+    self.logger.put("pptmode was set to: " + str(pptmode))
```
tanishqaggarwal, line 15: Since this is a script that is run by the launch provider it might be a good idea to add something more understandable and external-facing as well, like "Satellite is prepped for charging."

ptest/cases/prep_for_charging_case.py
```diff
@@ -0,0 +1,20 @@
+# Pre-integration procedure for battery charging.
+from .base import SingleSatOnlyCase
+
+
+class PrepForChargingCase(SingleSatOnlyCase):
```
tanishqaggarwal, line 5: This is tiny but can we rename to "PrepForCharging"? Makes it more understandable by the launch provider (they might be confused on why we'd call it a case.)

tanishqaggarwal (changes requested):

ptest/cases/prep_for_charging_case.py
```diff
@@ -0,0 +1,20 @@
...
+
+  def run_case_singlesat(self):
+    # Set PPT mode to 2 while charging the battery.
+    self.ws("gomspace.pptmode",2)
+    pptmode = self.rs("gomspace.pptmode")
```
tanishqaggarwal, line 10: Need to cycle once after setting a state field so that the change sticks.

tanishqaggarwal:
Heads up, because of #547 we might need to make changes here, but this PR should be good for now. Mentioning this so that this PR is linked against that ticket and we remember to make the change.

.github/workflows/hootl.yml
```diff
@@ -41,6 +41,7 @@ jobs:
       run: |
         python -m ptest runsim -c ptest/configs/ci.json -t CICase -ni
         python -m ptest runsim -c ptest/configs/ci.json -t EmptyCase -ni
+        python -m ptest runsim -c ptest/configs/ci.json -t PrepForCharging -ni
```
tanishqaggarwal, line 44: Could you move this into its own section in the file? It doesn't really fit in with these "empty cases" nor any of the other sections in the file. Maybe as part of "pre-integration cases"?

tanishqaggarwal (changes requested):
Quick little change

tanishqaggarwal (approved):
Good 2 go 🔋 🔌

PR #546, opened 2020-11-13
Adding Reset for Flight Test case and adding cycle.start/cycle.auto to EEPROM

knk38:
# Reset for Flight Test case

Fixes # 

### Summary of changes
- Added a new test case that resets ppt_mode, bootcount, and starts autocycling
- Changed cycle.start and cycle.auto in DebugTask.cpp to be EEPROM saveable

### Testing
-I deleted the telemetry file in root and ran the telemetry info generator to ensure that the two new state fields showed up as EEPROM saved
-I also checked that the fields showed up in the eprom.json file
-I ran reset for flight, then the empty case with autocycling and then the reset for flight case afterwards to ensure that the test case didn't error out when already autocycling. 

### Telemetry
cycle.start and cycle.auto are now EEPROM saveable and update every cycle (EEPROM save period of 1).

### Documentation Evidence
Should be sufficient inline documentation for the new functions


ptest/cases/reset_for_flight.py
```diff
@@ -0,0 +1,40 @@
+from .base import SingleSatOnlyCase, TestCaseFailure
+from .utils import Enums
+import math
+
+class ResetforFlight(SingleSatOnlyCase):
+
+  #Sets gomspace ppt mode to MPPT(1) and fails if it is not successfully updated
+  def set_ppt(self):
```
tanishqaggarwal, line 8: I did some digging today on the flight units, and it turns out that we've got a bug with PPT modes: setting the PPT mode like this does nothing if the spacecraft is turned off again after this testcase (which it has to be.) See #547 for more details. Likely, as part of #547 we will have to change to this testcase to write the PPT mode _hardware setting_ via a different state field. But let's leave this in for now.

ptest/cases/reset_for_flight.py
```diff
@@ -0,0 +1,40 @@
...
+    ppt_mode_cmd = int(self.rs("gomspace.pptmode_cmd"))
+    self.ws("gomspace.pptmode_cmd", 1)
+    self.cycle()
+    ppt_mode_updated = int(self.rs("gomspace.pptmode_cmd"))
+    if ppt_mode_cmd == ppt_mode_updated:
+      self.logger.put("Could not update pptmode")
+      self.failed = True
+
+  #sets the bootcount to 0 and confirms it was changed correctly
+  def set_bootcount(self):
```
tanishqaggarwal, line 18: The code for set_ppt and set_bootcount is very similar and is code that could be broken out into its own function, maybe something like "ws_and_verify"?

ptest/cases/reset_for_flight.py
```diff
@@ -0,0 +1,40 @@
...
+    
+  def run_case_singlesat(self):
+    self.print_header("Starting Reset")
+    self.ws( "cycle.auto", False )
+    self.set_ppt()
+    self.set_bootcount()
+
+    #ensure autocycling is happening 
+    #make sure spacecraft is shut down right after this test case so mission doesnt start executing
+    self.ws( "cycle.auto", True )
```
tanishqaggarwal, line 36: Need to cycle after applying this command.

tanishqaggarwal (changes requested):

tanishqaggarwal:
Can you also reset the deployment timer and deployment-related fields?

knk38:
> Can you also reset the deployment timer and deployment-related fields?

Where can i find these fields?

tanishqaggarwal:
MissionManager.cpp

ptest/cases/reset_for_flight.py
knk38: okay i will take a note of this as well

ptest/cases/reset_for_flight.py
knk38: When I cycle after this command, FC throws an error that says I am cycling too quickly or something?
Screen Shot 2020-11-13 at 5 49 45 PM

tanishqaggarwal: This is probably because autocycling is set to true in eeprom.json prior to booting up the testcase. Remove the file whenever you see this error

MCT/server-files/battery-telemetry.js
```diff
@@ -0,0 +1,143 @@
+/*
```
tanishqaggarwal, line 1: Remove this file

tanishqaggarwal (approved):

MCT/server-files/battery-telemetry.js
knk38: omg i deleted it idk why it came back, maybe its still on my remote

PR #548, opened 2020-11-15
Fix \xff on long commands HITL Bug

shihaocao:
# Fix \xff Bug

Fixes #534 

### Summary of changes
- Hotfix3 PR introduces a new bug where 0's dispayed as \xff in the log files will be introduced to the json commands, effectively bricking about 1% of commands written to spacecraft. The longer the command, the more likely this would occur. Discovered from detumble testing with new psim where there are many long (many bytes) commands.
-- The solution is to rewrite the loop looking for bytes to not log to buffer if no bytes are available

Problem:
image

### Testing
HITL tested on FC only HITL

### DIFF
--- a/src/common/debug_console.cpp
+++ b/src/common/debug_console.cpp
@@ -181,13 +181,16 @@ void debug_console::process_commands(const StateFieldRegistry& registry) {
     if (!found_input) return;
     input.copy(buf, sizeof(buf));
 #else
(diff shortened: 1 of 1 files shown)


tanishqaggarwal (approved):
I'm assuming there's been enough thorough testing that this works with existing testcases.

Issue #499, opened 2020-09-22
Telemetry auto-testing


tanishqaggarwal:
We have the infrastructure  we need in order to produce a telemetry auto-tester, which can produce all kinds of values and test them against the existing software implementations.

Now, it's time to put all of this software to the test and create the auto-tester.

PR #550, opened 2020-11-16
Restructure telemetry info generator

tanishqaggarwal:
A part of #499 .

### Summary of changes
Restructured the telemetry info generator to use C++ data structures rather than JSON. This change will make it easier to construct the telemetry auto-testing engine.

### Testing
Since this code reproduces the original telemetry file almost exactly, the change is OK.

### Telemetry
Minor precision changes and a rename of certain values.

### Documentation Evidence
The inline documentation is sufficient to capture the changes in implementation.


shihaocao:
Looks good to me!

shihaocao (approved):

PR #551, opened 2020-11-17
Orbit control task

### LINKED ISSUE #62: Orbit Controller
Creating this issue to track the new integration of the new orbit controller that will eventually be coded up in PSim.

fatimayousuf:
# Orbit Control Task

Fixes #62 

### Summary of changes
- Added the orbit controller

### Testing
I made some unit tests testing different parts of the orbit controller. I think Kyle said we will test the entire behavior on psim. The tests are mostly self explanatory. 

`schedule_valves()` takes in an impulse in the body frame of the spacecraft and gives the time each valve should open by looking for the minimum linear combination of impulses on each of the thrusters. Since there are 4 thrusters and 3 spatial dimensions, it means that the system is overconstrained. I had to manually find the general solution to Ax=b (which took ages to finish). but it works! :D


src/fsw/FCCode/OrbitController.cpp
```diff
@@ -23,12 +23,160 @@ OrbitController::OrbitController(StateFieldRegistry &r, unsigned int offset) :
 }
 
 void OrbitController::init() {
-    prop_planner_state_fp = FIND_READABLE_FIELD(unsigned char, prop.planner.state);
+    prop_cycles_until_firing_fp = FIND_WRITABLE_FIELD(unsigned int, prop.cycles_until_firing);
+    q_body_eci_fp = FIND_READABLE_FIELD(lin::Vector4f, attitude_estimator.q_body_eci);
+    max_pressurizing_cycles_fp = FIND_WRITABLE_FIELD(unsigned int, prop.max_pressurizing_cycles);
+    ctrl_cycles_per_filling_period_fp = FIND_WRITABLE_FIELD(unsigned int, prop.ctrl_cycles_per_filling);
+    ctrl_cycles_per_cooling_period_fp = FIND_WRITABLE_FIELD(unsigned int, prop.ctrl_cycles_per_cooling);
```
tanishqaggarwal, line 30: Do we need the init function or are these state field pointers initializable as part of the control task's constructor? Genuinely asking because I don't remember where the OrbitController is in the control cycle loop, but I feel like it should be possible to eliminate this init() function.

src/fsw/FCCode/OrbitController.cpp
```diff
@@ -23,12 +23,160 @@ OrbitController::OrbitController(StateFieldRegistry &r, unsigned int offset) :
...
+    return actuation.J_ecef;
+}
+
+unsigned int OrbitController::impulse_to_time(double impulse) {
+    double time = 0.024119 * impulse + 7.0092e-05;
+    int time_ms = time * 1000;
+    return time_ms;
+}
+
+void OrbitController::schedule_valves(lin::Vector3d J_body) {
```
tanishqaggarwal, line 134: Could you walk me through the algebra here at our next work session? I know it works, just wondering if there's any scope to simplify things.

src/fsw/FCCode/OrbitController.hpp
```diff
@@ -20,6 +25,43 @@ class OrbitController : public TimedControlTask<void>
     */
     void execute() override;
 
+    /**
+     * Returns the time in seconds when the satellitee reaches its next firing point
+     * This is just an estimation for the sake of prop planning. It's not a very accurate
+     * number.
+     */
```
tanishqaggarwal, line 32: To improve this comment, maybe talk about how exactly it finds this time and why it's an estimation

tanishqaggarwal (commented):
Looks great, code is nice and clean. Tests look solid. Just have a couple of questions and suggestions

Could you also get rid of the ptest/spam file?

src/fsw/FCCode/OrbitController.cpp
fatimayousuf: the orbit controller goes after the attitude estimator and before the prop controller in the main control loop. I'll move the attitude estimator field to the constructor, but I guess we need the init() function for the prop stuff

tanishqaggarwal: Ok let's just keep as-is then.

src/fsw/FCCode/OrbitController.cpp
```diff
@@ -23,12 +23,160 @@ OrbitController::OrbitController(StateFieldRegistry &r, unsigned int offset) :
...
-    sched_valve3_f.set(0);
-    sched_valve4_f.set(0);
+
+    // Collect time, position, velocity, and sun vector data
+    double t = time_fp->get();
+    lin::Vector3d r = pos_fp->get();
+    lin::Vector3d v = vel_fp->get();
+    lin::Vector3d dr = baseline_pos_fp->get();
+    lin::Vector3d dv = baseline_vel_fp->get();
+    lin::Vector3d sun; // Points from the Earth to the sun
```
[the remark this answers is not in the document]
fatimayousuf: uhhh I just used the model that Kyle showed me a while ago. idk if there is a difference between this and the adcs vector. @kkrol27 ?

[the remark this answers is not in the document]
kkrol27: We want the sun vector in ECI actually because we're just worried about where we are in our orbit. Attitude isn't really relevant here.

src/fsw/FCCode/OrbitController.cpp
```diff
@@ -23,12 +23,163 @@ OrbitController::OrbitController(StateFieldRegistry &r, unsigned int offset) :
...
+    gnc::env::sun_vector(t, sun);
+
+    // Calculate the normal vector of the satellite's orbital plane
+    lin::Vector3d orb_plane = lin::cross(r, v);
+    orb_plane = orb_plane / lin::norm(orb_plane);
+
+    // Project the sun vector onto the satellite's orbital plane
+    lin::Vector3d proj_sun = sun - ( lin::dot(sun, orb_plane) * orb_plane );
```
kkrol27, line 52: I think we may want to normalize the projection of the sun vector into the orbital plane here?

src/fsw/FCCode/OrbitController.hpp
```diff
@@ -20,21 +30,62 @@ class OrbitController : public TimedControlTask<void>
...
+    /*
+     * Convert the impulse of a thruster to the time the valve should be open in ms
+     */
+    unsigned int impulse_to_time(double impulse);
+
+    /**
+     * Calculates the time each valve should open to deliver a given impulse. 
+     * The impulse must be in the body frame of the satellite
+     */
+    void schedule_valves(lin::Vector3d J_body);
```
kkrol27, line 56: We can probably make some of these functions that take in vectors take them by constant reference as opposed to by value to prevent copies.

src/fsw/FCCode/OrbitController.cpp
```diff
@@ -23,12 +27,162 @@ OrbitController::OrbitController(StateFieldRegistry &r, unsigned int offset) :
...
+        lin::Vector3d J_eci;
+        gnc::utl::rotate_frame(q_eci_ecef, J_ecef, J_eci);
+
+        // Transform the impulse from eci frame to the body frame of the spacecraft
+        lin::Vector4f q_body_eci = q_body_eci_fp->get();
+        lin::Vector4d q_body_eci_d = { (double)q_body_eci(0), (double)q_body_eci(1), 
+            (double)q_body_eci(2), (double)q_body_eci(3) }; // cast to double
+
+        lin::Vector3d J_body;
+        gnc::utl::rotate_frame(q_body_eci_d, J_eci, J_body);
```
kkrol27, line 91: You should just be able to do something like this with `lin` if you'd like to be less verbose:

    gnc::utl::rotate_frame(lin::cast<double>(q_body_eci).eval(), J_eci, J_body);

The `lin::cast<...>(...)` function will stream an element wise case of the `lin` type and `.eval()` with force the creation of a `lin::Vector<...>` type (in this case a `lin::Vector4d`).

src/fsw/FCCode/OrbitController.cpp
```diff
@@ -23,12 +27,162 @@ OrbitController::OrbitController(StateFieldRegistry &r, unsigned int offset) :
...
+    double time_till_firing = time_till_node(theta, r, v);
+    double time_till_firing_cc = time_till_firing * 1000 / PAN::control_cycle_time;
+
+    // Schedule the valves for firing soon
+    if (time_till_firing_cc <= delta_time && prop_cycles_until_firing_fp->get() == 0) {
+        prop_cycles_until_firing_fp->set(time_till_firing_cc);
+    }
+
+    // Check if the satellite is at a firing point
+    if ( std::find(std::begin(firing_nodes), std::end(firing_nodes), theta) != std::end(firing_nodes) ) {
```
kkrol27, line 71: Sorry I'm not too familiar but are we trying to check if the satellite is exactly at a firing node here? That may be a little dangerous with floating point numbers...

kkrol27 (changes requested):
Overall structure looks great. Biggest outstanding issue is some trailing work for pesky coordinate frame conversion stuff; we can tag up about that though!

src/fsw/FCCode/OrbitController.cpp
kkrol27: nvm

tanishqaggarwal (approved):
Looks good to me!

src/fsw/FCCode/OrbitController.cpp
```diff
@@ -23,12 +27,164 @@ OrbitController::OrbitController(StateFieldRegistry &r, unsigned int offset) :
...
-    sched_valve4_f.set(0);
+
+    // Collect time, position, velocity
+    double t = time_fp->get();
+    lin::Vector3d r = pos_fp->get();
+    lin::Vector3d v = vel_fp->get();
+    lin::Vector3d dr = baseline_pos_fp->get();
+    lin::Vector3d dv = baseline_vel_fp->get();
+
+    // Get the sun vector in ECEF
```
kkrol27, line 46:
```suggestion
```
We can probably just remove this comment or change it to say ECI.

src/fsw/FCCode/OrbitController.cpp
```diff
@@ -23,12 +27,164 @@ OrbitController::OrbitController(StateFieldRegistry &r, unsigned int offset) :
...
+    gnc::env::sun_vector(t, sun_eci);
+
+    lin::Vector4d q_ecef_eci;
+    gnc::env::earth_attitude(t, q_ecef_eci);
+
+    lin::Vector3d sun_ecef;
+    gnc::utl::rotate_frame(q_ecef_eci, sun_eci, sun_ecef);
+
+    // Calculate the normal vector of the satellite's orbital plane
+    lin::Vector3d orb_plane = lin::cross(r, v);
```
kkrol27, line 57: Sorry I didn't make an explicit comment about this before, but this `v` here is in `ecef` and we really need it to be in `ecef0` (a non-rotating version of `ecef` at this instant in time). If the distinction here doesn't make sense lmk and we can talk through it!

You'll need Earth's angular rate which is the same as `ecef`'s angular rate to do this conversion. There's an example of it happening here:

https://github.com/pathfinder-for-autonomous-navigation/psim/blob/167be53e93c3463542de343a7b2ebd84d4e4bb85/src/gnc/orbit_controller.cpp#L119-L138

`this_v_ecef0` is going to be the comparable variable.

src/fsw/FCCode/OrbitController.cpp
```diff
@@ -23,12 +27,164 @@ OrbitController::OrbitController(StateFieldRegistry &r, unsigned int offset) :
...
+    double time_till_firing = time_till_node(theta, t, r, v);
+    double time_till_firing_cc = time_till_firing * 1000 / PAN::control_cycle_time;
+
+    // Schedule the valves for firing soon if the prop system is idle
+    if (time_till_firing_cc <= (prop_min_cycles_needed() + 10) && static_cast<prop_state_t>(prop_state_fp->get()) == prop_state_t::idle) {
+        prop_cycles_until_firing_fp->set(time_till_firing_cc);
+    }
+
+    // Check if the satellite is around a firing point and the prop system is ready to fire
+    if ( time_till_firing_cc < 20 && static_cast<prop_state_t>(prop_state_fp->get()) == prop_state_t::await_firing) {
```
kkrol27, line 79: Seems like we have a two second firing window then. This seems sufficient to me - @tanishqaggarwal just want to draw your attention to this. Should we maybe make the control cycle count window a tracked constant?

src/fsw/FCCode/OrbitController.cpp
```diff
@@ -23,12 +27,164 @@ OrbitController::OrbitController(StateFieldRegistry &r, unsigned int offset) :
...
+    for (size_t i=0; i < 3; i++) {
+        double time_til_node = (firing_nodes[i] - theta) / ang_vel;
+        if (time_til_node > 0 && time_til_node < min_time) {
+            min_time = time_til_node;
+        }
+    }
+
+    // Return the smallest positive time
+    return min_time;
+}
```
kkrol27, line 119: Looks good. The only thing I'd say is we should have this function take in `vel` is `ecef0` so we don't need to do the `w_earth` and `vel - lin::cross(w_earth, pos)` in this function - it'll just be moved outside to fixed one of my previous comments.

src/fsw/FCCode/OrbitController.hpp
```diff
@@ -1,13 +1,24 @@
 #pragma once
 
+#include <algorithm>
```
kkrol27, line 3: I think this header isn't needed anymore? Either way, any headers that are only needed for an implementation is the `.cpp` file should be included there to prevent them being included in every other translation unit that pulls in the orbit controller header.

kkrol27 (changes requested):
Looks really good - only big thing is moving the calculation of velocity in `ecef0` out of that one function and into the main execute function!

kkrol27 (approved):
Looks good to me!

src/fsw/FCCode/OrbitController.cpp
```diff
@@ -23,12 +27,167 @@ OrbitController::OrbitController(StateFieldRegistry &r, unsigned int offset) :
...
+
+    // If the satellite is within a certain delta time/cc from the next firing point, then the 
+    // propulsion system should get ready to fire soon.
+
+    // Get the time until the satellite reaches the next firing node in control cycles
+    double time_till_firing = time_till_node(theta, r, v);
+    double time_till_firing_cc = time_till_firing * 1000 / PAN::control_cycle_time;
+
+    // Schedule the valves for firing soon if the prop system is idle
+    if (time_till_firing_cc <= (prop_min_cycles_needed() + 10) && static_cast<prop_state_t>(prop_state_fp->get()) == prop_state_t::idle) {
```
athena255, line 79: Why do we need prop_min_cycles_needed() + 10?

src/fsw/FCCode/OrbitController.hpp
```diff
@@ -20,21 +29,64 @@ class OrbitController : public TimedControlTask<void>
...
     const ReadableStateField<lin::Vector3d>* const vel_fp;
     const ReadableStateField<lin::Vector3d>* const baseline_pos_fp;
     const ReadableStateField<lin::Vector3d>* const baseline_vel_fp;
 
+    // Converts from ECI to body frame
+    ReadableStateField<lin::Vector4f>* q_body_eci_fp;
+
     // Outputs
-    ReadableStateField<unsigned char>* prop_planner_state_fp;
+    WritableStateField<unsigned int>* prop_cycles_until_firing_fp;
```
athena255, line 86: Does the OrbitController set the schedules in the Prop Controller? It seems like these fields are declared here, and they do not use the PropController schedule fields

athena255 (commented):
Not sure if I'm missing something but it seems like the the PropController fields are not set here

src/fsw/FCCode/OrbitController.cpp
fatimayousuf: I remember you said that the prop system needs some time to get ready to fire. I thought prop_min_cycles_needed() gave that amount of time? So if the satellite is nearly at a firing node, the orbit control task would let the prop system know that it has to fire soon by setting prop_cycles_until_firing_fp.

I guess I'm not totally sure how the prop system works. How is the orbit control task supposed to set the prop schedule again?

src/fsw/FCCode/OrbitController.hpp
fatimayousuf: No this is a pointer to a field declared from the prop controller 
`prop_cycles_until_firing_fp = FIND_WRITABLE_FIELD(unsigned int, prop.cycles_until_firing);`

Are you talking about the sched_valvex_f fields below? They were already there when I started this branch, so I kinda just guessed that those fields were how the prop system would know how long to open the valves.

src/fsw/FCCode/OrbitController.cpp
athena255: Got it, makes sense. The orbit control task should just set prop schedule by writing to the sched_valve* values. The schedule needs to be set when it is in idle. When it detects a schedule, it will start going through the process of getting ready to fire such as pressurizing. The schedule can be changed as long as it is before Await_Firing iirc.

athena255 (approved):

fatimayousuf:
Thank you! @tanishqaggarwal

PR #554, opened 2020-11-19
log2i function fix

18kudlugim:


tanishqaggarwal (changes requested):
Need to fix build failures. It would also be helpful to reduce the number of formatting changes so that in the future, the diff to this file can be easily understood as just a change to log2i

tanishqaggarwal:
I'm going to close this PR due to #582

PR #558, opened 2020-11-26
Plot Psim Statefields

shihaocao:
# Plot Psim Statefields

### Summary of changes
- Psim plotting is now available, it's just like as if it was a statefield from the satellite.
-- usage: `plot truth.leader.attitude.w attitude_estimator.w_body
- In order to avoid name conflicts, quaternion postfixes are now: `.u, .x, .y, .z`
-- changes are in plotter.py

### Testing
Passes HOOTL, HITL not required since FSW not modified


ptest/cases/base.py
```diff
@@ -307,9 +311,37 @@ def print_rs(self, name):
...
         ret = self.sim.mysim[name]
+        
+        if type(ret) in {lin.Vector2, lin.Vector3, lin.Vector4}:
+            ret = list(ret)
+        
+        stripped = str(ret).strip("[]").replace(" ","")+","
+        
+        packet = {}
+        
+        # packet["t"] = int(self.sim.mysim["truth.t.ns"]/1e9/self.sim.dt) # t: number of cycles since sim start
```
tanishqaggarwal, line 328:
```suggestion
```

ptest/cases/base.py
```diff
@@ -307,9 +311,37 @@ def print_rs(self, name):
         ret = self.rs(name)
         self.logger.put(f"{name} is {ret}")
         return ret
-    
-    def print_rs_psim(self, name):
+
+    def rs_psim(self, name):
+        '''
+        Read a psim state field with <name>, log to logger, and return the python value
```
tanishqaggarwal, line 317:
```suggestion
        Read a psim state field with <name>, log to datastore, and return the python value
```

ptest/cases/base.py
```diff
@@ -307,9 +311,37 @@ def print_rs(self, name):
...
+
+        # log to datastore
+        for d in self._devices:
+            d.datastore.put(packet)
+
+        return ret
+
+    def print_rs_psim(self, name):
+        '''
+        Read a psim state field with <name>, log to logger, print to console and return the python value
```
tanishqaggarwal, line 342:
```suggestion
        Read a psim state field with <name>, log to datastore, print to console and return the python value
```

ptest/cases/psim_debug.py
```diff
@@ -40,25 +40,30 @@ def sim_initial_state(self):
...
-        self.print_rs('adcs_monitor.gyr_vec')
-        self.print_rs("attitude_estimator.q_body_eci")
-        self.print_rs("attitude_estimator.w_body")
-        self.print_rs("attitude_estimator.fro_P")
-        self.print_rs("adcs_cmd.mtr_cmd")
-        self.print_rs("adcs_cmd.rwa_torque_cmd")
+        self.rs_psim("truth.leader.attitude.w")
+        self.data_logs()
+        self.data_logs()
+        # self.data_logs()
```
tanishqaggarwal, line 67:
```suggestion
```

ptest/cases/psim_debug.py
```diff
@@ -40,25 +40,30 @@ def sim_initial_state(self):
...
 
         pass
```
tanishqaggarwal, line 69:
```suggestion
```

ptest/cases/base.py
```diff
@@ -359,6 +391,7 @@ class MissionCase(PTestCase):
     def populate_devices(self, devices):
         self.flight_controller_leader = devices["FlightControllerLeader"]
         self.flight_controller_follower = devices["FlightControllerFollower"]
+        self._devices = [self.flight_controller_leader, self.flight_controller_follower]
```
tanishqaggarwal, line 394: Why is `_devices` needed? This might conflict with some of @afoarce's work

tanishqaggarwal (changes requested):

ptest/cases/base.py
```diff
@@ -307,9 +311,37 @@ def print_rs(self, name):
...
         self.logger.put(f"{name} is {ret}")
         return ret
-    
-    def print_rs_psim(self, name):
+
+    def rs_psim(self, name):
+        '''
+        Read a psim state field with <name>, log to logger, and return the python value
+        '''
         ret = self.sim.mysim[name]
```
kkrol27, line 319: This can throw and exception if the field isn't found. Is that the kind of behavior that we want or would we rather use the `.get(...)` function that will return `None` if the field doesn't exist?

ptest/plotter.py
```diff
@@ -32,12 +32,11 @@ def find_timeseries(self, field, dataList = None):
             return None
 
         suffix = None
-        if any(field.endswith(x) for x in ['.w', '.x', '.y', '.z']):
+        if any(field.endswith(x) for x in ['.u', '.x', '.y', '.z']):
```
kkrol27, line 35: I can open a ticket up in PSim to correct this issue (i.e. rename angular rates from `w` to `omega`) if we'd like to go that route. It'd be a somewhat large refactoring but nothing more than an hour I'd say.

kkrol27 (commented):
Looks pretty good on the PSim end of things; just have the one question about using the `[...]` versus `get(...)` and then some of @tanishqaggarwal's concerns.

ptest/plotter.py
tanishqaggarwal: I feel like this is a much clever and less work-intensive solution.

kkrol27: I'm not arguing with that at all haha - just didn't know if we'd eventually want to return to the `(x, y, z, w)` convention for four dimensional vectors eventually or not.

shihaocao: I think this is the preferred method (with `.u`) since in general I'd like to minimize the length of these state field names haha

ptest/cases/base.py
shihaocao: Needed some container to store all devices within the test case object to reference later (in order to be clean). Which work were you referring to?

ptest/cases/base.py
shihaocao: Sure thing, caught it with a slightly nicer error message too

tanishqaggarwal (approved):

ptest/cases/base.py
shihaocao: Renamed to just self.devices

Issue #566, opened 2020-12-06
TLM webservice alternates between printing -1 and the actual value


tanishqaggarwal:
TODO figure out why this is?

tanishqaggarwal:
@shihaocao suggests: what if we change it to -2? Would that fix it?

PR #567, opened 2020-12-06
Autonomous mission manager

### LINKED ISSUE #476: Autonomous Mission Control PTest case

afoarce:
# Autonomous Mission Manager

Fixes #476 

### Summary of changes
- Added PTest case for radio HOOTL/HITL to test telemetry passing between spacecrafts
- Changes made to radio uplink logic resulting from radio HITL testing

### Testing
The case has been HOOTL testing and will be HITL tested soon as radio fixes are finalized.

### Constants
N/A

### Telemetry
N/A

### Documentation Evidence
- In line documentation shows the steps of the test case and radio session, additional documentation not required since it does not change how flight software functions.


ptest/cases/autonomous_mission_manager_case.py
```diff
@@ -0,0 +1,109 @@
...
+            #Pass telemetry between spacecraft 
+
+            #wait for leader's data to come down from Iridium (automatically sent)
+            orbit_data_fields = ["orbit.pos", "orbit.vel"]
+            while(self.leader.read_state("orbit.time") == None): #what does es actually respond with?
+                pass
+
+            downlinked_data_vals_leader = [lin.Vector3(str_to_val(self.leader.read_state(field))) for field in orbit_data_fields]
+            downlinked_data_vals_follower = [lin.Vector3(str_to_val(self.follower.read_state(field))) for field in orbit_data_fields]
+
```
tanishqaggarwal, line 84: Can you add a function here that processes the raw orbital data and computes propagated orbits? We can have the function do nothing for now and add in propagation once it's ready. This way, according to the code specification, the AMC is always sending the latest, propagated orbit.

ptest/cases/autonomous_mission_manager_case.py
```diff
@@ -0,0 +1,109 @@
...
+            baseline_orbit_data_fields = ["orbit.baseline_pos", "orbit.baseline_vel"]
+            baseline_orbit_data_vals_follower = [downlinked_data_vals_leader[i] - downlinked_data_vals_follower[i]
+                                                                             for i in range(len(baseline_orbit_data_fields))]
+            baseline_orbit_data_vals_leader = [-1*val for val in baseline_orbit_data_vals_follower]
+
+            baseline_orbit_data_vals_follower = [val for val in baseline_orbit_data_vals_follower]
+            baseline_orbit_data_vals_leader = [val for val in baseline_orbit_data_vals_leader]
+
+            self.follower.write_multiple_states(baseline_orbit_data_fields, baseline_orbit_data_vals_follower)
+            self.leader.write_multiple_states(baseline_orbit_data_fields, baseline_orbit_data_vals_leader)
```
tanishqaggarwal, line 95: So one problem with continuously calling write_multiple_states in a loop when `leader` or `follower` is a RadioSession is that the uplink timer keeps getting restarted (see [this line](https://github.com/pathfinder-for-autonomous-navigation/FlightSoftware/blob/master/ptest/radio_session.py#L154) in the definition of `write_multiple_states`). So the uplink will actually not get sent.

To prevent this, I think we should modify `uplink_queued` in RadioSession so that it _also_ returns true if there an `uplink.json` file exists. The existence of this file indicates that either the autonomous mission controller or a ground operator has tried to schedule an uplink to be sent up to the spacecraft.

By making this change, the call to `write_multiple_states` made in these two lines would be defeated by the `self.uplink_queued()` [check](https://github.com/pathfinder-for-autonomous-navigation/FlightSoftware/blob/master/ptest/radio_session.py#L140-L141) in `write_multiple_states`, and therefore the uplink timer won't keep getting restarted.

tanishqaggarwal: The above comment I've made applies only to RadioSession, but here's a secondary thought I had: because `write_multiple_states` never fails if it's being used to send data over USB or to a HOOTL FSW instance, but it might fail if it's being used to send data over the radio, we therefore have a difference in the rates with which the spacecraft get information about each other's orbits in HOOTL vs. HITL. 

This might cause unforeseen problems if we end up over-relying on HOOTL testing to verify our orbit-matching algorithms (since an algorithm that constantly gets updates about spacecraft location is probably going to perform better than an algorithm that only gets it once in a while). Down the road (unless you want to implement this now), we might want to some kind of option to this testcase that occasionally stops this call to `write_multiple_states` in HOOTL in order to make the testcase more flight-like.

Thoughts?

ptest/cases/autonomous_mission_manager_case.py
```diff
@@ -0,0 +1,109 @@
...
+
+    def mission_conditions_met(self):
+        #check leader state
+        leader_state = self.leader.read_state("pan.state")
+        if(leader_state == "3"): 
+            self.logger.put("Leader is in standby. Ending mission." )
+            return False
+        elif(leader_state == "10"):
+            self.logger.put("Leader is in state safehold. Ending mission.")
+            return False
```
tanishqaggarwal, line 17: In order to end the mission we should send a command to _both_ spacecraft to go back to standby or safehold. I think we should add some calls to write_state in order to accomplish this.

Note, though, that if we call `write_state` here, but on a previous loop of `run_case` we called write_state to instead update the orbital positions, the call to `write_state` will fail in HITL if you implement my recommendation below. What do you think we should do about this?

ptest/cases/autonomous_mission_manager_case.py
```diff
@@ -0,0 +1,109 @@
...
+
+        #check faulting for follower state -> should this also end mission?
+        follower_state = self.follower.read_state("pan.state")
+        if(follower_state == "3"):
+            self.logger.put("Alert: Follower is in standby.")
+            return True
+        elif(follower_state == "10"):
+            self.logger.put("Alert: Follower is in state safehold.")
+            return True
```
tanishqaggarwal, line 36: Yep, it should. So since the leader and follower have the same behavior, you might want to abstract away this code block and the earlier codeblock for the leader into a function

ptest/cases/autonomous_mission_manager_case.py
```diff
@@ -0,0 +1,109 @@
...
+
+        #check time since last comms
+        leader_time_since_comms = time.time() - self.leader_time_last_comms 
+        if(leader_time_since_comms > self.comms_time_threshold):
+            self.logger.put("Leader is experiencing comms blackout. Ending mission." + str(leader_time_since_comms))
+            return False
+        follower_time_since_comms = time.time() - self.follower_time_last_comms 
+        if(follower_time_since_comms > self.comms_time_threshold):
+            self.logger.put("Follower is experiencing comms blackout. Ending mission.")
+            return False
```
tanishqaggarwal, line 27: This is a repetition of the same code twice, it should be possible to abstract this away into a function.

tanishqaggarwal (changes requested):

ptest/cases/autonomous_mission_manager_case.py
athena255: I think this is simple enough for it to be okay that it is repeated.

ptest/cases/autonomous_mission_manager_case.py
```diff
@@ -0,0 +1,109 @@
...
+    def run_case_fullmission(self):
+
+        if(self.mission_conditions_met()): 
+
+            #Pass telemetry between spacecraft 
+
+            #wait for leader's data to come down from Iridium (automatically sent)
+            orbit_data_fields = ["orbit.pos", "orbit.vel"]
+            while(self.leader.read_state("orbit.time") == None): #what does es actually respond with?
+                pass
```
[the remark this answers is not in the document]
afoarce: Currently I had it as a placeholder, ideally it should wait until data exists in the elasticsearch database but I don't know what the response is for fields that aren't there yet. Do we not need to check for this initial case before data is transmitted?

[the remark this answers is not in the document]
fatimayousuf: I think I get what you're asking. If you try to search for a field that doesn't currently exist in the ElasticSearch database, then it will return `Unable to find field: {field} in index: {index}` (see tlm/webservice.py). I'm pretty sure the index would be statefield_report_[imei number], but idk if someone changed it since I last wrote ground software

ptest/cases/autonomous_mission_manager_case.py
afoarce: Why is it that it would fail? If it does fail, would it be able to succeed on the next iteration?

ptest/cases/autonomous_mission_manager_case.py
afoarce: That's a good point. For now I've updated RadioSession, I think as we do more HOOTL testing of this we can add random wait lengths or randomly decide if data will be written to slow down the connection.

fatimayousuf: Ahhh I'm sorry I didn't even think of that when I was writing radio session. Yeah if you keep writing multiple states repeatedly, then the timer will restart again and again and it won't send the telemetry packets. I think what Tanishq wrote sounds right. Adding that extra condition to `uplink_queued` should take care of the problem?

ptest/cases/autonomous_mission_manager_case.py
```diff
@@ -0,0 +1,112 @@
...
+            #uplink the leader's data to the follower and vice versa
+            baseline_orbit_data_fields = ["orbit.baseline_pos", "orbit.baseline_vel"]
+            baseline_orbit_data_vals_follower = [propagated_data_vals_leader[i] - propagated_data_vals_follower[i] for i in range(len(baseline_orbit_data_fields))]
+            baseline_orbit_data_vals_leader = [-1*val for val in baseline_orbit_data_vals_follower]
+
+            baseline_orbit_data_vals_follower = [val for val in baseline_orbit_data_vals_follower]
+            baseline_orbit_data_vals_leader = [val for val in baseline_orbit_data_vals_leader]
+
+            self.follower.write_multiple_states(baseline_orbit_data_fields, baseline_orbit_data_vals_follower)
+            self.leader.write_multiple_states(baseline_orbit_data_fields, baseline_orbit_data_vals_leader)
```
fatimayousuf, line 108: I think there should maybe be something here that checks if the uplink actually is sent? For example, if the uplink fails to send and `self.follower.write_multiple_states()` returns false, then it looks like that telemetry packet never gets sent to the other satellite. Also, the time of last comms should give the time of the last _successful_ uplink, not the last attempt to uplink

fatimayousuf (changes requested):
It's so cool seeing everything start to come together! This looks awesome! I probably will leave more comments later because I still sometimes have trouble remembering what exactly I did in ground software and radio session lol. but for now, I think this looks nearly done

ptest/cases/autonomous_mission_manager_case.py
```diff
@@ -0,0 +1,112 @@
...
+    def mission_conditions_met(self):
+        #check for operating leader state
+        leader_state_functional = self.state_check(self.leader, "Leader")
+        follower_state_functional = self.state_check(self.follower, "Follower")
+        if not leader_state_functional:
+            if follower_state_functional:
+                self.follower.write_state("pan.state", 3)
+            else:
+                self.follower.write_state("pan_state", 10)
+            return False
```
tanishqaggarwal, line 28: If follower state was standby, then state_check still returns false, so this logic will unnecessarily move the follower state to safehold.

ptest/cases/autonomous_mission_manager_case.py
```diff
@@ -0,0 +1,112 @@
...
+
+        #check time since last comms
+        leader_time_since_comms = time.time() - self.leader_time_last_comms 
+        if(leader_time_since_comms > self.comms_time_threshold):
+            self.logger.put("Leader is experiencing comms blackout. Ending mission." + str(leader_time_since_comms))
+            return False
+        follower_time_since_comms = time.time() - self.follower_time_last_comms 
+        if(follower_time_since_comms > self.comms_time_threshold):
+            self.logger.put("Follower is experiencing comms blackout. Ending mission." + str(follower_time_since_comms))
+            return False
```
tanishqaggarwal, line 43: We should also _actually_ end the mission by writing to the leader/follower mission state.

ptest/cases/autonomous_mission_manager_case.py
```diff
@@ -0,0 +1,112 @@
...
+            if follower_state_functional:
+                self.follower.write_state("pan.state", 3)
+            else:
+                self.follower.write_state("pan_state", 10)
+            return False
+
+        #check faulting state for follower state
+        if not follower_state_functional:
+            self.leader.write_state("pan.state", 3)
+            return False
```
tanishqaggarwal, line 33: This is a piece of logic that we should test on our next autonomous mission manager test. Are there other substantive chunks of logic that you think we should test? Can we make a list

ptest/cases/autonomous_mission_manager_case.py
```diff
@@ -0,0 +1,112 @@
...
+        return True
+    
+    def mission_conditions_met(self):
+        #check for operating leader state
+        leader_state_functional = self.state_check(self.leader, "Leader")
+        follower_state_functional = self.state_check(self.follower, "Follower")
+        if not leader_state_functional:
+            if follower_state_functional:
+                self.follower.write_state("pan.state", 3)
```
tanishqaggarwal, line 25: To make the write successful here, we need to modify the logic in `radio_session` that currently rejects writing states if an `uplink.json` file already exists.

ptest/cases/autonomous_mission_manager_case.py
```diff
@@ -0,0 +1,112 @@
...
+            self.logger.put("Leader is experiencing comms blackout. Ending mission." + str(leader_time_since_comms))
+            return False
+        follower_time_since_comms = time.time() - self.follower_time_last_comms 
+        if(follower_time_since_comms > self.comms_time_threshold):
+            self.logger.put("Follower is experiencing comms blackout. Ending mission." + str(follower_time_since_comms))
+            return False
+
+
+        #TODO notify if close to docking
+        #TODO exit gracefully if mission ended
```
tanishqaggarwal, line 47: For the sake of more carefully tracking all of the work we're doing, should we split these items into separate tickets that are "nice to haves" and reduce the scope of the current issue describing the AMC?

ptest/cases/autonomous_mission_manager_case.py
```diff
@@ -0,0 +1,112 @@
...
+        return float("inf")
+
+    def setup_post_bootsetup_follower(self):
+        self.using_radios = self.radio_follower != None #assumes only using one physical radio
+        self.leader = self.flight_controller_leader
+        self.follower = self.radio_follower if self.using_radios else self.flight_controller_follower
+
+        self.leader_time_last_comms = time.time()
+        self.follower_time_last_comms = time.time()
+        self.comms_time_threshold = 60*5*1000000000 #currently 5 minutes for testing
```
tanishqaggarwal, line 80:
```suggestion
        self.comms_time_threshold = 60*5 #currently 5 minutes for testing
```
time.time() returns a float in [seconds](https://docs.python.org/3/library/time.html#time.time)

tlm/email_processor.py
```diff
@@ -108,6 +109,8 @@ def check_for_email(self):
         mail_ids = data[0]
         id_list = mail_ids.split()
 
+        print(self.confirmation_mtmsn, self.mtmsn, self.send_uplinks)
```
tanishqaggarwal, line 112:
```suggestion
```
Don't want to commit this to master, although we will probably need to have this in be in for a while due to #566

ptest/radio_session.py
```diff
@@ -183,7 +183,8 @@ def uplink_queued(self):
                 'http://'+self.flask_server+':'+str(self.flask_port)+'/search-es',
                     params=payload, headers=headers)
 
-        if tlm_service_active and response.text.lower()=="true":
+        print(tlm_service_active, response.text)
```
tanishqaggarwal, line 186:
```suggestion
```
We resolved the issue that we were trying to get more information about with this print statement; don't need it anymore.

ptest/cases/autonomous_mission_manager_case.py
tanishqaggarwal: Second the last successful comms thing. @fatimayousuf what might be the best way to retrieve the last successful uplink time?

tanishqaggarwal (changes requested):

ptest/cases/autonomous_mission_manager_case.py
kylekrol: Is this something that we want to be done autonomously though?

ptest/cases/autonomous_mission_manager_case.py
afoarce: What would be the best way to update writing states? Should we read and edit an existing json file whenever we need to update a state, or should we change the uplink logic such that radio session maintains a dictionary and only writes the states and fields to a json file when it is time to send the uplink? Or, something else?

ptest/cases/autonomous_mission_manager_case.py
afoarce: We should definitely test it both in terms of a satellite faulting and then the next chunk of logic with regards to comms blackouts, which now should also put the satellites in standby. The changes being made to the uplink logic should also be tested to make sure all of the latest state values are being sent.

ptest/cases/autonomous_mission_manager_case.py
afoarce: To be more consistent with the logic of either satellites faulting which does change states, I've made it so that both satellites go into standby if comms are down.

ptest/cases/autonomous_mission_manager_case.py
```diff
@@ -0,0 +1,115 @@
...
+                self.follower.write_state("pan_state", 10)
+            return False
+
+        #check faulting state for follower state
+        if not follower_state_functional:
+            self.leader.write_state("pan.state", 3)
+            return False
+
+        #check time since last comms
+        leader_time_since_comms = time.time() - self.leader_time_last_comms
```
shihaocao, line 36: This should be last downlink time? Ask how to get those timestamps from elastic search?

ptest/cases/autonomous_mission_manager_case.py
```diff
@@ -0,0 +1,115 @@
...
+        return float("inf")
+
+    def setup_post_bootsetup_follower(self):
+        self.using_radios = self.radio_follower != None #assumes only using one physical radio
+        self.leader = self.flight_controller_leader
+        self.follower = self.radio_follower if self.using_radios else self.flight_controller_follower
+
+        self.leader_time_last_comms = time.time()
+        self.follower_time_last_comms = time.time()
+        self.comms_time_threshold = 60*5 #currently 5 minutes for testing
```
shihaocao, line 84: needs to go after removing boot utility, and should be more easily configurable?

ptest/cases/base.py
```diff
@@ -389,24 +390,30 @@ class MissionCase(PTestCase):
...
     def initial_state_follower(self):
-        raise NotImplementedError
+        return "follower"
 
     @property
     def fast_boot_leader(self):
-        raise NotImplementedError
+        return True
     @property
     def fast_boot_follower(self):
```
shihaocao, line 415: move to AMC

ptest/radio_session.py
```diff
@@ -183,7 +183,7 @@ def uplink_queued(self):
                 'http://'+self.flask_server+':'+str(self.flask_port)+'/search-es',
                     params=payload, headers=headers)
 
-        if tlm_service_active and response.text.lower()=="true":
+        if tlm_service_active and response.text.lower()=="true" and not os.path.exists("uplink.json"):
```
shihaocao, line 186: @tanishqaggarwal @afoarce Why did we add this?

tlm/email_processor.py
```diff
@@ -25,7 +25,7 @@ def __init__(self, radio_keys_config, elasticsearch, downlink_parser_path):
         self.confirmation_mtmsn=-1
 
         #send_uplinks, keeps track of if the ground can send more uplinks. This allows us to make sure we are only sending one uplink at a time.
-        self.send_uplinks=False
+        self.send_uplinks=True
```
shihaocao, line 28: @athena255 @tanishqaggarwal Are these changes all vetted?

ptest/cases/autonomous_mission_manager_case.py
```diff
@@ -0,0 +1,115 @@
...
+
+    def run_case_fullmission(self):
+
+        if(self.mission_conditions_met()):
```
shihaocao, line 88: breakout into subroutine, and make an independent stateless check for whether or not to enable telemetry

shihaocao (changes requested):
fix per call :) bet

ptest/cases/autonomous_mission_manager_case.py
afoarce: Since we really care more about how recent orbit data is, those time stamps should be okay and we could account for longer time to get downlinked to the database by increasing the threshold?

ptest/cases/autonomous_mission_manager_case.py
afoarce: Breaking out the logic makes the code messier/redundant since we have different outcomes from different conditions, so I have instead renamed it to continue_mission to avoid confusion of the function implying statelessness

ptest/cases/autonomous_mission_manager_case.py
shihaocao: Yeah I think this is fine. `orbit.time` is the time of the data which is probably the only relevantly important time stamp

shihaocao (approved):
Looks good! Wish we could have a smart read built into usb_session/radio session haha. Maybe a thing for the next PR lol

ptest/cases/autonomous_mission_manager_case.py
```diff
@@ -0,0 +1,129 @@
...
+
+    def state_check(self, satellite, designaton):
+        satellite_state = satellite.read_state("pan.state")
+        if(satellite_state == str(Enums.mission_states["standby"])):
+            self.logger.put(designaton + " is in standby. Ending mission." )
+            return False
+        elif(satellite_state == str(Enums.mission_states["safehold"])):
+            self.logger.put(designaton + " is in safehold. Ending mission.")
+            return False
+        return True
```
tanishqaggarwal, line 17:
```suggestion
    def state_check(self, satellite, designation):
        satellite_state = satellite.read_state("pan.state")
        if(satellite_state == str(Enums.mission_states["standby"])):
            self.logger.put(designation + " is in standby. Ending mission." )
            return False
        elif(satellite_state == str(Enums.mission_states["safehold"])):
            self.logger.put(designation + " is in safehold. Ending mission.")
            return False
        return True
```

ptest/cases/autonomous_mission_manager_case.py
```diff
@@ -0,0 +1,129 @@
...
+    
+    def continue_mission(self):
+        #check for operating leader state
+        leader_state_functional = self.state_check(self.leader, "Leader")
+        follower_state_functional = self.state_check(self.follower, "Follower")
+        if not leader_state_functional:
+            if follower_state_functional:
+                self.follower.write_state("pan.state", Enums.mission_states["standby"])
+            elif self.follower.read_state("pan.state") != str(Enums.mission_states["standby"]):
+                self.follower.write_state("pan_state", Enums.mission_states["safehold"])
```
tanishqaggarwal, line 27: Not convinced that this logic works. Let's say the follower state is follower. Then the first if statement (line 24) passes, then the follower's state will be forced into standby unnecessarily.

ptest/cases/autonomous_mission_manager_case.py
```diff
@@ -0,0 +1,129 @@
...
+    def fast_boot_follower(self):
+        return True
+        
+
+    def run_case_fullmission(self):
+
+        #setup
+        self.using_radios = self.radio_follower != None #assumes only using one physical radio
+        self.leader = self.flight_controller_leader
+        self.follower = self.radio_follower if self.using_radios else self.flight_controller_follower
```
tanishqaggarwal, line 93:
```suggestion
        self.leader = self.radio_leader if self.radio_leader !=None else self.flight_controller_leader
        self.follower = self.radio_follower if self.using_radios else self.flight_controller_follower
```
The `using_radios` variable isn't used anywhere else in the code as far as I can tell, so you could a) get rid of it, b) increase parity between leader and follower initializations.

ptest/cases/autonomous_mission_manager_case.py
tanishqaggarwal: I initially got confused by the use of `leader_time_since_comms` and `self.leader_time_last_comms` because they sound similar--on first glance the code seemed to keep rewriting self.leader_time_last_comms. Personally, I feel like it would be clearer just to leave this quantity unnamed inside the if statement on the next line.

ptest/cases/autonomous_mission_manager_case.py
tanishqaggarwal: Also are you guys making the AMC satellite-free? In that case you wouldn't need the radio logic right

tlm/email_processor.py
tanishqaggarwal: These changes have been working when we tested a while ago...

ptest/radio_session.py
tanishqaggarwal: Discussed offline, the idea was that we don't allow an uplink to be modified by _PTest_ once one has been formed and not yet sent. (It may still be modified by the mission operators through the UplinkTimer.) This way, there is no race condition between PTest/AMC writing to the uplink and the uplink timer actually sending (and destroying) the uplink.

tanishqaggarwal: Equivalently (this would be smarter) we could just have a class member containing the uplink and we check if its not None

tanishqaggarwal: I think the reason we did it this way though is so that concurrent processes (e.g. uplink timer, etc.) can read/write to the uplink file. Otherwise I think the class member is not visible from another thread (but this may be false)

tanishqaggarwal (commented):
Looks pretty good, just a couple of minor stylistic comments

ptest/cases/autonomous_mission_manager_case.py
afoarce: What would we want to do instead if the leader is not functional?

ptest/cases/autonomous_mission_manager_case.py
afoarce: I think this wouldn't work because both radios are initialized, but we only have one physical radio to test on which is why the leader is set to the flight_controller automatically. We will eventually have no satellites, but we will still need the radio_session connection to read and send uplinks.

ptest/cases/autonomous_mission_manager_case.py
tanishqaggarwal: Oh! I missed the if statement for the leader. Thanks

ptest/cases/autonomous_mission_manager_case.py
tanishqaggarwal: My original thought behind uplinking was that we shouldn't update any existing json file (or struct) via write_state. Otherwise, there will be a race condition between a writer to the json file and the sending logic (which sends the uplink precisely five minutes after it is created.) The only allowable updates after the initial creation of the packet would be from manual commanding, which can update the uplink up until 1 minute before its sending.

The dictionary thing makes sense from a memory perspective. I think I didn't implement it that way because threading made the data inaccessible to some parts of the program. But that assumption might be incorrect. Feel free to change it if it works

ptest/cases/autonomous_mission_manager_case.py
tanishqaggarwal: Sure, ok. So it's eventually on the docket though to make the AMC satellite-free

PR #577, opened 2020-12-10
Gomspace HOOTL Bug

### LINKED ISSUE #491: GomspaceCheckoutCase is broken in HOOTL

fatimayousuf:
# Gomspace HOOTL Bug

Fixes #491 

### Summary of changes
- Make the power cycle period a state field so that the checkout case can easily read it
- Thee checkout case should check that powercycling happens every period. 

### Testing
I ran the checkout case with fsw_native_leader, fsw_native_leader_realtime, and fsw_teensy36_hitl_leader. The power cycling bug did not show up.

### Constants
I got rid of the gomspace period tracked constant, since I turned it into a state field instead.


src/fsw/FCCode/GomspaceController.hpp
```diff
@@ -117,7 +117,8 @@ class GomspaceController : public TimedControlTask<void> {
     #endif
     // If PAN::one_day_ccno is very short due to the SPEEDUP flag, ensure the period is positive
     // to prevent a divide-by-zero error.
```
tanishqaggarwal, line 119: Can you move this comment into GomspaceController.cpp near the computation?

ptest/cases/gomspace_checkout_case.py
```diff
@@ -88,6 +88,8 @@ def run_case_singlesat(self):
                                                                + str(i) + "_cmd", "true"))
```
tanishqaggarwal, line 88: Can you revert back lines 162-164 (adding back in `TestCaseFailure`)?

src/fsw/FCCode/GomspaceController.cpp
```diff
@@ -74,6 +74,9 @@ GomspaceController::GomspaceController(StateFieldRegistry &registry, unsigned in
     heater_sr(),
     heater_f("gomspace.heater", heater_sr),
 
+    period_sr(),
+    period_f("gomspace.period", period_sr),
```
tanishqaggarwal, line 78:
```suggestion
    period_f("gomspace.powercycle_period", period_sr),
```
Just to be more descriptive with the state field name.

ptest/cases/gomspace_checkout_case.py
```diff
@@ -88,6 +88,8 @@ def run_case_singlesat(self):
...
                                   for i in range(1, 7)]
         self.cycle()
+        while int(self.read_state("pan.cycle_no")) % int(self.read_state("gomspace.period")) != 0:
```
tanishqaggarwal, line 91:
```suggestion
        while int(self.read_state("pan.cycle_no")) % int(self.read_state("gomspace.powercycle_period")) != 0:
```

tanishqaggarwal (changes requested):
You might have to recompile telemetry (`pio run -e gsw_telem_info_generator ; .pio/build/gsw_telem_info_generator/program telemetry`)

ptest/cases/gomspace_checkout_case.py
fatimayousuf: Ah sorry I didn't see that

tanishqaggarwal (approved):

PR #581, opened 2020-12-11
Fix log2 function implementation

tanishqaggarwal:
### Summary of changes
- Fix issue with integer log implementation

### Testing
Passes existing unit tests

### Constants and Telemetry
No changes

### Documentation Evidence
Minor implementation detail


tanishqaggarwal:
I think this is failing because the downlink producer (and all the serialization stuff) depends on the log function quite a bit, which is not great...going to close this PR and open an issue ticket instead.

PR #595, opened 2020-12-22
Updates to Serializers

### LINKED ISSUE #552: Ensure angular resolution for vector serializer is sufficient for orbits

tanishqaggarwal:
Fixes a bunch of serializer-related tickets.

### Summary of changes
- Slightly modified the GPS time serializer so that it uses fewer bits
- Updates serialization constants for a lot of fields
- Separated the vector and quaternion serializer implementations.
- Updated the bit_array to_int() implementations, which limited the int representations to 32 bits and thus indirectly limited the precision with which a vector serializer could specify a value.
- Completely rewrote the vector implementation--it serializes r, theta, and phi now instead of the janky Cartesian component scheme that we had going previously. 

Importantly, this last change **also changes the interface for the vector serializer constructor**. The new specification is as follows: if the bounds of the vector are between r_min and r_max, and the desired vector error is dx, pick the bitsize as something greater than log_2((r_max - r_min) / dx). The real bitsize of the vector then ends up being something like 3 times the value you specified, so don't be alarmed if you're specifying one thing in the C++ file and seeing another thing pop up in `telemetry`.
 - Updated the unit tests for serializer to reflect this interface change.

### Testing
Vector unit tests were updated to match the new constructor. Quaternion unit tests still pass. The bit_array tests were modified to reflect the changes in the bit_array interface.

In addition, I verified all of the unique kinds of vectors on the spacecraft fit reasonable bounds:
- [orbit_pos.pdf](https://github.com/pathfinder-for-autonomous-navigation/FlightSoftware/files/5773987/orbit_pos.pdf) Orbital position error is less than 1 mm
- [orbit_vel.pdf](https://github.com/pathfinder-for-autonomous-navigation/FlightSoftware/files/5777561/orbit_vel.pdf) Orbital velocity error is less than 1 mm/s
- [float_quaternions.pdf](https://github.com/pathfinder-for-autonomous-navigation/FlightSoftware/files/5727323/float_quaternions.pdf) Float quaternion error is approximately 1 degree
- [baseline_vel.pdf](https://github.com/pathfinder-for-autonomous-navigation/FlightSoftware/files/5742389/baseline_vel.pdf) Baseline velocity error is less than 1 mm/s.
- [baseline_pos.pdf](https://github.com/pathfinder-for-autonomous-navigation/FlightSoftware/files/5742390/baseline_pos.pdf) Baseline position error is less than 1 mm.
- [unit_vector.pdf](https://github.com/pathfinder-for-autonomous-navigation/FlightSoftware/files/5742392/unit_vector.pdf) Unit vector error is less than 0.1 degrees.

Successfully unit-tested on Teensy as well:
```
(venv) ➜  FlightSoftware git:(serializer_science) ✗ pio test -e teensy36_common --without-testing; sleep 1; pio device monitor
Verbose mode can be enabled via `-v, --verbose` option
Collected 39 items

Processing test_common_serializers in teensy36_common environment
--------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
Building...
Uploading...
Programming..............................................................................................................................................................................................................................................................................................................................................................
============================================================================= [PASSED] Took 8.64 seconds =============================================================================
--- Available filters and text transformations: colorize, debug, default, direct, hexlify, log2file, nocontrol, printable, send_on_enter, time
--- More details at http://bit.ly/pio-monitor-filters
--- Miniterm on /dev/cu.usbmodem45146401  9600,8,N,1 ---
--- Quit: Ctrl+C | Menu: Ctrl+T | Help: Ctrl+T followed by Ctrl+H ---
test/test_common_serializers/test_serializers.cpp:730:test_bool_serializer:PASS
test/test_common_serializers/test_serializers.cpp:731:test_unsigned_int_serializer:PASS
test/test_common_serializers/test_serializers.cpp:732:test_signed_int_serializer:PASS
test/test_common_serializers/test_serializers.cpp:733:test_unsigned_char_serializer:PASS
test/test_common_serializers/test_serializers.cpp:734:test_signed_char_serializer:PASS
test/test_common_serializers/test_serializers.cpp:735:test_float_serializer:PASS
test/test_common_serializers/test_serializers.cpp:736:test_double_serializer:PASS
test/test_common_serializers/test_serializers.cpp:737:test_f_vec_serializer:PASS
test/test_common_serializers/test_serializers.cpp:738:test_d_vec_serializer:PASS
test/test_common_serializers/test_serializers.cpp:739:test_f_quat_serializer:PASS
test/test_common_serializers/test_serializers.cpp:740:test_d_quat_serializer:PASS
test/test_common_serializers/test_serializers.cpp:741:test_gpstime_serializer:PASS

-----------------------
12 Tests 0 Failures 0 Ignored 
OK

--- exit ---
```

### Constants
Removed a bunch of dumbass constants from serializers. The vector and quaternion serializers are much easier to use now.

### Telemetry
Updated the vector serializer bounds. **Please review these carefully.**

### Documentation Evidence
See (pretty good) inline documentation of the vector serializer implementation. I will add more if while reviewing this PR there's a lot of confusion (which I bet there probably will be.)

An additional reference you might want to see is my [M. Eng paper](https://drive.google.com/file/d/1tzBea552j5Nk87NT9BFrWtJsmcb_GXco/view?usp=sharing), which lays out the implementations and testing in English.


src/common/fixed_array.hpp
```diff
@@ -251,6 +256,26 @@ class fixed_array<bool> : public fixed_array_base<bool> {
...
         to_string(str, offset, 0, size());
     }
+
+  private:
+    /**
+     * Checks if the number of bits required to specify the value
+     * fits within the specified type.
+     */
+    template<typename T>
+    bool num_val_bits_ok(unsigned long long val)
```
tanishqaggarwal, line 266:
```suggestion
    static bool num_val_bits_ok(unsigned long long val)
```

src/fsw/FCCode/ADCSBoxMonitor.cpp
```diff
@@ -23,7 +23,7 @@ ADCSBoxMonitor::ADCSBoxMonitor(StateFieldRegistry &registry,
 
     ssa_mode_rd(2),
     ssa_mode_f("adcs_monitor.ssa_mode", ssa_mode_rd),
-    ssa_vec_f("adcs_monitor.ssa_vec", dummy_vec_sr),
+    ssa_vec_f("adcs_monitor.ssa_vec", Serializer<lin::Vector3f>(1-1e-4,1+1e-4,0)),
```
tanishqaggarwal, line 26: I was thinking, shouldn't we be able to serialize the magnetic field vectors using the vector serializer as well? It should theoretically occupy some shell subspace of R^3 since the magnetic field is pretty close to 0.5 G in magnitude all over the Earth. @kylekrol @shihaocao

kylekrol: With respect to ADCS box monitor stuff, we decided to match all of the serializers to the range of sensor readings the ADCS box can output - i.e. the magnetometers and gyroscopes themselves report rectangular readings, the I2C interface uses rectangular serializations, so the state fields themselves use regularized readings. Keeping all the reading rectangular then gives us the best chance of being able to debug with direct ADCS Monitor downlinks.

src/common/SerializerTypes.inl
```diff
@@ -294,531 +294,471 @@ class Serializer<double> : public FloatDoubleSerializer<double> {
...
+        assert(max > 0); assert(max > min);
+        this->_min[0] = min;
+        this->_max[0] = max;
+    }
+
+  public:
+    void serialize(const std::array<T,3>& src) override {
+        lin::Vector<T, 3> normalized_vec {src[0], src[1], src[2]};
+        T magnitude = lin::norm(normalized_vec);
+        normalized_vec = normalized_vec / magnitude;
```
kylekrol, line 436: Also, just a fun aside, you can "view" another array as a `lin` tensor using the views module:

```
#include <lin/core.hpp>
#include <lin/views.hpp>

...

void serialize(const std::array<T,3>& src) override {
  lin::VectorView<T, 3> src_view(src.data());
  auto const normalized_vec = src_view / lin::norm(src_view);
  ...
}
```

kylekrol: You can also assign back to the view which just writes directly to the `std::array` although that wouldn't be helpful here :)

src/common/SerializerTypes.inl
```diff
@@ -294,531 +294,471 @@ class Serializer<double> : public FloatDoubleSerializer<double> {
...
+        phi_serializer.serialize(phi);
+
+        bit_array::iterator serialized_position = this->serialized_val.begin();
+        serialized_position[0] = xsign;
+        serialized_position++;
+        auto copy_bits = [&serialized_position](Serializer<T>& sr) {
+            auto const& bit_array = sr.get_bit_array();
+            std::copy(bit_array.begin(), bit_array.end(), serialized_position);
+            std::advance(serialized_position, bit_array.size());
+        };
```
kylekrol, line 453: Very nice!

src/common/SerializerTypes.inl
```diff
@@ -294,531 +294,471 @@ class Serializer<double> : public FloatDoubleSerializer<double> {
...
+        SerializerBase<std::array<T, 4>>::operator=(other);
+        
+        for(size_t i = 0; i < quaternion_element_serializers.size(); i++) {
+            this->quaternion_element_serializers[i] =
+                std::make_unique<Serializer<T>>(*other.quaternion_element_serializers[i]);
+        }
+        this->max_component = other.max_component;
+
         return *this;
     }
```
kylekrol, line 548: Can this not be defaulted? Seems like the most complex thing we have to deal with is a `std::array<std::shared_ptr<...>, ...>` which should have all the proper copy and move constructors/operators implemented.

src/common/SerializerTypes.inl
```diff
@@ -294,531 +294,471 @@ class Serializer<double> : public FloatDoubleSerializer<double> {
...
-        : VectorSerializer<float, N, SerializerConstants::fqsz,
-                                     SerializerConstants::fvcsz,
-                                     SerializerConstants::fqcsz>(min, max, size)
-    {
-        assert(size > SerializerConstants::min_fvsz);
-        static_assert(N == 3, "An argumented constructor can only be used for a vector.");
-    }
+    Serializer<std::array<float, 3>>(float min, float max, size_t bitsize)
+        : VectorSerializer<float>(min, max, bitsize)
+    {}
```
kylekrol, line 651: This can just be replaced with a `using VectorSerializer<float>::VectorSerializer;` statement I believe. It just pulls in the constructors from the parent class.

src/common/SerializerTypes.inl
```diff
@@ -294,531 +294,471 @@ class Serializer<double> : public FloatDoubleSerializer<double> {
...
-    {
-        static_assert(N == 4, "A default constructor can only be used for a quaternion.");                                        
-    }
+template<>
+class Serializer<std::array<double, 3>> : public VectorSerializer<double> {
+  public:
+    Serializer<std::array<double, 3>>(double min, double max, size_t bitsize)
+        : VectorSerializer<double>(min, max, bitsize)
+    {}
 };
```
kylekrol, line 660: Same comment here about the `using <class>::<constructor>`.

src/fsw/FCCode/ADCSBoxMonitor.cpp
kylekrol, line 26: Also here, just curious what the internals of the serializers do when passing those magnitude constraints to it - as in, how many bits are used for r, theta, and phi?

test/test_common_serializers/test_serializers.cpp
```diff
@@ -500,47 +500,39 @@ void test_vec_serializer() {
...
+        vector_t result;
         downlink_deserializer->deserialize(&result);
-
-        // normalize the vectors for angle calculation only
-        normalize<T, 3>(vec);
-        normalize<T, 3>(result);
-        T angle = angle_between(vec, result);
-
-        T mag_err = std::abs(magnitude_of(result)/magnitude_of(vec) - 1.0) ;
+        T mag_err = lin::norm(to_linvec(vec) - to_linvec(result));
```
kylekrol, line 530: Code like this may be too pervasive among existing unit tests, but using vector views here could prevent extra copies:
```
T mag_err = lin::norm(lin::VectorView<T, 3>(vec.data) - lin::VectorView<T, 3>(result.data));
```

kylekrol (commented):
All in all, looks pretty good. Just a couple suggestions to cut down on some unnecessary code and a few questions!

src/fsw/FCCode/ADCSBoxMonitor.cpp
tanishqaggarwal: Ah right, thanks for reminding me.

src/common/SerializerTypes.inl
```diff
@@ -3,12 +3,13 @@
 #include <cstdlib>
 #include <cassert>
 #include <lin.hpp>
+#include <iostream>
```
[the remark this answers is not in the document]
tanishqaggarwal: Oops thanks

src/common/SerializerTypes.inl
```diff
@@ -294,531 +294,471 @@ class Serializer<double> : public FloatDoubleSerializer<double> {
...
+    mutable Serializer<T> phi_serializer; // Using physics coordinates: this is xy angle
+
+    /*
+     * Let precision equal p. The meaning of the number is that
+     * the vector should be specified to within the Euclidean
+     * distance given by dx = (max - min) / 2^p.
+     * 
+     * We can approximate
+     * dx = sqrt((r dtheta)^2 + (r dphi)^2 + dr^2)
+     *    = max (dtheta + dphi) + dr
```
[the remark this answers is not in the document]
tanishqaggarwal: Oh shoot you're right, whoops. But yes the logic remains the same because this is still strictly less than `sqrt((r dtheta)^2 + (r dphi)^2 + dr^2)`

src/common/SerializerTypes.inl
tanishqaggarwal: True. This is a holdover from an existing, more complicated serializer.

src/fsw/FCCode/ADCSBoxMonitor.cpp
tanishqaggarwal: You can check via the `total_size` function in VectorSerializer: 2 bits are used for r, 14 bits for theta and phi serializers. So the total size is 31 bits.

src/common/SerializerTypes.inl
tanishqaggarwal: This does not seem to work because src is marked `const`

src/common/SerializerTypes.inl
```diff
@@ -3,12 +3,13 @@
...
 #include <cassert>
 #include <lin.hpp>
+#include <iostream>
 #include "GPSTime.hpp"
 #include "Serializer.hpp"
 #include "types.hpp"
 #include "constant_tracker.hpp"
 
-#include <lin.hpp> // for norm
+#include <lin.hpp>
```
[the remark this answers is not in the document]
tanishqaggarwal: Can't if we want to use vectorview.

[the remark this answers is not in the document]
tanishqaggarwal: The use of `zeroes` also prevents using just `core`

src/fsw/FCCode/FieldCreatorTask.hpp
```diff
@@ -25,10 +25,10 @@ class FieldCreatorTask : public ControlTask<void> {
       FieldCreatorTask(StateFieldRegistry& r) : 
         ControlTask<void>(r),
         time_f("orbit.time", Serializer<double>(0.0, 18'446'744'073'709'551'616.0, 64)),
-        pos_f("orbit.pos", Serializer<lin::Vector3d>(6771000,6971000, 28 + SerializerConstants::min_dvsz)),
-        vel_f("orbit.vel", Serializer<lin::Vector3d>(7500, 7700, 18 + SerializerConstants::min_dvsz)),
-        pos_baseline_f("orbit.baseline_pos", Serializer<lin::Vector3d>(0,2000, 21 + SerializerConstants::min_dvsz)),
-        vel_baseline_f("orbit.baseline_vel", Serializer<lin::Vector3d>(0,11,11 + SerializerConstants::min_dvsz)),
+        pos_f("orbit.pos", Serializer<lin::Vector3d>(6841000, 6901000, 26)),
```
tanishqaggarwal, line 28: 500 +/- 30 km. The orbital velocities are for values within this range.

shihaocao: This is dumb but, do we want this to be configurable? Or larger? Like we could get inserted into the wrong target orbit... Or if its too elliptical etc

src/common/Serializer.hpp
```diff
@@ -62,18 +53,13 @@ class SerializerBase : public SerializerType {
...
-     * serializer for a vector serializer), return from the constructor and do not resize the
-     * serialized bit array.
      */
     SerializerBase(T min, T max, size_t compressed_size, size_t _strlength) :
         strlength(_strlength),
         _min(min),
         _max(max),
         serialized_val()
     {
-        if (static_cast<signed int>(compressed_size) < 0) return;
```
tanishqaggarwal, line 76: This condition is removed because now it realistically cannot happen.

src/fsw/FCCode/FieldCreatorTask.hpp
tanishqaggarwal: We can't make it configurable for sure bc that's not how serializers work, but it would be good to cross-check this range against the launch provider's Interface Control Document (ICD), if that's possible. I don't know if we have precise numbers from them yet though

As a general note btw: we have to refer to the launch provider publicly as "the launch provider" due to FOUO reasons

src/common/SerializerTypes.inl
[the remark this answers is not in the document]
kylekrol: Ah gotcha. Not that it really matters that much but if you wanted zeros you would pull in the generators module with `#include <lin/generators.hpp>`.

src/common/SerializerTypes.inl
kylekrol: Yep you're totally right and that's what https://github.com/pathfinder-for-autonomous-navigation/lin/pull/20 is all about. Not worth waiting on that change in my opinion unless you want to.

src/fsw/FCCode/FieldCreatorTask.hpp
kylekrol: Just want to note here than while altitude will decrease as our orbit decays velocities are going to increase!

tanishqaggarwal: Agreed. I'm going to update this to 400 - 550 km and update the verifications appropriately, just so that we're extra generous.

src/common/SerializerTypes.inl
```diff
@@ -294,531 +291,448 @@ class Serializer<double> : public FloatDoubleSerializer<double> {
...
- *   the largest component in the unit vector, and then we serialize the two smallest components
- *   in the bounds +/- sqrt(2) and with bitsize 9.
- * 
- * For quaternions:
- * - Since the quaternion has a magnitude of size 1, we only need the three smallest components
- *   to specify the direction of the unit vector. We allow two bits to store the index of
- *   the largest component in the quaternion, and then we serialize the three smallest components
- *   in the bounds +/- sqrt(2)/2 and with bitsize 9.
- * 
+ * Common functions used by vector serializers.
```
tanishqaggarwal, line 294:
```suggestion
 * Common functions used by vector and quaternion serializers.
```

src/common/SerializerTypes.inl
```diff
@@ -294,531 +291,448 @@ class Serializer<double> : public FloatDoubleSerializer<double> {
...
+    mutable Serializer<T> theta_serializer; // Using physics coordinates: this is angle w.r.t. z axis
+    mutable Serializer<T> phi_serializer; // Using physics coordinates: this is xy angle
+
+    /*
+     * Let precision equal p. The meaning of the number is that
+     * the vector should be specified to within the Euclidean
+     * distance given by dx = (max - min) / 2^p.
+     * 
+     * We can approximate
+     * dx = sqrt((r dtheta)^2 + (r cos(theta) dphi)^2 + dr^2)
```
tanishqaggarwal, line 379:
```suggestion
     * dx = sqrt((r dtheta)^2 + (r sin(theta) dphi)^2 + dr^2)
```

src/common/SerializerTypes.inl
```diff
@@ -294,531 +291,448 @@ class Serializer<double> : public FloatDoubleSerializer<double> {
...
+            auto const& bit_array = sr.get_bit_array();
+            std::copy(bit_array.begin(), bit_array.end(), serialized_position);
+            std::advance(serialized_position, bit_array.size());
+        };
+        copy_bits(magnitude_serializer);
+        copy_bits(theta_serializer);
+        copy_bits(phi_serializer);
+    }
+
+    bool deserialize(const char* val, std::array<T, 3>* dest) override {
```
tanishqaggarwal, line 460: This should really be a constant function.

tanishqaggarwal: Maybe rename this to string parse.

tanishqaggarwal: I realize that this function does not call "serialize", while the quaternion version does call serialize. I think it makes sense to call serialize here too, because whenever we use this function, we should also update the bit array.

src/common/SerializerTypes.inl
```diff
@@ -294,531 +291,448 @@ class Serializer<double> : public FloatDoubleSerializer<double> {
...
-        // otherwise, when N == 3, src_norm is not normalized
-        if(N == 4) {
-            normd = {src[0], src[1], src[2], src[3]};
-        
-            normd = normd / lin::norm(normd);
-            
-            for(unsigned int i = 0; i<N; i++){
-                src_norm[i] = normd(i);
-            }
+    static constexpr size_t quat_component_sz() { return 10; }
```
tanishqaggarwal, line 520: Add a docstring.

src/fsw/FCCode/FieldCreatorTask.hpp
```diff
@@ -25,10 +25,10 @@ class FieldCreatorTask : public ControlTask<void> {
       FieldCreatorTask(StateFieldRegistry& r) : 
         ControlTask<void>(r),
         time_f("orbit.time", Serializer<double>(0.0, 18'446'744'073'709'551'616.0, 64)),
-        pos_f("orbit.pos", Serializer<lin::Vector3d>(6771000,6971000, 28 + SerializerConstants::min_dvsz)),
-        vel_f("orbit.vel", Serializer<lin::Vector3d>(7500, 7700, 18 + SerializerConstants::min_dvsz)),
-        pos_baseline_f("orbit.baseline_pos", Serializer<lin::Vector3d>(0,2000, 21 + SerializerConstants::min_dvsz)),
-        vel_baseline_f("orbit.baseline_vel", Serializer<lin::Vector3d>(0,11,11 + SerializerConstants::min_dvsz)),
+        pos_f("orbit.pos", Serializer<lin::Vector3d>(6771000, 6921000, 28)),
```
shihaocao, line 28: Looks like we can serialize orbits from 400 km up to 550 km

src/fsw/FCCode/FieldCreatorTask.hpp
```diff
@@ -25,10 +25,10 @@ class FieldCreatorTask : public ControlTask<void> {
...
+        vel_f("orbit.vel", Serializer<lin::Vector3d>(7570, 7685, 19)),
```
shihaocao, line 29: Using a worst case orbit where our apogee is 550 km and our perigee is 400 km, our orbital speeds can range from 6.573 km/s to 8.857 km/s.

What I just mentioned above might be way too conservative, but what are we shooting for/from here?

test/test_common_serializers/test_serializers.cpp
```diff
@@ -500,47 +422,39 @@ void test_vec_serializer() {
...
         // Generate random vector.
-        const T x = rand() / T(RAND_MAX) * 2;
-        const T t = rand() / T(RAND_MAX) * (2 * 3.14159265);
-        const T y = cos(t) * std::sqrt(4 - x*x);
-        const T z = sin(t) * std::sqrt(4 - x*x);
+        const T r = 0.1 + 1.5 * rand() / T(RAND_MAX);
+        const T t = VectorSerializer<T>::pi * rand() / T(RAND_MAX);
+        const T p = VectorSerializer<T>::pi * rand() / T(RAND_MAX);
+        const T x = r * sin(t) * cos(p);
+        const T y = r * sin(t) * sin(p);
```
shihaocao, line 444: Isn't y always positive here?

test/test_common_serializers/test_serializers.cpp
```diff
@@ -500,47 +422,39 @@ void test_vec_serializer() {
...
-        vector_t result;
-
-        // rand() returns the same thing every time
 
         // Generate random vector.
-        const T x = rand() / T(RAND_MAX) * 2;
-        const T t = rand() / T(RAND_MAX) * (2 * 3.14159265);
-        const T y = cos(t) * std::sqrt(4 - x*x);
-        const T z = sin(t) * std::sqrt(4 - x*x);
+        const T r = 0.1 + 1.5 * rand() / T(RAND_MAX);
```
shihaocao, line 440: R's magnitude seems to vary between 0.1 and 1.6, the magnitude of 2 comment earlier doesn't seem to make sense anymore

test/test_common_serializers/test_serializers.cpp
```diff
@@ -549,16 +463,10 @@ void test_vec_serializer() {
...
-        TEST_ASSERT_FLOAT_WITHIN_MESSAGE(0.001, 0, mag_err, err_str);
+            result_arr[2], mag_err);
 
-        // assert angle has error < 0.5 degrees
-        TEST_ASSERT_FLOAT_WITHIN_MESSAGE(0.5, 0, angle, err_str);
-
-        // Note no need to check if there are any sign flips,
-        // angle_err < 0.5 deg if no sign flips, or if error is small (across an interval)
+        // assert less than .1% magnitude of error vector
+        TEST_ASSERT_FLOAT_WITHIN_MESSAGE(0.002, 0, mag_err, err_str);
```
shihaocao, line 469: Is 0.002 still 0.1% error if the mangitude of the vector can be between 0.1 and 1.6?

test/test_common_serializers/test_serializers.cpp
```diff
@@ -619,30 +527,26 @@ void test_quat_serializer() {
...
 
+        quat_t result;
         downlink_deserializer->deserialize(&result);
+        lin::VectorView<T, 4> result_lin(result.data());
+        result_lin = result_lin / lin::norm(result_lin);
 
-        // normalize the result (though it should definitely be normalized, or atleast very close)
-        normalize<T, 4>(result);
-        T angle_err = angle_between<T, 4>(quat, result);
+        T angle_err = std::acos(std::abs(lin::dot(result_lin, quat_lin))) / 2;
```
shihaocao, line 549: Is there a source for this equation? It seems to be different that the one we used before in this unit test for quaternions

src/common/fixed_array.hpp
```diff
@@ -169,19 +169,10 @@ class fixed_array<bool> : public fixed_array_base<bool> {
...
-    bool set_int(unsigned int val) {
-        size_t val_num_bits = 32;
-        for (size_t i = 0; i < 32; i++) {
-            if (pow(2, i) > val) {
-                val_num_bits = i;
-                break;
-            }
-        }
-        if (val_num_bits > size()) return false;
+    bool set_ullong(unsigned long long val) {
```
shihaocao, line 172: This seems to be the only set method that is called, but we never seem to do anything with the boolean return. Is checking if the value fits or not still necessary? Or were we supposed to do something with it?

src/common/fixed_array.hpp
shihaocao, line 266: So if we ever set some value to 0, this method always returns true. Is this desired?

src/common/SerializerTypes.inl
```diff
@@ -294,531 +291,454 @@ class Serializer<double> : public FloatDoubleSerializer<double> {
...
+    }
+};
+
+template<typename T>
+class Serializer<lin::Vector<T, 4>> : public LinVectorSerializer<T, 4>
+{
+  public:
+    Serializer() : LinVectorSerializer<T, 4>()
+    {
+        this->_arr_sr = std::make_shared<Serializer<std::array<T,4>>>();
```
shihaocao, line 729: do we not need a `set_telemetry_info()` below this?

shihaocao (commented):
Heyo sorry for mega late review, finally got through everything. Got a few questions

src/fsw/FCCode/FieldCreatorTask.hpp
tanishqaggarwal: yep I changed the bounds to 400-550 km so that we're more generous

src/fsw/FCCode/FieldCreatorTask.hpp
tanishqaggarwal: Hm...I'm using circular orbit numbers:
https://www.wolframalpha.com/input/?i=orbital+velocity+at+550+km+altitude : 7.589 km/s
https://www.wolframalpha.com/input/?i=orbital+velocity+at+400+km+altitude : 7.673 km/s

Where did you get those numbers?

test/test_common_serializers/test_serializers.cpp
tanishqaggarwal: Good catch, p should be 2 pi

test/test_common_serializers/test_serializers.cpp
tanishqaggarwal: For some _weird_ reason if I set it to like 0.1 + 1.9 * rand, the magnitude value is sometimes in excess of 2. Not sure why but I'll investigate further (I forgot to; thanks for reminding me)

test/test_common_serializers/test_serializers.cpp
tanishqaggarwal: See above response; will investigate.

test/test_common_serializers/test_serializers.cpp
tanishqaggarwal: Oh shit I divided by two instead of multiplying by 2. But it's otherwise the same I think?

tanishqaggarwal: If N == 4, the original code
```
T inner_product = lin::dot(a, b);		

T angle;		
if(N == 4) {		
   // to account for quat could be flipped		
   inner_product = std::abs(inner_product);		
   angle = std::acos(inner_product)*2.0;		
}		
else		
   angle = std::acos(inner_product);		

angle = angle * 360.0 / (2 * 3.14159265);		

return angle;
```

simplifies to `T angle_err = std::acos(std::abs(lin::dot(result_lin, quat_lin))) * 2;`

src/common/SerializerTypes.inl
tanishqaggarwal: We don't. The purpose of `set_telemetry_info()` is to encode the minimum and maximum bound in a way that TelemetryInfoGenerator can process. Unit quaternions have known bounds, so these bounds don't need to be stored into `telemetry`.

src/common/fixed_array.hpp
tanishqaggarwal: The thing is, `unsigned long long` is the largest possible type in C++. So this method can't really fail, which is why checking the value is not really necessary.

The boolean interface is a bit extraneous but it doesn't really harm anything.

src/common/fixed_array.hpp
tanishqaggarwal: Yes because the value 0 can be fit within any number of bits

src/fsw/FCCode/FieldCreatorTask.hpp
kylekrol: Hmm I'd say that it's very unlikely we stray far from a circular orbit simply given how drag works (largest effect at perigee). That being said, bumping up the upper bound a little bit couldn't hurt.

tanishqaggarwal: @kylekrol 400 km - 550 km seems pretty generous already...you sure we need to bump it?

kylekrol: I guess I'm just thinking that it'd suck if our mission was over at launch because our orbital insertion was incorrect. That being said our FCC license is only good for a six month period so 400 km is probably very low.

kylekrol: Is our orbital velocity being specified in km/s there? Shouldn't we be looking at m/s?

tanishqaggarwal: This is definitely meters per second...

src/common/SerializerTypes.inl
```diff
@@ -294,531 +292,454 @@ class Serializer<double> : public FloatDoubleSerializer<double> {
...
-            0, 0),
-        _arr_sr(min, max, size)
-    {
-        // Store min and max information for access by the telemetry info generator
-        this->_min(0) = min;
-        this->_max(0) = max;
-        static_assert(N == 3, "An argumented constructor can only be used for a vector.");
-    }
+  protected:
+    std::shared_ptr<Serializer<std::array<T, N>>> _arr_sr;
```
kylekrol, line 677: Is there a reason for this to be a `shared_pointer` rather than a normal class member?

kylekrol: Ah scratch that I see why. We could add variadic template constructor args to the class but I don't really think it's worth it.

kylekrol (approved):
Looks pretty good to me.

src/common/SerializerTypes.inl
```diff
@@ -226,7 +223,7 @@ class FloatDoubleSerializer : public SerializerBase<T> {
 
   public:
     void serialize(const T& src) override {
-        const unsigned int num_intervals = (0b1 << this->serialized_val.size()) - 1;
+        const unsigned long long num_intervals = std::pow(2, this->serialized_val.size()) - 1;
```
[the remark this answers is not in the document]
tanishqaggarwal: Exactly, there's an overflow issue.

src/common/SerializerTypes.inl
```diff
@@ -294,531 +292,454 @@ class Serializer<double> : public FloatDoubleSerializer<double> {
...
-        static_assert(N == 4, "Default constructor may only be used for quaternions.");
-        construct_vector_serializer(0, 1, quat_sz);
+  public:
+    void serialize(const std::array<T,3>& src) override {
+        lin::Vector<T, 3> normalized_vec {src[0], src[1], src[2]};
+        T magnitude = lin::norm(normalized_vec);
+        normalized_vec = normalized_vec / magnitude;
+
+        T theta = std::acos(normalized_vec(2));
+        T phi = std::atan(normalized_vec(1) / normalized_vec(0));
```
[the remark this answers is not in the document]
tanishqaggarwal: Yeah I was discussing this with Shihao--atan2 would be cleaner but I don't feel like changing the implementation and re-validating all the stuff I did.

shihaocao (approved):
bet. thanks for grinding through this

PR #601, opened 2020-12-25
Normalize the sun sensor vector

### LINKED ISSUE #536: Normalize Sun Vector Reading from the ADCS Box
Given the sun vector is serialized and deserialized when passed over I2C, we should be sure we re-normalize to ensure we still have a unit vector at the end of the process.

tanishqaggarwal:
Fixes #536 

### Summary of changes
- Idk why we haven't done this, it takes 5 seconds

### Testing
Existing unit tests are sufficient.

One concern might be that if the returned vector has close-to-zero magnitude (e.g. if the I2C comms dropped out or bc the ADCS box pooped) we might end up crashing FSW. @kylekrol thoughts?

### Constants and Telemetry
No changes

### Documentation Evidence
This is just a minor implementation fix


shihaocao:
If at all possible, I think this code should actually be in the AttitudeEstimator. I think that ADCSBoxMonitor should do as little processing as possible so that the data we see is as representative as possible of the data coming over the I2C bus.

shihaocao (changes requested):
^

constants
```diff
@@ -95,8 +95,8 @@ src/fsw/FCCode/MissionManager.hpp:32: "initial_close_approach_trigger_dist" = "1
 src/fsw/FCCode/MissionManager.hpp:33: "initial_docking_trigger_dist" = "0.4"
 src/fsw/FCCode/MissionManager.hpp:39: "deployment_wait" = "PAN::one_day_ccno / (24 * 2)"
 src/fsw/FCCode/MissionManager.hpp:45: "initial_docking_timeout_limit" = "PAN::one_day_ccno"
-src/fsw/FCCode/OrbitController.hpp:18: "valve_time_lin_reg_slope" = "0.024119"
-src/fsw/FCCode/OrbitController.hpp:19: "valve_time_lin_reg_intercept" = "7.0092e-05"
+src/fsw/FCCode/OrbitController.hpp:17: "valve_time_lin_reg_slope" = "0.024119"
+src/fsw/FCCode/OrbitController.hpp:18: "valve_time_lin_reg_intercept" = "7.0092e-05"
```
shihaocao, line 99: lol confused as to what happened here, saw it locally too tho

shihaocao (approved):

constants
tanishqaggarwal: discussed offline: different PIO environments cause the line numbers to change around. Not sure why a different PIO environment's constant file was allowed to make it, though

PR #603, opened 2020-12-25
PSim Configuration in PTest

shihaocao:
# PSim Configuration in PTest

Complementary PR to PSim PR:
https://github.com/pathfinder-for-autonomous-navigation/psim/pull/285

### Summary of changes
- Added new property: sim_ic_map, a dictionary with psim state field names as strings, and values as initial conditions to overwrite
- Updated PTest simulation backend to support

### Supporting Bugfixes
- allowed int fields to be plotted by not blindly converting all fields to lists

### Testing
Tested on PsimDebug

### Example:
image
Define the sim_ic_map property as a dict with the things you want to mutate

And then, the ptest terminal will tell you what is mutated, and you can see the changes reflected in the sim
image
image


ptest/cases/base.py
```diff
@@ -320,16 +330,16 @@ def rs_psim(self, name):
         if(ret is None):
             raise NameError(f"ptest read failed: psim state field {name} does not exist!")
         
+        stripped = ret
         if type(ret) in {lin.Vector2, lin.Vector3, lin.Vector4}:
             ret = list(ret)
-        
-        stripped = str(ret).strip("[]").replace(" ","")+","
+            stripped = str(ret).strip("[]").replace(" ","")+","
```
tanishqaggarwal, line 336: What is this doing?

ptest/psim/simulation.py
```diff
@@ -13,6 +13,8 @@
 import psim
 import lin 
 import json
+from ..cases.utils import is_lin_vector
+from ..cases.utils import to_lin_vector
```
tanishqaggarwal, line 17:
```suggestion
from ..cases.utils import is_lin_vector, to_lin_vector
```

tanishqaggarwal (changes requested):

ptest/cases/base.py
shihaocao: This is actually a bug fix that prevented int fields from being plotted

shihaocao: Now psim int fields can be plotted in ptest

tanishqaggarwal (approved):

kylekrol (approved):
Seems good to me! Only thing that I wonder is if making a callback function that gives PSim related testcases a chance to operate directly on the `config` may be easier that using a dictionary passed down by a property.

That's just a suggesting, not at all a request, so going ahead and approving.

kylekrol:
@shihaocao forgot we need to bump the psim pointer again but yeah after that everything looks good!

shihaocao:
> @shihaocao forgot we need to bump the psim pointer again but yeah after that everything looks good! Perhaps add the psim debug case to CI as well?

Lol PsimDebug has just been this running testcase for me to check the functionality of each new feature I add, and was never really meant to be a staple ci case... Could def make it one tho

I'll consider it on the next PR where I delete ci.txt, and actually use this infrastructure for faster detumbling

kylekrol:
> > @shihaocao forgot we need to bump the psim pointer again but yeah after that everything looks good! Perhaps add the psim debug case to CI as well?
> 
> Lol PsimDebug has just been this running testcase for me to check the functionality of each new feature I add, and was never really meant to be a staple ci case... Could def make it one tho
> 
> I'll consider it on the next PR where I delete ci.txt, and actually use this infrastructure for the detumble ci case
> @kylekrol

Sounds good. But yeah was just thinking it may be good to get some of this going in CI to make sure bumping the psim pointer won't ever break some of this basic functionality.

PR #609, opened 2021-01-05
Prop fault response ptest

athena255:
PropFaultHandler Ptest Case
Summary of changes

    prop_fault_handler.py rewritten to have tests based on faked sensor values
    Fault.cpp modified signal() to set the faulted state when num_consecutive_signals exceeds persistence
    MainFaultHandler.cpp added PropFaultHandler to fault_handler_machines
    MissionManager.cpp significant modifications, see below
    PropController.cpp::execute() check_faults() now occurs at the start of execution instead of at the end of execution. This is to allow the ptest to cause faults by setting fake sensor values after the call to cycle()
    PropController.cpp::check_faults(), can_enter() modified to make sure that all faults call is_faulted() since the prior implementation had the calls in a boolean expression, which was short-circuited resulting in one-cycle delays when evaluating faults that should be triggered in the same cycle
    PropController.hpp functions for checking whether sensors exceed max safe values have been modified to use > instead of >=
    PropFaultHandler.cpp changes to cast prop state to unsigned int instead of unsigned char
    PropFaultHandler.h removed pressurize_fail_fault since PFH does not do anything with his fault
    test_main_fault_handler.cpp modified to reflect changes to MissionManager
    test_mission_manager.cpp modified to reflect changes to MissionManager
    test_prop_controller.cpp added tests for venting both tanks
    test_prop_faults.cpp modified to reflect one cycle delay due to checking for faults at the beginning of PropController execution, also added tests for resolving overpressured fault by suppressing and resolving underpressured faults by unsignaling and lowering the threshold pressure

Bug Fixes

    Previously, PropFaultHandler would recommend a standby when Prop was in handling_fault or venting. However, this causes the MissionManager to transition the prop state machine to idle. This resulted in a one cycle delay in handling_fault
    Because Fault.cpp previously requires a call to is_faulted() to actually set the fault state and PropController used the boolean || to compare fault states, this caused the evaluation to short-circuit, which made the detection of faults one cycle late when multiple faults are signaled at the same time

Significant Behavior Changes

    MissionManager no longer has references to dcdc spike and hold
    PropController sets dcdc pin to false upon entering idle and true upon entering Pressurizing or Venting
    The prop_state_t::disabled prevents the prop from venting. MissionManager was modified to not transition the prop to disabled once the prop has been turned on
    MissionManager no longer transition prop state to idle when the recommended fault response is standby or safehold. This fixes a one-cycle delay bug that occured when venting
    MissionManager.cpp now commands prop into idle during detumbling. The concern was that detumbling has the potential to cause pressure/temperature changes in the prop tank. This behavior change results in the dcdc spike and doc pin being set earlier
    signal() now sets the fault state when num_consecutive_signals exceeds persistence. Previously, the fault state was only evaluated upon calling is_faulted()
    Since PropController now calls check_faults() at the beginning of execution, faults are detected by PropFaultHandler one cycle late

TODO

    OrbitController should only schedule firings when it is in follower

Constants

    N/A

Telemetry

    N/A

prop


src/fsw/FCCode/PropController.cpp
```diff
@@ -570,10 +571,13 @@ void PropState_Venting::enter()
...
+    bool b_overpressured = controller->overpressure_fault_f.is_faulted();
+    bool b_tank2_temp = controller->tank2_temp_high_fault_f.is_faulted();
+    bool b_tank1_temp = controller->tank1_temp_high_fault_f.is_faulted();
     return PropulsionSystem.is_functional() &&
-           (controller->overpressure_fault_f.is_faulted() ||
-           controller->tank1_temp_high_fault_f.is_faulted() ||
-           controller->tank2_temp_high_fault_f.is_faulted());
+           (b_overpressured ||
+           b_tank2_temp ||
+           b_tank1_temp);
```
shihaocao, line 580: to prevent short circuit eval

ptest/cases/prop_state_machine_case.py
```diff
@@ -179,7 +178,7 @@ def print_object(self):
 
     @property
     def min_num_cycles(self):
-        return (int(self.ctrl_cycles_per_filling) + int(self.ctrl_cycles_per_cooling))*int(self.max_pressurizing_cycles) + 4
+        return (int(self.ctrl_cycles_per_filling) + int(self.ctrl_cycles_per_cooling))*int(self.max_pressurizing_cycles) + 4 + int(self.ctrl_cycles_per_filling)
```
athena255, line 181: modified to account for Prop State Machine change which requires waiting in Await_Pressurizing for 1 second in which DCDC pin is high.

ptest/cases/prop_state_machine_case.py
```diff
@@ -14,7 +14,6 @@ def fast_boot(self):
         return True
 
     def setup_post_bootsetup(self):
-        self.flight_controller.write_state("dcdc.SpikeDock_cmd", True)
```
athena255, line 17: Since prop sets this itself now, we do not need to set this anymore

ptest/cases/prop_state_machine_case.py
```diff
@@ -233,8 +230,11 @@ def test_idle_to_await_pressurizing(self):
         self.sched_valve2 = 700
         self.sched_valve3 = 800
         self.sched_valve4 = 400
-        self.cycles_until_firing = int(self.min_num_cycles) + 2
+        self.cycles_until_firing = int(self.min_num_cycles)
         self.cycle()
+        # Spike and Hold DCDC should be true
+        if not self.read_state("dcdc.SpikeDock_cmd") == "true":
```
athena255, line 236: Prop should set this right after it receives a valid schedule

ptest/cases/prop_state_machine_case.py
```diff
@@ -279,6 +279,8 @@ def test_firing_to_idle(self):
         print(f"[TESTCASE] cycles_until_change: {self.cycle_until_change()}")
         if not self.state == str(Enums.prop_states["idle"]):
             raise TestCaseFailure("state != idle")
+        if not self.read_state("dcdc.SpikeDock_cmd") == "false":
```
athena255, line 282: Prop should unset DCDC on transition to idle

src/common/Fault.cpp
```diff
@@ -30,6 +30,11 @@ void Fault::signal() {
         num_consecutive_signals++;
         last_fault_time = *cc;
     }
+    if (num_consecutive_signals > persistence_f.get()) {
```
athena255, line 33: signal() should set the fault upon detecting that num_consecutive_signals exceeds persistence

src/fsw/FCCode/MissionManager.cpp
```diff
@@ -69,9 +69,7 @@ MissionManager::MissionManager(StateFieldRegistry &registry, unsigned int offset
     wheel3_adc_fault_fp = find_fault("adcs_monitor.wheel3_fault.base", __FILE__, __LINE__);
     wheel_pot_fault_fp = find_fault("adcs_monitor.wheel_pot_fault.base", __FILE__, __LINE__);
     pressurize_fail_fp = find_fault("prop.pressurize_fail.base", __FILE__, __LINE__);
-
```
athena255, line 72: MissionManager no longer needs dcdc.SpikeDoc_cmd since prop sets it itself

src/fsw/FCCode/MissionManager.cpp
```diff
@@ -392,15 +382,6 @@ void MissionManager::transition_to(mission_state_t mission_state,
                                    adcs_state_t adcs_state,
                                    prop_state_t prop_state)
 {
-    if (prop_state == prop_state_t::disabled && mission_state != mission_state_t::docking)
```
athena255, line 395: MissionManager no longer sets SpikeHold DCDC

src/fsw/FCCode/OrbitController.cpp
```diff
@@ -189,5 +189,5 @@ unsigned int OrbitController::prop_min_cycles_needed() {
     return max_pressurizing_cycles_fp->get() *
                (ctrl_cycles_per_filling_period_fp->get() +
                 ctrl_cycles_per_cooling_period_fp->get()) +
-           4;
+           4 + ctrl_cycles_per_filling_period_fp->get();
```
athena255, line 192: The min_num_cycles to set for cycles_until_firing has been increased by 1 second (`ctrl_cycles_per_filling_period`) to require that AwaitPressurizing reads DCDC SpikeDock high for ~1 second

src/fsw/FCCode/PropController.cpp
```diff
@@ -100,6 +101,7 @@ PropState_Manual PropController::state_manual;
 
 void PropController::execute()
 {
+    check_faults();
```
athena255, line 104: Faults are checked at the beginning of the cycle to enable ptest cases, which cause faults by faking sensor values

src/fsw/FCCode/PropController.cpp
```diff
@@ -111,36 +113,42 @@ void PropController::execute()
...
     tank2_pressure_f.set(Tank2.get_pressure());
     tank2_temp_f.set(Tank2.get_temp());
     tank1_temp_f.set(Tank1.get_temp());
-
-    check_faults();
 }
 
 void PropController::check_faults()
 {
+    // Do not check_faults when we are disabled
```
athena255, line 140: Faults are no longer checked when the prop is in disabled

src/fsw/FCCode/PropController.cpp
```diff
@@ -111,36 +113,42 @@ void PropController::execute()
...
+    if (static_cast<prop_state_t>(prop_state_f.get()) == prop_state_t::disabled)
+        return;
     overpressure_fault_f.evaluate(is_tank2_overpressured());
     tank2_temp_high_fault_f.evaluate(is_tank2_temp_high());
     tank1_temp_high_fault_f.evaluate(is_tank1_temp_high());
+
```
athena255, line 146: Set SpikeHold DCDC upon detecting potential venting conditions. 
Warning: there is no code to unset this until prop naturally transitions to idle

src/fsw/FCCode/PropController.cpp
```diff
@@ -193,12 +201,16 @@ bool PropController::is_valid_schedule(unsigned int v1,
     return (v1 <= 1000 && v2 <= 1000 && v3 <= 1000 && v4 <= 1000 && ctrl_cycles_from_now > 1);
 }
 
-unsigned int PropController::min_cycles_needed() const
+unsigned int PropController::min_cycles_pressurizing_only() const
```
athena255, line 204: This is the old value of min_cycles_needed().

src/fsw/FCCode/PropController.cpp
```diff
@@ -193,12 +201,16 @@ bool PropController::is_valid_schedule(unsigned int v1,
...
 {
     return max_pressurizing_cycles.get() *
                (ctrl_cycles_per_filling_period.get() +
-                ctrl_cycles_per_cooling_period.get()) +
-           4;
+                ctrl_cycles_per_cooling_period.get()) + 4;
+}
```
athena255, line 209: Modified to account for the 1s in Await_Pressurizing where DCDC SpikeHold pin must be high

src/fsw/FCCode/PropController.cpp
```diff
@@ -240,6 +252,8 @@ void PropState_Disabled::enter()
     DD("[*] ==> entered PropState_Disabled\n");
     // Stop Interval Timer, Turn off all valves, clear the schedule
     PropulsionSystem.reset();
+    // Set dcdc to false
+    controller->sph_dcdc_fp->set(false);
```
athena255, line 256: Set SpikeHold DCDC to false upon entering disabled

src/fsw/FCCode/PropController.cpp
```diff
@@ -256,24 +270,19 @@ prop_state_t PropState_Disabled::evaluate()
 
 bool PropState_Idle::can_enter() const
 {
-#ifndef DESKTOP
-    return PropulsionSystem.is_functional();
-#else
     return true;
```
athena255, line 273: We can always enter idle

src/fsw/FCCode/PropController.cpp
```diff
@@ -256,24 +270,19 @@ prop_state_t PropState_Disabled::evaluate()
...
-#endif
 }
 
 void PropState_Idle::enter()
 {
-    DD("[*] ==> entered PropState_Idle\n");
+    DD("[*] ==> entered PropState_Idle...setting DCDC to false\n");
+    controller->sph_dcdc_fp->set(false);
```
athena255, line 279: Sets SpikeHold DCDC to false upon entering Idle

src/fsw/FCCode/PropController.cpp
```diff
@@ -287,25 +296,45 @@ bool PropState_AwaitPressurizing::can_enter() const
 {
     bool was_idle = controller->check_current_state(prop_state_t::idle);
     bool is_schedule_valid = controller->validate_schedule();
-    // Enter Await Pressurizing rather than Pressurizing if we have MORE than enough time
-    bool more_than_enough_time =
-        controller->cycles_until_firing.get() >= controller->min_cycles_needed();
-    bool is_functional = PropulsionSystem.is_functional();
+    // Enter Await Pressurizing if schedule is valid and we have enough time
```
athena255, line 299: Modified because prop no longer allows transitioning directly into Pressurizing.

src/fsw/FCCode/PropController.cpp
```diff
@@ -287,25 +296,45 @@ bool PropState_AwaitPressurizing::can_enter() const
...
 
-    return (was_idle && is_schedule_valid && more_than_enough_time && is_functional);
+    return (was_idle && is_schedule_valid && enough_time);
 }
 
 void PropState_AwaitPressurizing::enter()
 {
-    DD("[*] ==> entered PropState_AwaitPressurizing\n");
+    DD("[*] ==> entered PropState_AwaitPressurizing, setting DCDC pin high\n");
+    controller->sph_dcdc_fp->set(true);
```
athena255, line 309: SpikeHold DCDC pin is set to true upon entering AwaitPressurizing

src/fsw/FCCode/PropController.cpp
```diff
@@ -287,25 +296,45 @@ bool PropState_AwaitPressurizing::can_enter() const
...
+    // reset the dcdc pin high counter
+    n_cycles_dcdc_high = 0;
 }
 
 prop_state_t PropState_AwaitPressurizing::evaluate()
 {
+    // Require that dcdc pin is consecutively high
+    if (PropulsionSystem.is_functional())
```
athena255, line 317: Require that we digitalRead the actual SpikeHold pin as HIGH consecutively for ~1 second before allowing Prop to go into Pressurizing.

src/fsw/FCCode/PropController.cpp
```diff
@@ -287,25 +296,45 @@ bool PropState_AwaitPressurizing::can_enter() const
...
+        ++n_cycles_dcdc_high;
+    else
+        n_cycles_dcdc_high = 0;
+
     if (controller->can_enter_state(prop_state_t::handling_fault))
         return prop_state_t::handling_fault;
+
+    // If we will not have enough time to pressurize, then go back to idle
+    if (controller->cycles_until_firing.get() < controller->min_cycles_pressurizing_only() - 1)
```
athena255, line 326: Have the prop return to idle when it realizes that it could not possibly fire on time

src/fsw/FCCode/PropController.cpp
```diff
@@ -287,25 +296,45 @@ bool PropState_AwaitPressurizing::can_enter() const
...
+        return prop_state_t::idle;
+
+    // Block AwaitPressurizing from entering Pressurizing until n_cycles_dcdc_high is ~ 1 second
+    if (n_cycles_dcdc_high < controller->ctrl_cycles_per_filling_period.get())
```
athena255, line 330: Prevent prop from going into Pressurizing if it has not seen DCDC high consecutively for 1 second

src/fsw/FCCode/PropController.cpp
```diff
@@ -393,15 +425,15 @@ bool PropState_Pressurizing::can_enter() const
 {
     bool was_await_pressurizing = controller->check_current_state(prop_state_t::await_pressurizing);
     bool is_schedule_valid = controller->validate_schedule();
-    // Allow from idle because sometimes we can immediately pressurize
```
athena255, line 396: Do not allow entering this state from idle

src/fsw/FCCode/PropController.cpp
```diff
@@ -520,11 +552,14 @@ bool PropState_HandlingFault::can_enter() const
     // Return true if any of the faults are actually faulted.
     //  This allows us to ignore bad sensors if the ground
     // decides to suppress/override certain faults
-    return PropulsionSystem.is_functional() &&
-            (controller->pressurize_fail_fault_f.is_faulted() ||
-            controller->overpressure_fault_f.is_faulted() ||
-            controller->tank2_temp_high_fault_f.is_faulted() ||
-            controller->tank1_temp_high_fault_f.is_faulted());
+    bool b_pressurize_fail = controller->pressurize_fail_fault_f.is_faulted();
```
athena255, line 555: Modified to avoid short circuit evaluation of is_faulted()

src/fsw/FCCode/PropController.hpp
```diff
@@ -123,17 +128,17 @@ class PropController : public TimedControlTask<void>
 
     inline bool is_tank2_overpressured() const
     {
-        return tank2_pressure_f.get() >= max_safe_pressure;
+        return tank2_pressure_f.get() > max_safe_pressure;
```
athena255, line 131: tank2 is overpressured only if it exceeds max_safe_pressure rather than if it is at max_safe_pressure (assuming that being at max_safe_pressure is ok)

test/test_fsw_orbit_controller/test_orbit_controller.cpp
```diff
@@ -35,6 +35,9 @@ class TestFixture {
                 baseline_vel_fp = registry.create_readable_lin_vector_field<double>("orbit.baseline_vel", 0, 0, 100);
                 q_body_eci_fp = registry.create_readable_field<lin::Vector4f>("attitude_estimator.q_body_eci");
 
+                // Prop needs this
```
athena255, line 38: Create this statefield because Prop expects it.

test/test_fsw_prop_controller/prop_shared.h
```diff
@@ -113,11 +123,12 @@ class TestFixture
     }
 
     // Step forward the state machine by num control cycle.
-    inline void step(size_t num = 1)
+    inline fault_response_t step(size_t num = 1)
```
athena255, line 126: Modified to return the fault_response_t recommended by prop fault handler

test/test_fsw_prop_controller/test_prop_controller.cpp
```diff
@@ -44,125 +44,82 @@ void test_illegal_schedule()
     check_state(prop_state_t::idle);
 }
 
-// Test that a schedule with fire_cycle set to prop_controller->min_cycles_needed() and all valve
```
athena255, line 47: No more idle to pressurizing

test/test_fsw_prop_controller/test_prop_controller.cpp
```diff
@@ -44,125 +44,82 @@ void test_illegal_schedule()
...
-
-void test_idle_to_pressurizing()
-{
-    test_idle_to_pressurizing_helper();
-}
-
-// Test that a schedule with fire_cycle set to tf.pc->min_cycles_needed() + 1 will cause
+// Test that a schedule with fire_cycle set to tf.pc->min_cycles_needed() will cause
 // Prop to enter await_pressurizing for 1 cycle before entering pressurizing
-void test_idle_to_await_pressurize_helper(bool functional = true)
```
athena255, line 70: Since prop sets DCDC SpikeHold pin, we do not need to test the case where the pin is not already set

test/test_fsw_prop_controller/test_prop_controller.cpp
```diff
@@ -389,18 +351,145 @@ void test_vent_inner_tank()
     TEST_ASSERT_TRUE(Tank1.is_valve_open(0));
 }
 
+void test_vent_both_tanks()
```
athena255, line 354: Added tests to verify the changed Prop State Machine behaviors

test/test_fsw_prop_controller/test_prop_controller.cpp
```diff
@@ -468,8 +556,20 @@ int test_prop_controller()
     RUN_TEST(test_use_backup);
     RUN_TEST(test_vent_outer_tank);
     RUN_TEST(test_vent_inner_tank);
+    RUN_TEST(test_vent_both_tanks);
     RUN_TEST(test_temp_sensor_logic);
     RUN_TEST(test_pressure_sensor_logic);
+    RUN_TEST(test_not_enough_cycles_for_dcdc_high);
```
athena255, line 562: Called multiple times since previously, some of these tests failed when they were called a second time.

test/test_fsw_prop_faults/test_prop_faults.cpp
```diff
@@ -8,18 +8,18 @@ unsigned int &cc_count = TimedControlTaskBase::control_cycle_count;
...
 {
-    // If faults are signalled, but we are in disable, then we should not be handling faults are detecting faults
+    // Faults cannot be signaled in disabled
     TestFixture tf;
     simulate_overpressured(); // we are about to blow up...
+    tf.check_sph_dcdc_on(false);
     // signal the fault to persistence times then detect it on the 11th
-    tf.step(get_persistence(overpressure_fault_f) + 1);
-    assert_fault_state(true, overpressure_fault_f);
+    tf.step(get_persistence(overpressure_fault_f) + 2);
```
athena255, line 16: Faults are no longer detected if prop is in disabled

test/test_fsw_prop_faults/test_prop_faults.cpp
```diff
@@ -28,43 +28,48 @@ void test_underpressured_detect()
...
     assert_fault_state(true, overpressure_fault_f);
 }
 
 void test_tank1_temp_high_detect()
 {
     TestFixture tf;
+    tf.set_state(prop_state_t::idle);
     simulate_tank1_high();
-    tf.step(get_persistence(tank1_temp_high_fault_f) + 1);
+    tf.step(get_persistence(tank1_temp_high_fault_f) + 2);
```
athena255, line 50: Fault is detected a cycle late because check_faults() now occurs at the beginning of the cycle.

test/test_fsw_prop_faults/test_prop_faults.cpp
```diff
@@ -134,10 +141,40 @@ void test_underpressured_response()
     check_state(prop_state_t::firing);
 }
 
+void test_underpressured_unsignal_response()
```
athena255, line 144: Added test for more realistic response to underpressured event. (The ground is expected to unsignal the fault and then lower the threshold pressure)

src/fsw/FCCode/PropController.cpp
```diff
@@ -287,25 +296,45 @@ bool PropState_AwaitPressurizing::can_enter() const
...
+    if (controller->cycles_until_firing.get() < controller->min_cycles_pressurizing_only() - 1)
+        return prop_state_t::idle;
+
+    // Block AwaitPressurizing from entering Pressurizing until n_cycles_dcdc_high is ~ 1 second
+    if (n_cycles_dcdc_high < controller->ctrl_cycles_per_filling_period.get())
+        return this_state;
+
     if (controller->can_enter_state(prop_state_t::pressurizing))
+    {
+        n_cycles_dcdc_high = 0; // reset this just in case
```
shihaocao, line 335: what's this for?

src/fsw/FCCode/MainFaultHandler.cpp
```diff
@@ -53,6 +53,7 @@ void MainFaultHandler::init()
 
     fault_handler_machines.push_back(std::make_unique<QuakeFaultHandler>(_registry));
     fault_handler_machines.push_back(std::make_unique<PiksiFaultHandler>(_registry));
+    fault_handler_machines.push_back(std::make_unique<PropFaultHandler>(_registry));
```
shihaocao, line 56: How does this change interface with line 26 of this file? In other words do we need to remove line 26?

src/fsw/FCCode/MissionManager.cpp
```diff
@@ -191,7 +189,7 @@ void MissionManager::dispatch_startup()
         is_deployed_f.set(true);
         transition_to(mission_state_t::detumble,
                       adcs_state_t::detumble,
-                      prop_state_t::disabled);
+                      prop_state_t::idle);
```
shihaocao, line 192: The only time we derail the state machine. If we are handling a fault (like venting and stuff) when transitioning from some other startup to detumble though (which is a small chance), it would appear as if we could potentially mess up the state machine?

I am okay with this small probability though

src/fsw/FCCode/OrbitController.cpp
shihaocao: Where is the setting that actually defines this ~1 second wait

shihaocao: oh i see it in the constants now

src/fsw/FCCode/PropController.cpp
```diff
@@ -111,36 +113,42 @@ void PropController::execute()
...
             prop_state_f.set(static_cast<unsigned int>(next_state));
             get_state(next_state).enter();
         }
         else
         {
-            // This could happen if next_state is idle but is_functional returns false
-            // or if powercycling is happening 
+            // This could happen if next_state requires is_functional(), but is_functional is not set
             DD("[-] Could not enter state!!\n\n\n");
             prop_state_f.set(static_cast<unsigned int>(prop_state_t::disabled));
```
shihaocao, line 128: Do we ever expect this to happen if all hardware is performing nonmially?
If not, could we tag this with an "[ ERROR ]"?

shihaocao: on the debug print out that is

src/fsw/FCCode/PropController.cpp
shihaocao: Why? Is it possible to still have faults checked, but just have them not let us leave disabled?

src/fsw/FCCode/PropController.cpp
shihaocao: Seems good to me

If a ground command simultaneously suppressed all faults, set prop state to idle, and set dcdc off, it should be as if it were in a normal idle state though right?

src/fsw/FCCode/PropController.cpp
shihaocao: What's the 4 about again?

src/fsw/FCCode/PropController.cpp
shihaocao: The above code looks unmodified?

src/fsw/FCCode/PropFaultHandler.cpp
```diff
@@ -28,7 +28,6 @@ void PropFaultHandler::init()
 
     // Faults
     overpressure_fault_fp = find_fault("prop.overpressured.base", __FILE__, __LINE__);
-    pressurize_fail_fault_fp = find_fault("prop.pressurize_fail.base", __FILE__, __LINE__);
```
shihaocao, line 31: curious as to why prop_pressurize_fail is not in here?

shihaocao (commented):
Got a bunch of questions, but otherwise, looking good.

src/fsw/FCCode/PropController.cpp
athena255: We could, but I think we decided that prop should not check faults while in disabled since there's nothing it could do about it anyway

src/fsw/FCCode/PropController.cpp
athena255: yes

src/fsw/FCCode/PropController.cpp
athena255: The min_cycles_needed is the minimum number of cycles until we can fire. That means after min_cycles_needed, we should be in firing. There's an offset of four cycles due to the state transitions: 
idle: schedule is set
1. idle -> await_press: upon detecting valid schedule
2. await_pres -> pres: 
3. pres -> await_firing
4. await_firing->firing

src/fsw/FCCode/PropController.cpp
athena255: Mb, I meant to put this comment on the next line where min_cycles_needed was modified

src/fsw/FCCode/PropController.cpp
athena255: The n_cycles_dcdc_high is the variable to count the number of cycles in which dcdc was digitalRead to be HIGH. We reset this when we enter pressurizing since this variable only matters in await_pressurizing. This is not truly necessary here since the Await_Pressurizing::can_enter() method resets this anyway, but I set this to 0 just in case .

src/fsw/FCCode/MainFaultHandler.cpp
athena255: Line 26 for me is: `find_fault("adcs_monitor.wheel_pot_fault.base", __FILE__, __LINE__)`

src/fsw/FCCode/MissionManager.cpp
athena255: We never truly mess up the state machine. If prop is transitioned to idle when it wants to be in venting, it will immediately transition itself to handling_fault and then to venting again. The machine would only truly be disabled if it was constantly commanded into idle. Worst case scenario would be that venting is two cycle delayed every time it's forced into idle.

src/fsw/FCCode/PropController.cpp
athena255: This case actually tends to happen when the dcdc is switched off while the prop needed it. For example, if prop was in Await_Firing, it would attempt to transition to Firing when it's time to fire. But since Firing requires that the dcdc be on, it will fail to transition and go into disabled. The test cases for powerloss that I added in this pr should demonstrate this behavior (i believe).

src/fsw/FCCode/PropFaultHandler.cpp
athena255: Because PropFaultHandler does not do anything with pressurize_fail. If we fail to pressurize, we go into handling fault until the ground does something about it.

src/fsw/FCCode/PropController.cpp
shihaocao: I think we should still check faults, that way we can still query these faults to get diagnostic data for the spacecraft

src/fsw/FCCode/MainFaultHandler.cpp
shihaocao: Oops meant line 30, `find_fault("prop.pressurize_fail.base", __FILE__, __LINE__),`

My question is like, should that fail pressurize fault still there? Is this like double counted?

src/fsw/FCCode/MissionManager.cpp
shihaocao: Ah bet

src/fsw/FCCode/PropController.cpp
shihaocao: Gotcha. I guess my question is like, it would only be the case that DCDC is off when we needed it to be on if our code was bugged, or if the hardware was not acting like we expected to right?

In other words, if we see this message, we need to do bug hunting before launch? (I think the answer is yes)?

src/fsw/FCCode/PropFaultHandler.cpp
shihaocao: Ah I see

src/fsw/FCCode/MainFaultHandler.cpp
shihaocao: I think my own answer to this is, yes we should still have line 30 because that allows us to ask the mission manager to go to standby if we can't pressurize the tanks

Thus nothing is double-counted, and this line change is good too.

src/fsw/FCCode/PropController.cpp
athena255: We would only see this message in power loss, but yea, this check is essentially a sanity check.

test/test_fsw_mission_manager/test_mission_manager.cpp
```diff
@@ -88,6 +90,7 @@ void test_dispatch_detumble() {
     tf.step();
     tf.check(adcs_state_t::point_standby);
     tf.check(mission_state_t::standby);
+    // TODO
```
shihaocao, line 93: Todo flags?

test/test_fsw_mission_manager/test_mission_manager.cpp
```diff
@@ -39,7 +39,7 @@ void test_dispatch_startup() {
     // should transition to the detumble state.
     tf.step();
     tf.check(mission_state_t::detumble);
-    tf.check_sph_dcdc_on(false);
+    tf.check_sph_dcdc_on(true);
```
shihaocao, line 42: has the todo on line 36 been resolved?

shihaocao (changes requested):
We should set prop to disabled and dcdc to off if we're in safehold, and to go back to any kind of normal mode from safehold, the ground will be responsible for re-enabling the DCDC and setting prop to idle

test/test_fsw_mission_manager/test_mission_manager.cpp
athena255: The TODO was to note that the check_sph_dcdc_on calls are redundant since we manually set dcdc at the beginning of the test case anyway.

athena255:
> We should set prop to disabled and dcdc to off if we're in safehold, and to go back to any kind of normal mode from safehold, the ground will be responsible for re-enabling the DCDC and setting prop to idle

What about the case where ground forgets to re-enable the DCDC and the subsystems that require DCDC start reporting faults? It may be safer to do the re-enable in software. That will be one less statefield in the packet and will not require ground (students) to remember to re-enable DCDC. 

Do you think we should put this feature in a different PR?

shihaocao (approved):
Bet. Thanks for grinding so hard on this

tanishqaggarwal:
I haven't really been able to keep up with the complex discussion on this PR, so please don't wait for my approval before merging

PR #623, opened 2021-02-04
Set PPT mode to AUTO (1)

shihaocao:
# Set PPT mode to AUTO (1)

Note:
This PR no longer has changes related to forcing a loop to trigger the watchdog timer since the reset command seems to fully reset the spacecraft succesfully.

### Changes
- Change the default ppt mode to auto and max tracking
- Change the unit test to accept an FSW default

### Testing
Tested on 02/04: POC Shihao Cao


kylekrol (changes requested):
Changes look good to me; although, it seems that we need to run some unit tests locally to fix CI. I'm sure we broke a couple HOOTL, HITL, and unit tests.

src/fsw/FCCode/GomspaceController.hpp
```diff
@@ -8,7 +8,7 @@
 #include <common/constant_tracker.hpp>
 
 TRACKED_CONSTANT_SC(unsigned int, default_pv_cmd, 4000);
-TRACKED_CONSTANT_SC(unsigned char, default_ppt_mode, 2);
+TRACKED_CONSTANT_SC(unsigned char, default_ppt_mode, 1);
```
kylekrol, line 11: Wooo!

shihaocao:
Athena pointed out that if we went into this while loop while the inner tank was venting to the outer tank, we'd definitely overpressurize and explode. 

She suggested that instead of using a while loop, we just implement a "mute" on GomSpace behavior

kylekrol:
Ah okay fair enough. So something like if the reboot flag is true then the GOMSpace control task just doesn't execute any code?

src/fsw/FCCode/GomspaceController.cpp
```diff
@@ -191,6 +191,10 @@ GomspaceController::GomspaceController(StateFieldRegistry &registry, unsigned in
      }
 
 void GomspaceController::execute() {
+    // Prevent watchdog timer petting
+    if(gs_reset_cmd_f.get())
+        return;
```
shihaocao, line 196: doing it this way still allows other control tasks to attempt to continue working, notably, PropSystem still maintains the ability to close the intertank valve.

src/fsw/FCCode/GomspaceController.cpp
```diff
@@ -211,8 +212,6 @@ void GomspaceController::execute() {
         pv2_output_cmd_f.set(gs.hk->vboost[1]);
         pv3_output_cmd_f.set(gs.hk->vboost[2]);
 
-        ppt_mode_cmd_f.set(gs.hk->pptmode);
```
shihaocao, line 214: removing this, so that FlightSoftware still has authority to set a default

shihaocao:
I removed the reboot loop because the reset command fully resets the spacecraft. Thus we should just not use the reboot command, and only the reset command.

fatimayousuf (approved):

kylekrol (approved):

PR #625, opened 2021-02-05
Gomspacelogger

knk38:
# Adding Gomspace Logging Test Case

### Summary of changes
- New test case that reads important gomspace fields for the gomspace charging capability test
- Add test case to init

### DIFF
--- a/ptest/cases/__init__.py
+++ b/ptest/cases/__init__.py
@@ -31,3 +31,4 @@ from .hardware_stress_test_case import HardwareStressCheckoutCase
 from .torquer_rotate_case import MTorquerCase
 from .reset_for_flight import ResetforFlight
 from .prep_for_charging_case import PrepForCharging
+from .gomspace_logger import GomspaceLog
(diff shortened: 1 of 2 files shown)


kylekrol:
Is there a chance that we would want to create more of these cases in the future? If so, we could make a base case like `LoggerCase` that you can inherit from and then just pass a list of state fields to the parent class - not a request just a thought.

shihaocao (approved):
Looks good to me

shihaocao:
> Is there a chance that we would want to create more of these cases in the future? If so, we could make a base case like `LoggerCase` that you can inherit from and then just pass a list of state fields to the parent class - not a request just a thought.

We could totally do that on a next PR, getting this in to prep for spotlight testing today

knk38:
> > Is there a chance that we would want to create more of these cases in the future? If so, we could make a base case like `LoggerCase` that you can inherit from and then just pass a list of state fields to the parent class - not a request just a thought.
> 
> We could totally do that on a next PR, getting this in to prep for spotlight testing today

Agreed, I'll merge this in for now.

PR #632, opened 2021-02-11
Updates to DITL Case Including Instructions

kylekrol:
### Summary of changes

- Updated the DITL case wheel speed to be the true starting speed.
- Raised the halting battery voltage per the GOMSpace datasheet recommendation.
- Added additional comments to describe running, logging, and documenting DITL case runs.

### Testing

The testcase was run on the Follower unit on 02-10-21.
POC: @kylekrol

### Constants

NA

### Telemetry

NA

### Documentation Evidence

Inline with the testcase itself.

### DIFF
--- a/ptest/cases/ditl_case.py
+++ b/ptest/cases/ditl_case.py
@@ -1,11 +1,29 @@
 """Day-In-The-Life (DITL) testcase used to check battery health.
 
-The intention here is to run the satellites actuators using realistic actuator
-commands to gain a sense of how healthy the batteries on the flight units are.
-The actuators of primary concern are the reaction wheels and quake radio.
+The original intention here was to run the satellite actuators using realistic
+actuator commands to see how long the satellite can run on battery power to gain
+insight into battery health.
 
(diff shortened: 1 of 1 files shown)


ptest/cases/ditl_case.py
```diff
@@ -1,11 +1,29 @@
 """Day-In-The-Life (DITL) testcase used to check battery health.
 
-The intention here is to run the satellites actuators using realistic actuator
-commands to gain a sense of how healthy the batteries on the flight units are.
-The actuators of primary concern are the reaction wheels and quake radio.
+The original intention here was to run the satellite actuators using realistic
```
tanishqaggarwal, line 3: Why change to past tense?

tanishqaggarwal (approved):

ptest/cases/ditl_case.py
kylekrol: Because the the case doesn't exactly do that - we aren't running the radio primarily. So the purpose has changed slightly - i.e. favoring consistency in testing over making it as absolutely realistic as possible.

Issue #629, opened 2021-02-10
Fix RadioSession Uplink Logic with AMC


afoarce:
Currently, RadioSession does not let AMC write states multiple times due to the json file being made only on the first write. This means as AMC steps through logic, the commands aren't actually being written to send to Iridium. Options for fixing this issue are mentioned in the discussion of #567

Issue #622, opened 2021-02-01
[PTest] Test RadioSession with AMC


afoarce:
Currently, the way RadioSession makes and stores statefields in a json for uplinks might not allow changes to states already written to or the addition of states to the packet after writing multiple states once. We need to adjust RadioSession as necessary so that the states written in each iteration of the AMC get written to an uplink, and test that this behavior works successfully.

afoarce:
Fixed by #649

PR #635, opened 2021-02-13
Pure autonomous mission controller w/o bootutility

### LINKED ISSUE #619: [PTest] - Pure Autonomous Mission Controller [Framework Only]
# Why
We need a PTest case (or some modification of one), to eventually evolve into or just be, the system that will run the mission of Satellites once they are launched.

# What we have right now
Pending Andreea's PR, we will have an AMC, but it currently also is responsible for booting HITL/HOOTL instances.

# What we need to do
- We need to divorce the current AMC of actually "starting" any HITL/HOOTL instances or at least a way to configure a PTest case to do so
- Specifically, AMC should be configured to connect to exactly two RadioSessions, as this is the flight-like condition

# AMC's specific features to keep in mind (can be deferred to future tickets):
- AMC must uplink data about one satellite to the other, and vice versa so that they can perform a successful rendezvous
- AMC's automatic uplinks must be able to "mesh" nicely with manual commands/uplinks
- AMC must make intelligent decisions upon when to set satellites into specific modes, such as setting spacecraft roles
- Further documentation is found here: https://github.com/pathfinder-for-autonomous-navigation/FlightSoftware/issues/476

# What's after this?
Once the AMC framework is created, the we can then start on giving the AMC its actual brains and features outlined here:
https://github.com/pathfinder-for-autonomous-navigation/FlightSoftware/issues/476

afoarce:
# Pure AMC w/o bootutility

Fixes #619 

### Summary of changes
- Added checks for if config json has any devices before initializes leader/follower for mission cases to avoid booting up hitl/hootl instances of the satellites
- AMC assumes config file has two radios, for testing can be the same radio
- Original Autonomous Mission Manager(autonomous_mission_manager_case.py) copied/modified into autonomous_mission_manager_pure_radio.py

Future work includes testing and fixing the RadioSession interactions, as described in #622 and #629 

### Testing
Changes tested in HOOTL but as noted above needs to be tested further.


ptest/cases/base.py
```diff
@@ -426,30 +428,32 @@ def fast_boot_follower(self):
...
-        self.boot_util_follower = BootUtil(
-            self.flight_controller_follower, self.logger, self.initial_state_follower, 
-            self.fast_boot_follower, self.one_day_ccno_follower, self.suppress_faults)
-        self.boot_util_leader.setup_boot()
-        self.boot_util_follower.setup_boot()
-        self.setup_post_bootsetup_leader()
-        self.setup_post_bootsetup_follower()
+        if self.devices != None:
+            self.setup_pre_bootsetup_leader()
+            self.setup_pre_bootsetup_follower()
```
shihaocao, line 433: expose this out of the != None

shihaocao: nvm

ptest/cases/autonomous_mission_manager_pure_radio.py
```diff
@@ -0,0 +1,114 @@
...
+    def sim_model(self):
+        return DualAttitudeOrbitGnc
+
+    @property
+    def sim_mapping(self):
+        return "ci_mapping.json"
+        
+    @property
+    def sim_duration(self):
+        return float("inf")
```
shihaocao, line 71: still running psim, we should take this out

shihaocao: set float("inf") -> 0

ptest/cases/autonomous_mission_manager_pure_radio.py
```diff
@@ -0,0 +1,114 @@
...
+    def sim_model(self):
+        return DualAttitudeOrbitGnc
+
+    @property
+    def sim_mapping(self):
+        return "ci_mapping.json"
+        
+    @property
+    def sim_duration(self):
+        return 0
```
kylekrol, line 71: Just curious here why we are providing simulation information? Shouldn't we not be using a simulation at all?

kylekrol: Discussed on Discord, these fields need to remain to avoid a crash.

kylekrol (commented):
Seems pretty good to me. Although I'll defer to @shihaocao for a more in depth review - I'm still iffy on `ptest` related things.

shihaocao (approved):

shihaocao:
Looks good. Let's demo this asap

PR #638, opened 2021-02-17
Utilize Arduino's Loop Function

kylekrol:
# Utilize Arduino's Loop Function

### Summary of changes
- Update the `targets/teensy.cpp` file to move the main control loop execute calls into Arduino's `void loop(void)` function. It was discovered that the Arduino library call rely on loop being called repeatedly for certain functionality.

### Testing

Tested in HITL with a mission level testcase. Observed nominal state transitions, ADCS system initialization, etc.

### Constants
NA

### Telemetry
NA

NA

### DIFF
--- a/src/fsw/targets/teensy.cpp
+++ b/src/fsw/targets/teensy.cpp
@@ -12,23 +12,15 @@
 #include <core_pins.h>
 #include <wiring.h>
 
-void pan_system_setup() {
-    StateFieldRegistry registry;
-    MainControlLoop fcp(registry, PAN::flow_data);
-
-    while (true) {
-        fcp.execute();
-    }
-}
-
 // "ifndef UNIT_TEST" used to stop "multiple definition" linker errors when running
 // tests
 #ifndef UNIT_TEST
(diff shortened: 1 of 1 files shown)


tanishqaggarwal (approved):
Obviously, don't merge until HITL. I have a suspicion HITL will fail--initializing the StateFieldRegistry statically may cause some weird failures.

kylekrol:
> Obviously, don't merge until HITL. I have a suspicion HITL will fail--initializing the StateFieldRegistry statically may cause some weird failures.

If my understanding of `static` in this context is correct, the variables won't be initialized until the function is called for the first time - it's part of the singleton design pattern. That being said, C++ consistently surprises me haha

shihaocao:
looks good to me too, we'll hitl this on saturday

shihaocao:
bump reminder to hitl test this today

kylekrol:
This was HITL tested through a startup -> detumble -> standby -> safehold transition testcase.

POC: @kylekrol

PR #639, opened 2021-02-20
Email scraper

### LINKED ISSUE #618: [PTest] - in:sent Scraper
# Why
We need to mimic uplinks to HOOTL/HITL instances without physical Quake radios, that still utilize binary data to test our serializer and deserializer, as well as our GSW suite

# What we have right now
Normally we uplink to HOOTL/HITL instances without Quakes using `ws` through USBSession.

# What we need to do
- Write a script that automatically runs that checks the `sent` portion of the iridium email, and tags the IMEI number associated with the current USBSession, grabs it, and uses `do_uplink` to write binary uplink telemetry to the HOOTL/HITL instance
- The above script might actually be better off as a pure function, and instead, be called during the normal execution loop of PTest, that way no additional thread is created

fatimayousuf:
# Email Scraper for Dual HOOTL Testing

Fixes #618 

### Summary of changes
- We want to test communication between satellites without actually using Quake radios/Iridium
- When using the radio session http endpoint to send an uplink, the endpoint will mark the sent message and uplink packet as "unseen"
- Then we will continuously look into the Sent Mail box in the pan email account, get all the uplink packets addressed to the specific satellite, and forward it to the flight computer. 
- There are three ways to scrape emailed uplinks: 
   - upon starting ptest, you can add the --scrape_uplinks option, which will start a thread in usb_session to continuously scrape uplinks. 
   - In a ptest case with a simulation, you can set the sim property `scrape_uplinks` to `True`. While the simulation is running, it will continuously check for uplinks in the Sent Mail box
   - In the command prompt, you can run `scrapeuplinks` and it will look through the sent mail box _once_.
 
### Testing
I can send an email using the radio session endpoint, and the ptest terminal will show that the uplink packet has been sent to Flight Software.


ptest/radio_session.py
```diff
@@ -70,9 +76,13 @@ def __init__(self, device_name, imei, uplink_console, port, send_queue_duration,
...
-        #email
-        self.username=tlm_config["email_username"]
-        self.password=tlm_config["email_password"]
+        # Connect to email
+        try:
+            self.mail = imaplib.IMAP4_SSL("imap.gmail.com", 993)
+            self.mail.login(self.username, self.password)
+            self.mail.select('"[Gmail]/Sent Mail"')
+        except Exception as e:
+            print("Not connected to email")
```
kylekrol, line 85: Do we want to raise a `TestCaseFailure` here or does this run in cases without email too?

ptest/runsim.py
```diff
@@ -95,7 +95,10 @@ def set_up_devices(self):
                     # pty isn't defined because we're on Windows
                     self.stop_all(f"Cannot connect to a native binary for device {device_name}, since the current OS is Windows.")
 
-            device_session = USBSession(device_name, self.uplink_console, device["http_port"], is_teensy, self.simulation_run_dir)
+            if 'imei' in device:
+                device_session = USBSession(device_name, self.uplink_console, device["http_port"], is_teensy, self.simulation_run_dir, self.tlm_config, device['imei'])
+            else:
+                device_session = USBSession(device_name, self.uplink_console, device["http_port"], is_teensy, self.tlm_config, self.simulation_run_dir)
```
kylekrol, line 101: I must be missing something - where is the first `USBSession` constructor?

kylekrol: Nevermind - indeed I was just being dumb!

ptest/usb_session.py
```diff
@@ -56,6 +59,16 @@ def __init__(self, device_name, uplink_console, port, is_teensy, simulation_run_
 
         # Open a connection to elasticsearch
         self.es = Elasticsearch([{'host':"127.0.0.1",'port':"9200"}])
+
+        #connect to email
+        self.username=tlm_config["email_username"]
+        self.password=tlm_config["email_password"]
+        self.mail = None
```
kylekrol, line 66: This would prevent crashes if `tlm_config` entries weren't populated. My thought here is we wouldn't have to, for example, send our email password to the launch provider.
```suggestion
        #connect to email
        self.username=tlm_config.get("email_username", "")
        self.password=tlm_config.get("email_password", "")
        self.mail = None
```

ptest/usb_session.py
```diff
@@ -88,6 +101,11 @@ def connect(self, console_port, baud_rate):
                 name=f"{self.device_name} logger thread",
                 target=self.check_console_msgs)
             self.check_msgs_thread.start()
+            self.get_uplinks = True
+            self.scrape_uplinks_thread = threading.Thread(
+                name=f"{self.device_name} uplinks",
+                target=self.scrape_uplinks)
+            self.scrape_uplinks_thread.start()
```
kylekrol, line 108: Should this be conditional on whether or not `self.mail` is `None`?

kylekrol: Oh I see the while condition in the thread. To reduce overhead we could maybe avoid spawning the thread in the first place is no email is present?

ptest/usb_session.py
```diff
@@ -427,6 +445,55 @@ def dbtelem(self):
...
         return not failed
 
+    def scrape_uplinks(self):
+        '''
+        For the AMC tests, we need the Flight Computer to read sent uplinks
+        without actually using Iridium. This method reads from the "sent"
+        box in the PAN email account (attempted uplinks) and passes the uplink packet
+        directly to the Flight computer.
+        '''
+        while self.get_uplinks == True and self.mail != None:
```
kylekrol, line 455: I'm assuming the `self.get_uplinks` will allow us to conditionally enable or disable the email scraper thread? Should we add in a `time.sleep(...)` call here so this isn't just running all the time?

kylekrol:
I think overall this looks pretty awesome! Would like to talk over how we spawn/kill/pause the email thread with @afoarce, @shihaocao, and @Duncan-McD too if possible.

ptest/radio_session.py
kylekrol: nvm this too haha - I see missing email is handled later too!

kylekrol:
Also this seems to be the stacktrace from the failing CI cases:
```
Traceback (most recent call last):
 _______     _       ____  _____                                
  File "/opt/hostedtoolcache/Python/3.6.12/x64/lib/python3.6/runpy.py", line 193, in _run_module_as_main
|_   __ \   / \     |_   \|_   _|                            
    "__main__", mod_spec)
  File "/opt/hostedtoolcache/Python/3.6.12/x64/lib/python3.6/runpy.py", line 85, in _run_code
    exec(code, run_globals)
  File "/home/runner/work/FlightSoftware/FlightSoftware/ptest/__main__.py", line 29, in <module>
    sys.exit(main())
  File "/home/runner/work/FlightSoftware/FlightSoftware/ptest/__main__.py", line 26, in main
    return _command_map[sys.argv[1]].main(sys.argv[2:])
  File "/home/runner/work/FlightSoftware/FlightSoftware/ptest/runsim.py", line 250, in main
    test.start()
  File "/home/runner/work/FlightSoftware/FlightSoftware/ptest/runsim.py", line 49, in start
    self.set_up_devices()
  File "/home/runner/work/FlightSoftware/FlightSoftware/ptest/runsim.py", line 101, in set_up_devices
    device_session = USBSession(device_name, self.uplink_console, device["http_port"], is_teensy, self.tlm_config, self.simulation_run_dir)
  File "/home/runner/work/FlightSoftware/FlightSoftware/ptest/usb_session.py", line 49, in __init__
    self.datastore = Datastore(device_name, simulation_run_dir)
  File "/home/runner/work/FlightSoftware/FlightSoftware/ptest/data_consumers.py", line 59, in __init__
    self.dataLog = open(f"/{data_dir}/{filename}", "w")
FileNotFoundError: [Errno 2] No such file or directory: "/{'email_username': '', 'email_password': '', 'webservice': {'server': '', 'port': 
```
May just be passing a dictionary instead of a string or something?

ptest/usb_session.py
```diff
@@ -89,6 +103,11 @@ def connect(self, console_port, baud_rate):
                 target=self.check_console_msgs)
             self.check_msgs_thread.start()
 
+            self.scrape_uplinks_thread = threading.Thread(
+                name=f"{self.device_name} uplinks",
+                target=self.scrape_uplinks)
+            self.scrape_uplinks_thread.start()
+
```
kylekrol, line 110: When exactly is this getting enabled here? Are we not just letting the body of the full mission testcases handle their own instance of threading, not threading, etc if they want to?

ptest/usb_session.py
```diff
@@ -427,6 +446,68 @@ def dbtelem(self):
...
 
+    def scrape_uplinks(self):
+        '''
+        For the AMC tests, we need the Flight Computer to read sent uplinks
+        without actually using Iridium. This method reads from the "sent"
+        box in the PAN email account (attempted uplinks) and passes the uplink packet
+        directly to the Flight computer.
+        '''
+        while self.scrape == True and self.mail != None:
+            self.scrape()
```
kylekrol, line 457: Main reason for my first comment is I just worry having a thread looping without a lock, sleep, or synchronization barrier could bog down laptops and such once all the other systems are booted (OpenMCT, ElastiSearch, etc).

ptest/http_cmd.py
```diff
@@ -129,6 +129,11 @@ def send_telem():
         sender = "pan.ssds.qlocate@gmail.com"
         subject = imei
         SendMessage(sender, to, subject, "", "", 'uplink.sbd')
+
+        #### FOR AMC TESTING ONLY!!! #####
+        if radio_session.username!="":
+            radio_session.mark_message_unseen()
```
shihaocao, line 135: what's this for?

shihaocao (approved):

PR #640, opened 2021-02-21
Add prop faults suppression to utils.py

shihaocao:
# Add prop faults suppression to utils.py

Prevents prop valves from firing in EDU HITL cases.

### Summary of changes
- Add suppression of prop faults to default utils.py boot utility
- Decreased safehold time to prevent safehold timeout


knk38 (approved):
looks good I changed the value too.

PR #641, opened 2021-02-21
Piksi Interrupt

### LINKED ISSUE #553: Piksi interrupt
- interrupt that shows when new data has come in
- just look at while serial.available- when it goes from 0 to nonzero
- timestamp the first byte
- serialEvent()
- should be part of driver
- call from control task

madeleinemc:
Closes #553

Calculate microdelta = current time - time Piksi data started arriving in the serial port
- Use IntervalTimer for interrupt
- If bytes increase, update timestamp (in terms of times interrupt was called)
- In Piksi.read_all, copy serial port into local buffer and read from that
- Change uart_read to read from local buffer instead of serial port

Tested in PiksiCheckoutCase with Piksi simulation
- Mostly fixed RTK 
- Microdelta in expected range [0, 100000]


src/fsw/FCCode/Drivers/Piksi.cpp
```diff
@@ -3,6 +3,9 @@
 
 #ifndef DESKTOP
 #include <Arduino.h>
+// The below are declared in Arduino.h
+static void interrupts() {}
+void noInterrupts() {}
 #endif
```
kylekrol, line 9: Shouldn't this be something like:
```suggestion
#ifndef DESKTOP
#include <Arduino.h>
#else
// The below are declared in Arduino.h
static void interrupts() {}
static void noInterrupts() {}
#endif
```

src/fsw/FCCode/Drivers/Piksi.cpp
```diff
@@ -271,21 +287,43 @@ unsigned char Piksi::read_all() {
...
+      last_bytes = 0;
+    }
+
+    int initial_time = micros();
+
+    int buffer[bytes];
+    // copy buffer up to available
+    for (int i = 0; i < bytes; i++){
+        buffer[i] = Serial4.read();
+    }
```
kylekrol, line 307: Apologies if I'm jumping on this PR too early, but an `int` takes up (generally) four bytes in memory and `Serial4.read()` only returns single bytes. We'll want our buffer to be of type `unsigned char buffer[bytes]` to keep the memory continguous.

src/fsw/FCCode/Drivers/PropulsionSystem.cpp
```diff
@@ -284,7 +284,9 @@ bool _PropulsionSystem::open_valve(Tank &tank, size_t valve_idx)
 }
 
 void _PropulsionSystem::close_valve(Tank &tank, size_t valve_idx)
-{
+{if (bytes > past_bytes && past_bytes == 0) sendtime = micros();
+    past_bytes = bytes;if (bytes > past_bytes && past_bytes == 0) sendtime = micros();
+    past_bytes = bytes;
```
kylekrol, line 289: This seems like it's not supposed to be here?

src/fsw/FCCode/Drivers/Piksi.cpp
```diff
@@ -333,7 +371,7 @@ void Piksi::clear_bytes() {
 
 u32 Piksi::_uart_read(u8 *buff, u32 n, void *context) {
     #ifndef DESKTOP
-    Piksi *piksi = (Piksi *)context;
+    Piksi *piksi = (Piksi *)context; // so it can access serial port
     
     HardwareSerial &sp = piksi->_serial_port;
```
kylekrol, line 376: We can just go ahead and delete this function - either that or replace the function body with the new implementation (latter is probably easier so you don't have to go and find every `_uart_read` and change it).

kylekrol:
Also the telemetry file is deleted and that doesn't seem right...?

src/fsw/FCCode/Drivers/PropulsionSystem.cpp
madeleinemc: oh god

src/fsw/FCCode/Drivers/Piksi.cpp
```diff
@@ -350,6 +400,25 @@ u32 Piksi::_uart_read(u8 *buff, u32 n, void *context) {
...
+//             buff[i] = sp.read();
+//         else
+//             break;
+//     }
+//     return i;
+//     #else
+//     return 0;
+//     #endif
+// }
+
```
kylekrol, line 421:
```suggestion
```

src/fsw/FCCode/Drivers/Piksi.cpp
```diff
@@ -26,18 +26,56 @@ sbp_msg_callbacks_node_t Piksi::_user_data_callback_node;
...
+        if (time - last_time > 100){
+            start_time = time;
+        }
+
+    last_bytes = bytes;
+    last_time = time;
+   }
+
+    #endif
+}
```
kylekrol, line 59: Just noting here that we have the potential to overflow an unsigned integer here. Now this was an issue with the `micros()` timer before but now we actually have the opportunity to fix it. Setting the interrupt count back to zero when we used to rewrite start time could help avoid this issue. Something like this:
```suggestion
static volatile int last_bytes = 0;
static volatile unsigned long interrupt_count = 0;
static volatile unsigned long last_interrupt_count;

void Piksi::check_bytes(){
#ifndef DESKTOP
    interrupt_count++;
    int bytes = Serial4.available();
    unsigned long count = interrupt_count;

    // if bytes are entering the buffer, update the timestamp
    if (bytes > last_bytes) {
        if (++interrupt_count - last_time > 100) {
            interrupt_count = 0;
        }

        last_bytes = bytes;
        last_interrupt_count = interrupt_count;
   }
#endif
}

```
We'd then have to update I think the calculation of `piksi.microdelta` at little bit as well. This way though we'd be completely guarded against overflows and we could even move back to `unsigned int` types to make the serializer's happy again!

src/fsw/FCCode/Drivers/Piksi.cpp
```diff
@@ -271,32 +309,45 @@ unsigned char Piksi::read_all() {
...
+    buffer_end = buffer;
+    // copy buffer up to available
+    for (int i = 0; i < bytes; i++){
+        *(buffer_end++) = Serial4.read();
+    }
+    clear_bytes();
+
+    microdelta = (time - start_time)*100; //convert to us
+
+    interrupts();
```
kylekrol, line 339: Also, now because the timing is strictly tied to the interrupt schedule I _think_ we should actually get rid of the timing code here. Like it doesn't allow us to gain a better estimate of the time the messages arrived so we may as well just leave it be other than the `last_bytes` value if that makes sense:
```suggestion
    noInterrupts();

    int bytes = Serial4.available();
    last_bytes = 0;

    buffer_begin = buffer;
    buffer_end = buffer;
    // copy buffer up to available
    for (int i = 0; i < bytes; i++){
        *(buffer_end++) = Serial4.read();
    }
    clear_bytes();

    microdelta = interrupt_count*100; //convert to us

    interrupts();

```
Also, with the small booking keeping change in the ISR, I think we can define the ` microdelta` just based on the interrupt count.

src/fsw/FCCode/Drivers/Piksi.cpp
```diff
@@ -271,32 +309,45 @@ unsigned char Piksi::read_all() {
...
+    }
+    last_bytes = 0;
+
+    buffer_begin = buffer;
+    buffer_end = buffer;
+    // copy buffer up to available
+    for (int i = 0; i < bytes; i++){
+        *(buffer_end++) = Serial4.read();
+    }
+    clear_bytes();
```
kylekrol, line 335: Also, just curious like what's goin on with this function here? Shouldn't the buffer already be clear? Wouldn't this only potentially wipe bytes we don't want to wipe?

kylekrol (changes requested):
We should really take the code time stamping the the main execute loop because it can't really do anything - like whatever it sets the timestamp to is just going to be what the interrupt set it to a tiny bit ago. Also not really sure what the `clear_bytes()` is doing/if it needs to be there?

If you feel like making the other book keeping changes go for it! Otherwise if it works it works haha

src/fsw/FCCode/Drivers/Piksi.cpp
```diff
@@ -271,32 +307,34 @@ unsigned char Piksi::read_all() {
     #ifdef DESKTOP
     return _read_return;
     #else
+
+    noInterrupts();
+
+    int bytes = Serial4.available();
+    last_bytes = 0;
```
kylekrol, line 314: Sorry I didn't notice this the first time but I think we should set `bytes` to the `last_bytes` seen by the interrupt. Otherwise, we could see data passed in between the interrupt and the start of this function call that doesn't get timestamped. It could also mess up timing for the next packet I think.
```suggestion
    int bytes = last_bytes;
    last_bytes = 0;
```

src/fsw/FCCode/Drivers/Piksi.cpp
```diff
@@ -271,32 +307,34 @@ unsigned char Piksi::read_all() {
...
-        delayMicroseconds(5);
-
-        if(micros()-initial_time >= READ_ALL_LIMIT){
-            clear_bytes();
-            return 5;
+    if(buffer_begin < buffer_end){ 
+        bool crc_error;
+        while(buffer_begin < buffer_end){
+            crc_error = process_buffer() < 0;
         }
```
kylekrol, line 337: Should we make this or equals @shihaocao? Unfamiliar with how Piksi works but it seems safer...
```suggestion
        bool crc_error = false;
        while(buffer_begin < buffer_end){
            crc_error |= process_buffer() < 0;
        }

```

src/fsw/FCCode/PiksiControlTask.cpp
```diff
@@ -7,23 +7,25 @@ PiksiControlTask::PiksiControlTask(StateFieldRegistry &registry,
     unsigned int offset, Devices::Piksi &_piksi) 
     : TimedControlTask<void>(registry, "piksi", offset),
     piksi(_piksi),
-    pos_f("piksi.pos", Serializer<d_vector_t>(6771000, 6921000, 28)),
-    vel_f("piksi.vel", Serializer<d_vector_t>(7570, 7685, 19)),
-    baseline_pos_f("piksi.baseline_pos", Serializer<d_vector_t>(0,2000,22)),
-    current_state_f("piksi.state", Serializer<unsigned char>(14)),
+    pos_f("piksi.pos", Serializer<d_vector_t>(0,100000,100)),
+    vel_f("piksi.vel", Serializer<d_vector_t>(0,100000,100)),
+    baseline_pos_f("piksi.baseline_pos", Serializer<d_vector_t>(0,100000,100)),
```
kylekrol, line 12: I don't think we want to change these serializer values here! Pretty sure @tanishqaggarwal set these to very specific values! Did we mean to do this or did things just get lost in the shuffle while merging?

madeleinemc: I see! I don't remember changing the values, and the changes don't make much sense to me, so I'll switch it back

src/fsw/FCCode/Drivers/Piksi.cpp
shihaocao: I think `|=` is the safer option because if any particular message has a crc_error, its probably safer to toss the whole read out.

src/fsw/FCCode/Drivers/Piksi.cpp
shihaocao: @madeleinemc , can we take out the `clear_bytes()` command and see if things still work?

src/fsw/FCCode/Drivers/Piksi.cpp
madeleinemc: src/fsw/FCCode/Drivers/Piksi.cpp:336:46: error: 'crc_error' may be used uninitialized in this function [-Werror=maybe-uninitialized]
             crc_error |= process_buffer() < 0;

kylekrol (approved):
Looks good to me - this is very exciting to have all wrapped up!

src/fsw/FCCode/Drivers/Piksi.cpp
```diff
@@ -331,16 +369,15 @@ void Piksi::clear_bytes() {
     #endif
     }
 
+// edit this
```
shihaocao, line 372: remove comment

src/fsw/FCCode/Drivers/Piksi.hpp
```diff
@@ -2,11 +2,13 @@
 #define PIKSI_HPP_
 
 #ifndef DESKTOP
+#include <Arduino.h>
```
shihaocao, line 5: is this necessary?

src/fsw/FCCode/Drivers/Piksi.hpp
```diff
@@ -336,10 +338,20 @@ class Piksi {
...
 
+    // check if bytes are entering the buffer
+    static void check_bytes();
+
+    // begin checking bytes in buffer at set interval
+    void start_interrupt();
+
+    // return time delay between now and time data entered buffer
+    unsigned long get_microdelta();
+
```
shihaocao, line 349: Change these to /** doc string comments?

src/fsw/FCCode/PiksiControlTask.cpp
```diff
@@ -7,23 +7,25 @@ PiksiControlTask::PiksiControlTask(StateFieldRegistry &registry,
...
+    vel_f("piksi.vel", Serializer<d_vector_t>(7570,7685,19)),
     baseline_pos_f("piksi.baseline_pos", Serializer<d_vector_t>(0,2000,22)),
     current_state_f("piksi.state", Serializer<unsigned char>(14)),
     fix_error_count_f("piksi.fix_error_count", Serializer<unsigned int>(1001)),
     time_f("piksi.time", Serializer<gps_time_t>()),
-    last_fix_time_f("piksi.last_fix_time"),
-    last_rtkfix_ccno_f("piksi.last_rtkfix_ccno")
+    microdelta_f("piksi.microdelta", Serializer<unsigned int>()),
+    last_rtkfix_ccno_f("piksi.last_rtkfix_ccno"),
+    no_bytes_available_f("piksi.bytes_available")
```
shihaocao, line 18: what's this field for? I don't think we're using it? We should remove it

shihaocao (changes requested):
All the microdelta stuff looks good to me. Just a few cosmetic changes to make this production-ready.

tanishqaggarwal:
Whenever I see the word "microdelta" I uncontrollably think about this SNL skit https://www.youtube.com/watch?v=vEA4gzG-ooo

tanishqaggarwal (commented):
(serious) I'd recommend making sure the propulsion system interrupt loop doesn't conflict with this one. Try running PropStateMachineCase on EDU.

kylekrol:
Something weird happen with the piksi control task files during merging it seemed like. Why does it think their are changes on every line? Did we go tabs to spaces or the other way around? Did we adjust indenting maybe? The diff shouldn't be that large...

tanishqaggarwal:
Might be a line endings thing.

PR #642, opened 2021-02-24
Move ADCS DCDC Control Logic to Mission Manager

shihaocao:
# Move ADCS DCDC Control Logic to Mission Manager

Fixes DCDC Checkout Case failures
- Also allows ADCS DCDC to be off during detumble (until the last cycle)

### Summary of changes
- DCDC is enabled in the last cycle of detumble, then it transitions to standby happens on the next cycle
- Now all test cases will be in detumble twice
- Updated unit tests to reflect ADCS DCDC being turned off on faults into safehold

### Testing
- TODO

### DIFF
--- a/src/fsw/FCCode/MissionManager.cpp
+++ b/src/fsw/FCCode/MissionManager.cpp
@@ -209,9 +209,10 @@ void MissionManager::dispatch_detumble()
 
     if (momentum <= threshold * threshold) // Save a sqrt call and use fro norm
     {
-        transition_to(mission_state_t::standby,
-                      adcs_state_t::point_standby);
-        adcs_dcdc_fp->set(true);
+        if(!adcs_dcdc_fp->get()) // cause a cycle where DCDC is turned on then wheels turn on
+            adcs_dcdc_fp->set(true);
+        else
+            transition_to(mission_state_t::standby, adcs_state_t::point_standby);
     }
 }
 
@@ -313,7 +314,7 @@ void MissionManager::dispatch_docking()
         have_set_docking_entry_ccno = false;
         transition_to(mission_state_t::standby,
                       adcs_state_t::point_standby);
-        adcs_dcdc_fp->set(true);
+        // ADCS dcdc should already be on, no need to re-assert
     }
 }
 
--- a/src/fsw/FCCode/MissionManager.cpp
+++ b/src/fsw/FCCode/MissionManager.cpp
@@ -406,13 +406,6 @@ void MissionManager::transition_to(mission_state_t mission_state,
         sph_dcdc_fp->set(true);
     }
 
-    if (mission_state == mission_state_t::standby)
-        adcs_dcdc_fp->set(true);
-    else if(mission_state == mission_state_t::safehold)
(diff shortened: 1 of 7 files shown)


src/fsw/FCCode/MissionManager.cpp
```diff
@@ -71,6 +71,7 @@ MissionManager::MissionManager(StateFieldRegistry &registry, unsigned int offset
     pressurize_fail_fp = find_fault("prop.pressurize_fail.base", __FILE__, __LINE__);
 
     sph_dcdc_fp = find_writable_field<bool>("dcdc.SpikeDock_cmd", __FILE__, __LINE__);
+    adcs_dcdc_fp = find_writable_field<bool>("dcdc.ADCSMotor_cmd", __FILE__, __LINE__);
```
kylekrol, line 74: You can use:

    FIND_WRITABLE_FIELD(bool, dcdc.ADCSMotor_cmd);

:) Don't have to be just a reminder if you're unaware!

src/fsw/FCCode/MissionManager.cpp
```diff
@@ -378,6 +383,12 @@ void MissionManager::set(sat_designation_t designation)
 void MissionManager::transition_to(mission_state_t mission_state,
                                    adcs_state_t adcs_state)
 {
+    if (mission_state == mission_state_t::standby)
+        adcs_dcdc_fp->set(true);
+    else if(mission_state == mission_state_t::safehold)
+        adcs_dcdc_fp->set(false);
+    // all other transitions shall leave the DCDC's alone
```
kylekrol, line 390: I would strongly suggest the following change:
```suggestion
   if(mission_state == mission_state_t::safehold)
        adcs_dcdc_fp->set(false);
    // all other transitions shall leave the DCDC's alone
```
Leaving this assertion that the DCDCs are enabled here is extra noise that can only serve to confuse us down the road. We want to strictly enforce that all transitions to standby that require booting the DCDC go through the detumble state otherwise initialization will fail. In my opinion adding this line here can only serve to make this fact tougher to decipher for those reading the code and lead to quite HITL failures that may not happen in HOOTL during mission rehearsals.

src/fsw/FCCode/MissionManager.cpp
```diff
@@ -395,6 +406,12 @@ void MissionManager::transition_to(mission_state_t mission_state,
...
 
+    if (mission_state == mission_state_t::standby)
+        adcs_dcdc_fp->set(true);
+    else if(mission_state == mission_state_t::safehold)
+        adcs_dcdc_fp->set(false);
+    // all other transitions shall leave the DCDC's alone
+
     set(mission_state);
     set(adcs_state);
     set(prop_state);
```
kylekrol, line 417:
```suggestion
    transition_to(mission_state, adcs_state);
    set(prop_state);
```

kylekrol (approved):
Looks good to me!

kylekrol:
Code was tested with the DCDC testcase. Full mission state transition HITL testing is being deferred till Friday after this was merged.

PR #645, opened 2021-02-26
Quakefault

18kudlugim:
# Quake Fault Handler Changes
Fixes #. (Optional; sometimes PRs don't have an associated ticket)

### Summary of changes
- Sets quake to disabled before powercycling, then to config once powercycling is over.
- No longer requires quake to be in wait for quake fault handler to powercycle device.

### Testing
Have to change unit testing to represent new changes bc we are not requiring quake to be in wait in order for power cycling to occur. The quake can be in any state before power cycling.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -7,6 +7,3 @@
 [submodule "lib/common/psim"]
 	path = lib/common/psim
 	url = https://github.com/pathfinder-for-autonomous-navigation/psim.git
-[submodule "lib/common/lin"]
-	path = lib/common/lin
-	url = https://github.com/pathfinder-for-autonomous-navigation/lin.git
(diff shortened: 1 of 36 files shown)


kylekrol (approved):
Looks good to me!

shihaocao (approved):
Everything in here looks good to me. Can we move this to a full PR and get it merged?

To clarify, it passes on HITL, but not on HOOTL?

18kudlugim:
Expected behavior for power cycling with these changes do not work in HOOTL but should work in HITL due to power cycle flags being extraneously reset in HOOTL.

shihaocao:
> Expected behavior for power cycling with these changes do not work in HOOTL but should work in HITL due to power cycle flags being extraneously reset in HOOTL.

Gotcha. When do we plan on HITLing?

PR #646, opened 2021-02-26
Add Reaction Wheel Frame Matrices

### LINKED ISSUE #282: Populate Leader and Follower Body Frame to "RWA Frame" Matrices

kylekrol:
Closes #282.

### Summary of Changes

* Populate reaction wheel to body and body to reaction wheel transformation matrices.

The current `+x` command mapped to a `+z` command, the current `+y` command mapped to a `-x` command, and the current `+z` command mapped to a `-y` command. Therefore, the transformation required to go from body frame to reaction wheel frame is:

    [ 0.0 -1.0  0.0 ]
    [ 0.0  0.0 -1.0 ]
    [ 1.0  0.0  0.0 ]

The inverse of this is just the transpose as it's an orthonormal matrix:

    [  0.0   0.0 1.0 ]
    [ -1.0   0.0 0.0 ]
    [  0.0  -1.0 0.0 ]

Note that the spacecraft were assembled identically and the mapping is the same for both spacecraft. This was verified by @kylekrol and @millie-schwartz.

### Original Notes

```
Leader:
 - rwa +x -> +z
 - rwa +y -> -x
 - rwa +z -> -y

Follower:
 - rwa +x -> +z
 - rwa +y -> -x
 - rwa +z -> -y
```


shihaocao (approved):
Same for both spacecraft?

kylekrol:
> Same for both spacecraft?

Indeed!

tanishqaggarwal:
Do we have data to support the determination? I think it would be good to review the data and have it linked to this PR just so we don't question our decisions in the future

millie-schwartz (approved):

kylekrol:
> Do we have data to support the determination? I think it would be good to review the data and have it linked to this PR just so we don't question our decisions in the future

Agree that data would be nice but at the same time what exactly would you be looking for? Like, the telemetry mapping is just as determined by us as the command mapping is?

Perhaps I could talk to @shihaocao and see how specific the ADCS checkout case examines data to potentially catch an issue in the mapping?

tanishqaggarwal:
I guess this comes down to how the mapping was determined. If it was just visual inspection then no point to data I guess. But if there was some ptest data that we could independently re-verify down the line if we ever doubt it, we could link that data now.

On the other hand, "down the line" is a rapidly decreasing interval, and this being one of the most important characterizations ever on the spacecraft, I imagine you would've double- and triple-checked this anyway. Historically, we don't tend to screw up physical verifications.

kylekrol:
> I guess this comes down to how the mapping was determined. If it was just visual inspection then no point to data I guess. But if there was some ptest data that we could independently re-verify down the line if we ever doubt it, we could link that data now.
> 
> On the other hand, "down the line" is a rapidly decreasing interval, and this being one of the most important characterizations ever on the spacecraft, I imagine you would've double- and triple-checked this anyway. Historically, we don't tend to screw up physical verifications.

I agree - this was all done via a visual inspection by both @millie-schwartz and I so I'm fairly confident in the results. I will be sure tho to verify next time I run a battery discharge test.

PR #647, opened 2021-02-26
adding real cycling

knk38:
# Changed code to actually cycle
This PR fixes some bugs in two checkout cases
### Summary of changes
- deleted incrementing cycle_no
- added self.cycle
-changed enums to not use depreciated getter

# Testing
Checkout cases were tested in HITL. The test logs should be pushed to ptest/logs

### DIFF
--- a/ptest/cases/adcs_checkout_case.py
+++ b/ptest/cases/adcs_checkout_case.py
@@ -283,7 +283,7 @@ class ADCSCheckoutCase(SingleSatOnlyCase):
 
         self.print_header("TORQUE TESTS: ")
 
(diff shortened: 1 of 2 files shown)


kylekrol (approved):
Looks great! If you can just leave a comment that it was HITL tested too for documentation purposes that'd be great.

18kudlugim (commented):
Did we want to wait for the full 900 cycles to complete for the power cycling check? I think 600 is probably good enough (for the gomspace checkout case).

kylekrol:
> Did we want to wait for the full 900 cycles to complete for the power cycling check? I think 600 is probably good enough (for the gomspace checkout case).

What exactly was the reason for changing the wait time from `600` to `900`? If changing it back would require another HITL test and this code passed on HITL, then my vote is to just stick with this.

PR #649, opened 2021-02-27
RadioSession dictionary

### LINKED ISSUE #629: Fix RadioSession Uplink Logic with AMC

afoarce:
# RadioSession dictionary for uplink fields

Fixes #629 

### Summary of changes
- Added dictionary to RadioSession to automatically update fields written from ptest and command prompt cases and  before sending uplink when the timer runs out
- Changed file names of uplinks created from HTTP endpoints to avoid ptest overriding manual inputs
- Added merging feature between fields written in http_uplink.json and the RadioSession dictionary before fields are written to uplink.json, giving priority to the endpoint commands when conflicts exist

### Testing
Tested in ptest by having the tlm server running with elasticsearch. While running a testcase using radios, used combinations of sending ws commands through ptest only, ptest and http endpoints, and http endpoint only to verify emails are sent as expected. Tested using adcs.state, potential issues exist when sending other types of fields such as booleans which will need to be further tested but is not directly related to the dictionary update.


ptest/configs/ci.json
```diff
@@ -12,8 +12,8 @@
         {
             "connected_device": "FlightController",
             "imei": "",
-            "connect": false,
-            "send_queue_duration" : 0,
+            "connect": true,
```
shihaocao, line 15: What does this do?

ptest/radio_session.py
```diff
@@ -137,23 +141,19 @@ def write_multiple_states(self, fields, vals, timeout=None):
...
-        t.start()
-
+        # make the uplink.json file so we know there is data to send and http_cmd
+        # can see it as well
+        if not os.path.exists("uplink.json"):   
+            with open('uplink.json', 'w'): pass    
+            t = threading.Thread(target=self.timer.start, name="Uplink timer thread")
+            t.start()
+
+        print(self.statefield_dict)
```
shihaocao, line 156: should this comment go?

ptest/radio_session.py
```diff
@@ -137,23 +141,19 @@ def write_multiple_states(self, fields, vals, timeout=None):
...
+            self.statefield_dict[fields[i]] = self.uplink_console.get_val(vals[i])
 
         # Start the timer. Timer will send uplink once after waiting for the
         # configured send queue duration.
-        t = threading.Thread(target=self.timer.start, name="Uplink timer thread")
-        t.start()
-
+        # make the uplink.json file so we know there is data to send and http_cmd
+        # can see it as well
+        if not os.path.exists("uplink.json"):
```
shihaocao, line 151: Should we return false if the file already exists?

ptest/radio_session.py
```diff
@@ -183,23 +183,31 @@ def uplink_queued(self):
                 'http://'+self.flask_server+':'+str(self.flask_port)+'/search-es',
                     params=payload, headers=headers)
 
-        if tlm_service_active and response.text.lower()=="true" and not os.path.exists("uplink.json"):
+        if tlm_service_active and response.text.lower()=="true" and not os.path.exists("uplink.json") and not os.path.exists("http_uplink.json"):
```
shihaocao, line 186: this change looks good, but it looks like we don't use this function anymore? Will it ever be used in the future?

ptest/radio_session.py
```diff
@@ -183,23 +183,31 @@ def uplink_queued(self):
...
+                self.statefield_dict[field] = queued_http_uplink[field]
+
+        #merged and updated fields to send
+        fields, vals = self.statefield_dict.keys(), self.statefield_dict.values()
+        with open('uplink.json', 'w') as telem_file:
+                    json.dump(self.statefield_dict, telem_file)
 
         # Create an uplink packet
-        success = self.uplink_console.create_uplink(fields, vals, "uplink.sbd") and os.path.exists("uplink.sbd")
+        success = self.uplink_console.create_uplink(fields, vals, "uplink.sbd", "uplink.http") and os.path.exists("uplink.sbd")
```
shihaocao, line 210: what is uplink.http, and what is in it?

shihaocao: Seems like we also refer to as the `json_filename` as an argument?

shihaocao (changes requested):
Almost looks good, just wondering about a print statement and a few questions?

ptest/configs/ci.json
fatimayousuf: It opens a radio session connected to the radio with that imei number. If connect is set to false, it won't open a radio session even if you put all that info in the config file

ptest/radio_session.py
fatimayousuf: @afoarce I think maybe you meant http_uplink.json here?

ptest/http_cmd.py
```diff
@@ -95,22 +95,31 @@ def send_telem():
...
-
+            if os.path.exists("http_uplink.json"):
+                # Organize the requested telemetry into a json object
+                requested_telem = {}
+                for field_val in uplink:
+                    requested_telem[field_val["field"]] = field_val["value"]
+
+                # Get the queued uplink
+                with open('http_uplink.json', 'r') as telem_file:
+                    queued_uplink = json.load(telem_file)
```
fatimayousuf, line 110: The queued uplink connected to the uplink timer is `uplink.json`, and the requested uplink sent over http is 'http_uplink.json', right? So shouldn't `queued_uplink` hold the json object from uplink.json? Otherwise, in in line 114, when you update `queued_uplink`, you're just updating the requested http uplink, which loses the telemetry queued to be sent when the timer ran out

fatimayousuf (dismissed):

ptest/http_cmd.py
afoarce: Currently, uplink.json is empty and just used as a flag for the existence of an uplink to send on the timer, since RadioSession holds all the data in a dictionary and adds the http_uplink data right before making and sending the uplink when the timer runs out. I think then we only need to update http_uplink since it is the only json file holding data?

ptest/radio_session.py
afoarce: I don't believe so - the point of the dictionary is that even if states have been written before we can continue doing so while the uplink timer is running. If an uplink doesn't exist yet, this statement makes the file and starts the timer.

ptest/radio_session.py
afoarce: That's a good point - the purpose for it that I saw was to make sure we don't overwrite uplinks when we could only write states once, but it says it can be used by ptest to determine whether to send uplinks autonomously. Does it actually tell us if Iridium received the last uplink, and should we add it back, or is it only checking that the service is active and we have an uplink to send?

ptest/radio_session.py
afoarce: I think this one should be uplink.json, the file just created with the merged data

shihaocao (approved):

PR #651, opened 2021-03-01
Add QuakeManager init()

shihaocao:
Before opening the PR, think: 
- Is this PR bulletproof? Can I think of no ways to improve this PR beyond its current state?
- Is the PR short enough to ensure a quick but thorough review, but long enough to represent a complete feature? For large state machines, let's try opening PRs for only a few states at a time.

If the answer to either of these questions is "no", refactor your code and open your PR when the answer to both questions is "yes".

--------------------------

# Quake Init

This PR is necessary to fix the issue of QuakeManager not interesting to the next packet within a snapshot due to max Downlink fields not being initialized after DownlinkProducer being initialized

### Summary of changes
- Updated QuakeManager unit tests
- Updated MCL initialization order
- Future use cases of QuakeManager must additionally call the `init()` function.

### Testing
- This PR moves downlinking code in the right direction on HOOTL tests at least since telem/parsetelem shows moving to the next packet now.


kylekrol (approved):
Looks good to me!

tanishqaggarwal (approved):

PR #659, opened 2021-03-09
Renamed Reserved Symbol

kylekrol:
### Summary of changes

- Symbols matching `__.*` or `_[A-Z].*` are reserved for the implementation and therefore are considered undefined behavior if used in user level code; therefore, `__field_creator` was renamed to `_field_creator`.

### Testing
NA

### Constants
NA

### Telemetry
NA

### Documentation Evidence
NA


shihaocao (approved):
big scary

PR #666, opened 2021-03-12
Microdelta check

### LINKED ISSUE #663: Microdelta verification
Mark piksi data as invalid if microdelta is greater than 100 ms

madeleinemc:
Closes #663

Add a microdelta threshold; throws CRC error if microdelta is above threshold.
PiksiCheckoutCase performs normally when threshold set to 100ms. Throws CRC errors over half the time when threshold set to 50ms (as expected).


kylekrol:
@madeleinemc want to merge in or rebase off master so those previous Piksi commits aren't in here?

madeleinemc:
I did that afterward as part of the above commit :/

src/fsw/FCCode/PiksiControlTask.cpp
```diff
@@ -44,6 +46,10 @@ void PiksiControlTask::execute()
     int microdelta = piksi.get_microdelta();
     microdelta_f.set(microdelta);
 
+    //Throw CRC error if microdelta is not in expected range
+    int md_threshold = 100000; // 100 ms
+    if (microdelta < 0 || microdelta > md_threshold) read_out = 3;
```
kylekrol, line 51: Also we should be careful here with comparing `signed` versus `unsigned` integers -- it can get weird depending on the situation. Iirc `microdelta` is signed so we only need to check that it's less than a threshold. It'd also probably be best PAN practice to make the threshold a tracked constant within flight software, something like:

    TRACKED_CONSTANT_SC(unsigned int, PIKSI_MD_THRESHOLD, 100000);
    ...

kylekrol, line 51: Lastly, referring to the Piksi mode enum directly rather than hard coding a three is probs a could idea in case the piksi modes get shifted around -- it'll be more robust to code changes going forward.

shihaocao: I think kyle means that `microdelta` is unsigned (not signed), and thus we only need to check it's below a certain threshold. I agree that the tracked constant is useful.

shihaocao (changes requested):
PR looks good to me at a high level :) We should remove the psim changes tho, and we should also implement Kyle's suggestions

src/fsw/FCCode/PiksiControlTask.cpp
madeleinemc: The read_out result is a little different from the actual index of the respective state in piksi_mode_enum. read_out has 6 possible values while piksi has 14 modes

madeleinemc: Will do

shihaocao: Oh that's right, i forgot the read out values are hardcoded, that sucks

shihaocao: my bad lul

kylekrol: Cool and yeah sorry I meant unsigned -- update the comment.

kylekrol: Ooohhh aight then yeah never mind -- hard coded it is!

kylekrol:
> PR looks good to me at a high level :) We should remove the psim changes tho, and we should also implement Kyle's suggestions

There are also still the phanton commits that ended up getting squashed in the previous Piksi related PR -- perhaps updating the branch below will fix it?

shihaocao (approved):
Good to go!

PR #667, opened 2021-03-13
Quake Fault Handler ptest case

18kudlugim:
- replaces quakefaulthandler.py PTest case
- I am unsure whether the power cycling behavior translates to HITL within the same cycle limit/timeframe as in HOOTL. It might take some more time


ptest/cases/quake_power_cycling.py
```diff
@@ -0,0 +1,131 @@
...
+    def sim_model(self):
+        return SingleAttitudeOrbitGnc
+
+    @property
+    def sim_mapping(self):
+        return "ci_mapping.json"
+
+    @property
+    def sim_duration(self):
+        return float("inf")
```
shihaocao, line 28: Looks like we got simulation active in this test case, I don't think its necessary right?

ptest/cases/utils.py
```diff
@@ -138,12 +138,21 @@ class Enums(object):
...
 
     radio_states = FSWEnum([
-        "disabled",
         "wait",
         "transceive",
         "read",
         "write",
-        "config"
+        "config",
+        "disabled"
```
shihaocao, line 146: wow, thanks for catching this, would probably screw us over later in debugging

ptest/cases/quake_power_cycling.py
```diff
@@ -0,0 +1,131 @@
...
+        self.read_state("qfh.state")
+        self.read_state("radio.last_comms_ccno")
+        self.read_state("gomspace.power_cycle_output3_cmd")
+
+    def run_case_singlesat(self):
+        # The satellite has been in a blackout since startup. Cycle count starts at 1.
+        self.cycles_since_blackout_start = self.rs("pan.cycle_no") - 1
+
+        # These steps necessary in HOOTL, unsure about how it will affect HITL
+        self.ws("radio.state", 0)
```
shihaocao, line 106: What is 0? Can we change this to a radio state enum ref?

ptest/cases/quake_power_cycling.py
```diff
@@ -0,0 +1,131 @@
...
+    def check_radio_in_config(self):
+        if not self.rs("radio.state") == Enums.radio_states["config"]:
+            raise TestCaseFailure("Radio should be in config after QuakeFaultHandler power cycled.")
+
+    def is_powercycle_state(self):
+        return self.rs("qfh.state") in [Enums.qfh_states["powercycle_1"],Enums.qfh_states["powercycle_2"] ,Enums.qfh_states["powercycle_3"]]
+
+    # Cycle until the QFH transitions to the next state
+    def advance_to_next_qfh_state(self,time):
+        while True:
```
shihaocao, line 56: can we change this to a for loop just so its a bit more clear?

ptest/cases/quake_power_cycling.py
```diff
@@ -0,0 +1,131 @@
...
+        # Check second condition valid in HITL
+        if (powercycled and not should_powercycle) or (should_powercycle and not powercycled):
+            raise TestCaseFailure("QuakeFaultHandler failed to power cycle the output channel.")
+       
+
+    def diagnostics(self):
+        self.read_state("radio.state")
+        self.read_state("qfh.state")
+        self.read_state("radio.last_comms_ccno")
+        self.read_state("gomspace.power_cycle_output3_cmd")
```
shihaocao, line 99: did you want these to be `print_rs` calls?

shihaocao: otherwise these readstates wont do anything since ptest doesn't have a caching system

ptest/cases/quake_power_cycling.py
```diff
@@ -0,0 +1,131 @@
...
+            # Check power cycle behavior
+            self.check_correct_powercycle_behavior(time)
+
+            if (self.cycles_since_blackout_start >= time):
+                break
+
+
+    def check_correct_powercycle_behavior(self,time):
+        powercycled = self.check_powercycle()
+        output = self.rs("gomspace.output.output3")
```
shihaocao, line 74: seems like this is unused

ptest/cases/quake_power_cycling.py
```diff
@@ -0,0 +1,131 @@
+# Empty test case. Gets cycle count purely for diagnostic purposes
+from .base import SingleSatOnlyCase
+from psim.sims import SingleAttitudeOrbitGnc
+from .utils import Enums, TestCaseFailure
+
+
+class QuakePowerCycling(SingleSatOnlyCase):
+    @property
+    def sim_duration(self):
+        return float("inf")
```
shihaocao, line 10: probably remove this entirely

shihaocao (changes requested):
Thanks for getting this out so fast! I think there are a few small things to fix (outlined in comments)

I guess the one feature I think this is still missing is checking that the QauekFaultHanlder suggests the right `fault_response_t`, I think we should be stepping flight software forward, and always check that the right fault response was raised by seeing if MissionManager switches to the correct states

ptest/cases/quake_power_cycling.py
18kudlugim: yea I had a comment underneath about a section that prints the gomspace output channel (for power cycling testing in HITL), but was unsure how it'll translate. As is, the checkout case essentially power cycles immediately in HOOTL and checks this behavior, but power cycling may lag a bit in HOOTL and I wasn't sure whether to add this because of that.

ptest/cases/quake_power_cycling.py
18kudlugim: I think it's a bit noisy to print every cycle. I mainly had the read_states so the values could be accessed for plotting

ptest/cases/utils.py
tanishqaggarwal: Agreed. Also, i've confirmed it's in the same order as the corresponding fsw enum file

shihaocao (changes requested):
Looks good on the MissionManager checks, I just had one more requested change sorry

It seems like given that we modified the QuakeFaultHandler to also mutate quake state, could we check that the quake state gets mutatued properly? (ex disabled when disabled, config when config)

ptest/cases/quake_power_cycling.py
```diff
@@ -0,0 +1,124 @@
...
+    # Cycle until the QFH transitions to the next state
+    def advance_to_next_qfh_state(self,time):
+        for i in range(time):
+            self.cycle()
+            self.diagnostics()
+            self.cycles_since_blackout_start += 1
+
+            # Check mission state when in QFH forced standby
+            if self.qfh_state() == Enums.qfh_states['forced_standby'] and not self.mission_state == "standby":
+                    raise TestCaseFailure(f"QuakeFaultHandler did not force satellite into standby during during QuakeFaultHandler's `forced standby` state. State was: {self.mission_state}. Current control cycle: {self.rs('pan.cycle_no')}")
```
shihaocao, line 47:
```suggestion
                raise TestCaseFailure(f"QuakeFaultHandler did not force satellite into standby during during QuakeFaultHandler's `forced standby` state. State was: {self.mission_state}. Current control cycle: {self.rs('pan.cycle_no')}")
```

ptest/cases/quake_power_cycling.py
```diff
@@ -0,0 +1,124 @@
...
+
+        # These steps necessary in HOOTL, unsure about how it will affect HITL
+        self.ws("radio.state", Enums.radio_states["wait"])
+        self.ws("gomspace.power_cycle_output3_cmd",True)
+
+        # Simulate one day of no comms
+        self.logger.put(f"Creating a comms blackout of 24 hours, starting on control cycle: {self.rs('pan.cycle_no')}")
+        self.advance_to_next_qfh_state(self.one_day_ccno)
+        if not self.mission_state == "standby":
+                    raise TestCaseFailure(f"QuakeFaultHandler did not force satellite into standby after 24 hours of no comms. State was: {self.mission_state}. Current control cycle: {self.rs('pan.cycle_no')}")
```
shihaocao, line 93:
```suggestion
            raise TestCaseFailure(f"QuakeFaultHandler did not force satellite into standby after 24 hours of no comms. State was: {self.mission_state}. Current control cycle: {self.rs('pan.cycle_no')}")
```

shihaocao (approved):
nice

PR #673, opened 2021-03-15
Automatic ElasticSearch Index in MCT from config file argument

### LINKED ISSUE #668: MCT Dynamic IMEI number from statefield report
Current:
It's really annoying to change a .js file every time we have a new IMEI number during testing

Fix:
Have MCT pull from a config file to set the IMEI number on request to streamline dual HOOTL/HITL tests

Duncan-McD:
# Configuration for MCT ES Indexes (Part One: Single Sat)

Fixes #668 

### Summary of changes
- Added the ability to send a configuration file path to the telemetry.js file from the main npm start script like this:
`npm start ptest/configs/dual_usb.json`
- path must be relative to Flight Software (such as ptest/configs/dual_usb.json)
- it can be a file with just usbs or a file with just radios it does not matter
- if it has neither usb or radio in given file, it defaults to the same index as ci.json's index
- default path (if no path argument is entered is ptest/configs/ci.json)

Note: at the moment this is only single sat configuration (because all that is in master is single mct), however looking forward this same config variable passed in can be used to indicate to mct whether you want it to run with 2 satellites or 1 satellite (depending on the number of radios or usbs in the config file) -> this will be implemented into the dual_mct_isolated branch or a seperate branch after that is merged in

### Testing

Tested with many different configuration files and imei numbers to see if it automatically checks the correct indexes and it does. This was verified by viewing the data in OpenMCT

### Constants

N/A

### Telemetry

N/A

### Documentation Evidence

- Added information to the MCT README to explain the usage of the configuration argument
- Added a few relevent in-line comments

### DIFF
--- a/MCT/server-files/server.js
+++ b/MCT/server-files/server.js
@@ -1,6 +1,10 @@
 /**
  * Basic implementation of a history and realtime server for MCT.
  */
+
+//get arguments
+var myArgs = process.argv.slice(2);
+
 var Telemetry= require('./telemetry');
 var RealtimeServer = require('./boilerplate-mct-servers/realtime-server');
 var HistoryServer = require('./boilerplate-mct-servers/history-server');
@@ -16,7 +20,12 @@ var credentials = {key: privateKey, cert: certificate};
 var app = express()
 expressWs(app);
 
-var spacecraft = new Telemetry();
+//sets default config file to ci.json
+if(myArgs[0] == undefined){
+  myArgs[0] = "ptest/configs/ci.json"
+}
+
+var spacecraft = new Telemetry(myArgs[0]);
 var realtimeServer = new RealtimeServer(spacecraft);
 var historyServer = new HistoryServer(spacecraft);
 
--- a/MCT/server-files/telemetry.js
+++ b/MCT/server-files/telemetry.js
@@ -21,53 +21,53 @@ var searchURl = 'http://localhost:5000/search-es';
  */
 function Telemetry(configuration) {
 
-    var FlightSoftware = path.resolve(__dirname, '../..')
-    var config_file = FlightSoftware + "/" + configuration
-    var config_json = require(config_file);
-    /**
-    * The index of the Elastic Search database
-    */
-    if(config_json.devices.length>0){
-      this.searchIndex = 'statefield_report_' + config_json.devices[0].imei
-    }
-    else if(config_json.radios.length>0){
-      this.searchIndex = 'statefield_report_' + config_json.radios[0].imei
-    }
-    else{
-      this.searchIndex = 'statefield_report_'
-    }
-    //This state function takes in initial values from the state-variables.js file
-    this.initialState = variables;
+  var FlightSoftware = path.resolve(__dirname, '../..')
+  var config_file = FlightSoftware + "/" + configuration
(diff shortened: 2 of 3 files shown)


MCT/server-files/telemetry.js
```diff
@@ -9,20 +9,33 @@ Then it adds it to the hisorical and realtime servers through the history and li
...
+    */
+    if(config_json.devices.length>0){
+      this.searchIndex = 'statefield_report_' + config_json.devices[0].imei
+    }
+    else if(config_json.radios.length>0){
+      this.searchIndex = 'statefield_report_' + config_json.radios[0].imei
+    }
+    else{
+      this.searchIndex = 'statefield_report_'
+    }
```
shihaocao, line 38: this would just break everything right? since no imei postfix is provided?

If this branch should never be nominally taken, we should throw an error instead

MCT/server-files/server.js
```diff
@@ -16,7 +20,12 @@ var credentials = {key: privateKey, cert: certificate};
 var app = express()
 expressWs(app);
 
-var spacecraft = new Telemetry();
+//sets default config file to ci.json
+if(myArgs[0] == undefined){
+  myArgs[0] = "ptest/configs/ci.json"
+}
```
shihaocao, line 26: does ci have an imei number by default?

Duncan-McD: ci doesn't have an imei number so elasticsearch uses statefield_report + ""

MCT/server-files/telemetry.js
Duncan-McD: Yea that is fair - To be honest I initially thought that if the config file was missing devices and radios we could just have it defualt to what ci's index is (statefield_report_) but your right, error would be more robust here

shihaocao (commented):
To clarify does this only work for single sat configs rn?

MCT/server-files/server.js
shihaocao: Oh so it would work then?

MCT/server-files/telemetry.js
shihaocao: well if this statefield_report would actually work with ci, then I guess this is okay? I just wanted to check

Duncan-McD:
> To clarify does this only work for single sat configs rn?

Yep but it allows for easy implementation of dual v single when dual_mct is merged in

MCT/server-files/server.js
Duncan-McD: Yea it works (I tested with ci.json too)

MCT/server-files/telemetry.js
Duncan-McD: Yea statefield_report_ does work as per ci.json but u might be right about throwing error being better anyways?

Duncan-McD:
oh wait no i misread that - no it can use any config file cuz it just takes first device or first radio rn but it only configures for one sat rn

MCT/server-files/telemetry.js
Duncan-McD: I changed it to throw an error if there are no devices or radios in the config file.

shihaocao (approved):

shihaocao (approved):

PR #677, opened 2021-03-17
New Data visualization plugins and Object Persistence in MCT

### LINKED ISSUE #571: Add Layout support to MCT
### Layouts need to be added to OpenMCT to customize telemetry visualization

This can be accomplished through OpenMCTs [CouchDB](https://github.com/nasa/openmct/tree/master/platform/persistence/couch) plugin or OpenMCTs [Elasticsearch](https://github.com/nasa/openmct/tree/master/platform/persistence/elastic) plugin which hosts all locally added domain objects and layouts made in the My Items folder in the linked database. 

This plugin could possibly be modified to use a different database service as well.

Duncan-McD:
# New Data visualization plugins and Object Persistence in MCT

Fixes #571 

### Summary of changes

1. Installs the Elasticsearch plugin from openmct to allow persistence in user created objects
- This will allow us to store layouts on the server once GSW is deployed and view them in our browsers on different machines
- CORS had to be enabled in order for openmct's plugin to work. Essentially CORS  (or Cross-Origin Resource Sharing) allows to servers to share information between eachother and this is what OpenMCT needs in order for it to store layouts in elasticsearch using the plugin

2. Bash script has been added which adds the two necessary configuration lines into the elasticsearch.yml file 
- it puts it in the config lines in the default location on each operating system but if you installed elasticsearch a different way the directory may be different for you
- The bash script enables http cors for elastic search and then lets the incoming origin be localhost (which allows for safety instead of allowing any incoming CORS requests)
- This means during testing on our own machines if we want to use persistent object and layout creation, we will have to run this script and add the configuration to elasticsearch

Note: once we get the GSW up and running on server, if the elasticsearch database will be running on the same machine as OpenMCT's server then it should be fine to leave the origin setting to be localhost but if elasticsearch and openmct are hosted on different servers then we will have to specify a different origin in the elasticsearch.yml file.

3. installed the Summary Widget Plugin from OpenMCT
-  Let's you add a layout that lets you set what the nominal conditions of a piece of telemetry is and display different colors or messages on that layout depending on the values of the telemetry

4. installed the Auto Flow View plugin from OpenMCT
- Let's you add a layout that lets you view the most recent value of a set of telemetry points or the most recent within a certain time range.

### Testing

I ran all the plugins on my machine and was able to create multiple layouts using them with incoming telemetry streams

I also was able to create a layout in one browser then switch to another browser and view that same layout (indicating working persistence)

The script successfully configures the elasticsearch.yml on linux. However, I was unable to test on mac and windows. But, according to elasticsearch the paths in the script for windows and mac are the default for the elasticsearch.yml file so since the script logic is the same as it is for linux then it should work given the location of the config file is in the correct spot.

### Constants

N/A

### Telemetry

N/A

### Documentation Evidence

I updated the MCT/generators Readme file as well as added in line comments in the index.html file and the bash script


shihaocao (approved):
nice

PR #686, opened 2021-03-28
quick fix

fatimayousuf:
# Remove static cast method

### Summary of changes
- Removed a method that was preventing us from building flight software


kylekrol:
Nice!

kylekrol (approved):

PR #692, opened 2021-03-31
Enum Print

shihaocao:
# PTest Enum Print Support

### Summary of changes
- Enum printing is now supported so long as it is defined as appropriate in `utils.py`
- If in the ptest terminal, you try to read a field that doesn't exist, you'll get a slightly more verbose message now.
- Tested and working

### DIFF
--- a/ptest/cmdprompt.py
+++ b/ptest/cmdprompt.py
@@ -114,11 +114,17 @@ class StateCmdPrompt(Cmd):
             return
 
         start_time = timeit.default_timer()
-        read_result = self.cmded_device.read_state(args[0])
+        try:
+            read_result = self.cmded_device.smart_read(args[0])
+        except NameError:
+            read_result = None
+            print('Field with that name does not exist!')
+            return 
+            
         elapsed_time = int((timeit.default_timer() - start_time) * 1E6)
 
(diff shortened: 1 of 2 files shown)


ptest/cmdprompt.py
```diff
@@ -114,11 +114,17 @@ def do_rs(self, args):
             return
 
         start_time = timeit.default_timer()
-        read_result = self.cmded_device.read_state(args[0])
+        try:
+            read_result = self.cmded_device.smart_read(args[0])
```
shihaocao, line 118: smart_read is the way to go :)

ptest/cmdprompt.py
```diff
@@ -114,11 +114,17 @@ def do_rs(self, args):
...
+        except NameError:
+            read_result = None
+            print('Field with that name does not exist!')
+            return 
+            
         elapsed_time = int((timeit.default_timer() - start_time) * 1E6)
 
         try:
-            human_readable_result = Enums()[args[0]]
+            human_readable_result = Enums()[args[0]][read_result]
```
shihaocao, line 127: lmao

kylekrol (approved):

PR #699, opened 2021-04-03
Remove Satellite Designation from Flows and from EEPROM

shihaocao:
# Remove Satellite Designation from Flows and from EEPROM

This field does not need to be EEPROM saved, and does not need to be continuously downlinked. Its information is only relevant if the mission state is in leader or in follower. It is essentially just a commanding field now.

### Summary of changes
- Remove sat designation from EEPROM saved fields
- Remove sat designation from EEPROM tests as a result of the above bullet
- Remove sat designation from Flows
- Clear sat designation on fault handling


src/fsw/FCCode/MissionManager.cpp
```diff
@@ -16,19 +16,19 @@ const constexpr std::array<mission_state_t, 5> MissionManager::fault_responsive_
...
+docking_timeout_limit_f("docking_timeout_limit",
+                        Serializer<unsigned int>(0, 2 * PAN::one_day_ccno, 6)),
+adcs_state_f("adcs.state", Serializer<unsigned char>(10)),
+docking_config_cmd_f("docksys.config_cmd", Serializer<bool>()),
+enter_docking_cycle_f("docksys.enter_docking"),
+mission_state_f("pan.state", Serializer<unsigned char>(12), 1),
+is_deployed_f("pan.deployed", Serializer<bool>(), 1000),
+deployment_wait_elapsed_f("pan.deployment.elapsed", Serializer<unsigned int>(15000), 500),
+sat_designation_f("pan.sat_designation", Serializer<unsigned char>(2)),
+enter_close_approach_ccno_f("pan.enter_close_approach_ccno")
```
shihaocao, line 31: changing spacing

tanishqaggarwal (approved):

Issue #683, opened 2021-03-24
Kill Switch Upgrade to Mission Manager


shihaocao:
Need to implement a way such that on command, the spacecraft shuts down the radio system, and no longer attempts to make contact again with the ground. Perhaps make a state to transition into in MissionManager.

Also need to save something to EEPROM so that it no longer tries to make contact with the ground.

shihaocao:
Potentially, just pigeonhole the state machine in a startup with a `while(true)` loop.

Additionally, make sure the check condition isn't actually a True false, but perhaps a `while(byte = magic_byte)` loop. This is so that a single bit corruption doesn't screw us over.

PR #705, opened 2021-04-06
Kill switch

angelaloh09:
# Kill Switch Upgrade to Mission Manager 

Fixes:  #683

### Summary of changes
- Created kill_switch state field. Write 127 to kill switch field so spacecraft no longer tries to make contact with ground. 
- Placed the check for kill switch and while-true loop under dispatch_startup(), so every time the spacecraft reboots, it will check the kill switch flag in startup mode.  

### Testing
- Ran a HOOTL test with PsimDebug, set kill switch flag to 127 with `ws kill_switch 127`, and sent ptest to not respond/error out because of the while true loop when running PsimDebug again. 
- Ran a HITL test with the EduSat with EmptyCase. First, uncommented line 129 of MainControlLoop `eeprom_controller.execute_on_time()` to execute the eeprom_controller, which was original commented out to save write cycles. Then, read the initial kill_switch flag, which returned 0. Wrote the kill_switch flag to 127. Waited for 200 cycles. (Current cycle update for kill_switch is 100). Reran EmptyCase, the ptest terminal errored out because of the while true loop (screenshot attached, titled "Image 1"). To double-check the EduSat can function properly with the kill_switch code, we changed the checked flag value to 126 `if (kill_switch_f.get() == 126)`, and the ptest returned back to normal without any errors. 

Screen Shot 2021-04-03 at 2 48 59 PM

_Image 1: Ptest errored out when re-running EmptyCase (after setting kill switch flag to 127._ 

Screen Shot 2021-04-03 at 2 50 35 PM
_Image 2: Set flag check value to 126, reran EmptyCase, Ptest worked without erroring out._

### Documentation Evidence
- Inline documentation is sufficient. 
- Before the kill_switch flag check, wrote "Step 0. If kill switch flag is set, shuts down radio connection." Inside the while true loop, commented "This will cause gomspace to reboot the spacecraft and we will get stuck in this loop forever." The inline documentation is sufficient because it clearly states what the kill_switch flag is for do, how to set it to either shut down or not shut down the radio connection, and the effects of enabling the kill_switch.


ptest/usb_session.py
```diff
@@ -532,11 +532,11 @@ def disconnect(self):
...
 
-        self.http_thread.terminate()
-        self.http_thread.join()
+        # self.http_thread.terminate()
+        # self.http_thread.join()
 
-        self.http_thread.terminate()
-        self.http_thread.join()
+        # self.http_thread.terminate()
+        # self.http_thread.join()
```
shihaocao, line 539: let's remove all the changes on this file for now, and we can address this issue in a more permanent fashion this week

src/fsw/FCCode/EEPROMControllerTeensy.cpp
```diff
@@ -21,7 +21,10 @@ void EEPROMController::read_EEPROM(){
 }
 
 void EEPROMController::update_EEPROM(unsigned int position){
-  EEPROM.put(addresses[position], _registry.eeprom_saved_fields[position]->get_eeprom_repr());
+
+  while (EEPROM.read(addresses[position]) != _registry.eeprom_saved_fields[position]->get_eeprom_repr()){
```
shihaocao, line 25: we should only attempt it once, would an `if` statement work instead?

We don't want to get stuck in a while loop if EEPROM is laggy or locks up

src/fsw/FCCode/MissionManager.cpp
```diff
@@ -105,11 +107,7 @@ void MissionManager::execute()
             transition_to(mission_state_t::safehold, adcs_state_t::startup, prop_state_t::disabled);
             return;
         }
-        else if (fault_response == fault_response_t::standby
-            && state != mission_state_t::safehold
-            && state != mission_state_t::initialization_hold
-            && state != mission_state_t::detumble
-            && state != mission_state_t::standby)
+        else if (fault_response == fault_response_t::standby && state != mission_state_t::safehold && state != mission_state_t::initialization_hold && state != mission_state_t::detumble && state != mission_state_t::standby)
```
shihaocao, line 110: let's undo this line change? I think the old formatting was good

src/fsw/FCCode/MissionManager.hpp
```diff
@@ -194,6 +194,12 @@ class MissionManager : public TimedControlTask<void>
      */
     InternalStateField<unsigned int> enter_close_approach_ccno_f;
 
+    /**
+     * @brief The state at which we want to shut down all communication 
+     * from spacecraft with ground
```
shihaocao, line 199: Perhaps this comment should be revised to talk about the switch as a command, rather than the specific state?

shihaocao (changes requested):
Solid code changes, just some quick formatting items, and one question about the while loop

src/fsw/FCCode/MissionManager.cpp
```diff
@@ -169,8 +167,18 @@ bool MissionManager::check_adcs_hardware_faults() const
 
 void MissionManager::dispatch_startup()
 {
+    // Step 0. If kill switch flag is set, shuts down radio connection
+    if (kill_switch_f.get() == 127)
```
shihaocao, line 171: Let's put this as a tracked constant. Perhaps look at how many other constants are wrapped in TRACKED_CONSTANT xxxxx etc

src/fsw/FCCode/MissionManager.cpp
```diff
@@ -29,7 +29,8 @@ MissionManager::MissionManager(StateFieldRegistry &registry, unsigned int offset
       is_deployed_f("pan.deployed", Serializer<bool>(), 1000),
       deployment_wait_elapsed_f("pan.deployment.elapsed", Serializer<unsigned int>(15000), 500),
       sat_designation_f("pan.sat_designation", Serializer<unsigned char>(2), 1),
-      enter_close_approach_ccno_f("pan.enter_close_approach_ccno")
+      enter_close_approach_ccno_f("pan.enter_close_approach_ccno"),
+      kill_switch_f("pan.kill_switch", Serializer<unsigned char>(3))
```
shihaocao, line 33: This code seems incomplete, there's no EEPROM save duration constructor call?

Additionally, what is a good EEPROM save duration? Perhaps once every 100 cycles?

src/fsw/FCCode/EEPROMControllerTeensy.cpp
kylekrol: Don't we not want to include EEPROM changes in this PR?

src/fsw/FCCode/MissionManager.hpp
```diff
@@ -194,6 +194,13 @@ class MissionManager : public TimedControlTask<void>
      */
     InternalStateField<unsigned int> enter_close_approach_ccno_f;
 
+    /**
+     * @brief The command to shut down all communication 
+     * from spacecraft with ground when set to 127.
+     */
+    WritableStateField<unsigned char> kill_switch_f;
+    TRACKED_CONSTANT_SC(unsigned char, kill_switch_value, 0);
```
shihaocao, line 202: we should have this as 127

shihaocao: the magic byte value

src/fsw/FCCode/MissionManager.cpp
shihaocao: and then use the tracked constant here

shihaocao (changes requested):
Looks good! Lets make sure to use the tracked constant, and unstage the changes to psim

src/fsw/FCCode/MissionManager.cpp
```diff
@@ -28,8 +29,10 @@ MissionManager::MissionManager(StateFieldRegistry &registry, unsigned int offset
       mission_state_f("pan.state", Serializer<unsigned char>(12), 1),
       is_deployed_f("pan.deployed", Serializer<bool>(), 1000),
       deployment_wait_elapsed_f("pan.deployment.elapsed", Serializer<unsigned int>(15000), 500),
-      sat_designation_f("pan.sat_designation", Serializer<unsigned char>(2)),
-      enter_close_approach_ccno_f("pan.enter_close_approach_ccno")
+      sat_designation_f("pan.sat_designation", Serializer<unsigned char>(2), 1),
+      enter_close_approach_ccno_f("pan.enter_close_approach_ccno"),
+      kill_switch_f("pan.kill_switch", Serializer<unsigned char>(3), 100)
```
kylekrol, line 34: Two things here:

 * Why did we add the `sat_designation_f` field to EEPROM? I thought @shihaocao had gotten rid of this?
 * The serializer for the kill switch right now can only take on values from `0-3` which means we'd never be able to uplink the required kill switch value.

ptest/usb_session.py
```diff
@@ -19,6 +19,7 @@
 from .cases.utils import str_to_val
 import lin
 
+
 class USBSession(object):
```
kylekrol, line 23: Are any of the changes here required for the PR? If not we should probably just avoid changes and keep the diff for this PR as simple as possible.

kylekrol (approved):
lgtm!

PR #707, opened 2021-04-06
Eeprom put vs update

samuelqian01:
_EEPROM Update revision to save write cycles_

**Fixes:** 

Revised update_EEPROM method to include a conditional statement that checks if memory has been written to already with a particular state. If it has, memory does not update to save write cycles. Updated statefieldregistry to clean up the assert statement to better support update_EEPROM.

**Summary of changes**

- Added if statement that checks if the what you are writing is the same as what is already saved in memory. If so, don't update the EEPROM. If not, update the EEPROM.
- Changed assert statement that was causing errors with EEPROM functionality due to logical error

**Testing**
Tested with hardware in the loop. Checked by adding a print statement to whenever the EEPROM updates and used ptest to write states that were initialized, the same as the previous write, and different from the previous write. When initialized and writing a state different from the previous state, 'EEPROM Updated' prints out. When the same as previous write, nothing prints out to the console, proving functionality of update_EEPROM.

**Documentation Evidence**

Screen Shot 2021-04-05 at 8 07 31 PM

Image below contains some quake errors which were from the quake radio not being connected to the Teensy which I fixed later in the image above. Added for additional evidence.
Screen Shot 2021-04-05 at 8 03 00 PM


src/fsw/FCCode/EEPROMControllerTeensy.cpp
```diff
@@ -6,27 +6,38 @@
...
     _registry.eeprom_saved_fields[i]->set_from_eeprom(field_val);
   }
 }
 
 void EEPROMController::update_EEPROM(unsigned int position){
-  EEPROM.put(addresses[position], _registry.eeprom_saved_fields[position]->get_eeprom_repr());
+  
+  if (EEPROM.read(addresses[position]) != _registry.eeprom_saved_fields[position]->get_eeprom_repr()){
+    EEPROM.update(addresses[position], _registry.eeprom_saved_fields[position]->get_eeprom_repr());
+    debug_console::printf("\nEEPROM updated\n");
```
shihaocao, line 31: lets remove the debug printf for the code that's going on master

test/StateFieldRegistryMock.hpp
```diff
@@ -21,8 +21,8 @@ class StateFieldRegistryMock : public StateFieldRegistry {
         static std::shared_ptr<ReadableStateField<T>>
         create_readable_field(const std::string& name, const Serializer<T>& sr)
         {
-            if (!SerializableStateField<T>::is_eeprom_saveable()) assert(false);
-            else return std::make_shared<ReadableStateField<T>>(name, sr, eeprom_save_period);
+            assert(SerializableStateField<T>::is_eeprom_saveable());
```
shihaocao, line 24: lol nice change

shihaocao (changes requested):
Looks good here! Let's take out the changes in USB session and then the debug print.

Once that's done, we can merge this in before kill switch for simplicity sake

shihaocao (approved):
Nice

PR #709, opened 2021-04-09
Piksi off for deployment wait

madeleinemc:
# Piksi off for deployment wait

Fixes # 697

### Summary of changes
- Create field gomspace.piksi_off in GomspaceController.cpp
- Set piksi_off = true if wait < deployment_wait in MissionManager
- Statement to turn off gomspace output to Piksi if piksi_off in GomspaceController

### Testing
- Compiles
- Passes tests in fsw_native_leader
- Plot of state field gomspace.piksi_off is correct when plotted in PsimDebug
- Tested on EDU (run, make sure that Piksi light does not turn on at the beginning and then turns on after setting deployment elapsed to 18000)


shihaocao (approved):
Sweet!

PR #712, opened 2021-04-10
Allowed for different gains in near-field vs far-field

govindchari:
# Near-Field GNC

### Summary of changes
- Allowed for different gains in near-field vs far-field
- Added J_ecef statefield

### Testing
-Still need to run HOOTL


kylekrol:
It looks like we have the right idea here for sure. Two things tho:

* We need to open a PR in `psim` first. The `psim` submodule essesntially acts as it's own git repo here so you can `cd` into it, start another branch, add your changes to it, and open a PR. Once merged, checkout the updated master in the submodule and add the submodule change in the fsw repo.
* The base gains are stored in the `gnc/constants.hpp` header for reference.

PR #713, opened 2021-04-13
PTest Restructuring

kylekrol:
# PTest Restructuring

### Summary of changes
- The overcontrolled boot utility is gone and replaced with `self.initial_state`, which will be now defaulted to `startup`. If you want to begin a ptest case in the `manual` state, we need to do `ws` the state manually in the `post_boot()` function.
- Restructured how a simulation plugs into a simulation. Now, the simulation is pulled in as a mixin class and cycling in the ptest terminal actually now steps the simulation forward as well.
- `base.py` is no more; it is now broken apart in the `base` folder which contains `PTestCase`, `SingleSatCase`, `DualSatCase`, `PSimCase`
- PTest cases will now inherit from either `SingleSatCase` or `DualSatCase` to support whichever test condition is required.
- PTest cases will now inherit from `PSimCase` if they want PSim capabilities. Always put PSimCase second in the class inheritance list.
- Moved to constructor flags instead of properties to reduce clutter and actually leverage inheritance.

### Skipping Deployment Wait
Please set the `self.skip_deployment_wait` member variable to `True`. See the `boot_to.py` file for an example. Please use this member variable instead of using the `speed_up` flag. This will allow us to be a bit more flight-like for testing. This flag will work in both HITL and HOOTL.

### Renaming
- `ci.json` is no more. This is now `hootl_speedup.json`.
- There is also now just `hootl.json` instead of `ci_realtime.json` which we should prefer for _all_ mission testing.
- `hitl_singlesat.json` is now `hitl.json`
- `PsimDebug` is now `PSimDebug`; `DualPsim` is now `DualPSim`; Please be careful to notice the capitalization differences!

### PTestXPSim Restructuring
- `run_single_satcase()` and `run_fullmission_case()` replaced with one thing: `run()`
- `pre_bootsetup` and `post_bootsetup` is replaced with `post_boot()` and `pre_boot()`
- Calling `self.cycle()` now also handles all `psim` state transactions automatically

### Removing outdated PtestCases
- base.py moved
- AMC OG; no longer used
- qft_test.py; no longer used
- quake_fault_handler.py; never was working

### Testing

We'll be happy if all the tests still work!


ptest/cases/boot_to.py
```diff
@@ -0,0 +1,53 @@
+from .base import SingleSatCase, PSimCase
```
tanishqaggarwal, line 1: The fact that psim works cleanly enough to allow this kind of drop-in testcase writing is just *chef's kiss*

ptest/cases/boot_to.py
```diff
@@ -0,0 +1,53 @@
...
+from .utils import TestCaseFailure
+import lin
+
+class BootToStartupCase(SingleSatCase, PSimCase):
```
tanishqaggarwal, line 5: Technically this doesn't need to be a psim case right? Is there value to creating a separate non-psim case?

ptest/cases/piksi_checkout_case.py
```diff
@@ -24,6 +25,8 @@ def setup_post_bootsetup(self):
 
         self.most_common_mode = Enums.piksi_modes["no_fix"]
 
+        self.mission_state = "manual"
```
tanishqaggarwal, line 28: Why?

ptest/cases/prop_fault_handler.py
```diff
@@ -1,12 +1,18 @@
...
 # python -m ptest runsim -c ptest/configs/ci.json -t PropFaultHandler
-class PropFaultHandler(SingleSatOnlyCase):
+class PropFaultHandler(SingleSatCase, PSimCase):
     def __init__(self, *args, **kwargs):
         super(PropFaultHandler, self).__init__(*args, **kwargs)
 
+        self.initial_state = "standby"
+        self.skip_deployment_wait = True
+        self.psim_configs += ["truth/standby"]
+        self.psim_config_overrides["truth.leader.attitude.w"] = lin.Vector3([0.01,0.0711,-0.01])
```
tanishqaggarwal, line 14: Why is this override needed?

ptest/cases/prop_state_machine_case.py
```diff
@@ -1,19 +1,15 @@
...
-    def initial_state(self):
-        return "manual"
-
-    @property
-    def fast_boot(self):
-        return True
+class PropStateMachineCase(SingleSatCase):
+    def post_boot(self):
+        self.flight_controller.write_state(
+            "pan.state", Enums.mission_states["manual"])
```
tanishqaggarwal, line 10:
```suggestion
        self.mission_state = 'manual'
```

ptest/cases/psim_debug.py
```diff
@@ -1,58 +1,22 @@
...
+
+class PSimDebug(SingleSatCase, PSimCase):
+    """
+    comments
+    """
+    def __init__(self, *args, **kwargs):
+        super(PSimDebug, self).__init__(*args, **kwargs)
+        self.initial_state = "startup"
+        self.psim_configs += ['truth/deployment']
+        self.psim_config_overrides["truth.leader.attitude.w"] = lin.Vector3([0.01,0.0738,-0.01])
```
tanishqaggarwal, line 16: Why is this needed?

ptest/cases/quake_power_cycling.py
```diff
@@ -1,19 +1,18 @@
...
-        return "follower"
+class QuakePowerCycling(SingleSatCase):
+    def __init__(self, *args, **kwargs):
+        super(QuakePowerCycling, self).__init__(*args, **kwargs)
 
-    @property
-    def fast_boot(self):
-        return True
+#        self.psim_configs += ["truth/standby"]
+#        self.initial_state = "standby"
```
tanishqaggarwal, line 12:
```suggestion
```

ptest/cases/safehold_reboot.py
```diff
@@ -1,16 +1,20 @@
...
-class SafeholdReboot(SingleSatOnlyCase):
-    @property
-    def initial_state(self):
-        return "standby"
+class SafeholdReboot(SingleSatCase, PSimCase):
+    def __init__(self, *args, **kwargs):
+        super(SafeholdReboot, self).__init__(*args, **kwargs)
+        self.initial_state = "standby"
+        self.psim_configs += ['truth/standby']
+        self.psim_config_overrides["truth.leader.attitude.w"] = lin.Vector3([0.01,0.0711,-0.01])
```
tanishqaggarwal, line 10: Why this override

ptest/cases/safehold_reboot.py
```diff
@@ -1,16 +1,20 @@
-from .base import SingleSatOnlyCase
+from .base import SingleSatCase, PSimCase
 from .utils import FSWEnum, Enums, TestCaseFailure
+import lin
```
tanishqaggarwal, line 3: Can remove this import if the override is not needed

ptest/cases/simple_fault_handlers.py
```diff
@@ -1,16 +1,15 @@
...
-class ADCSWheelFaultHandler(SingleSatOnlyCase):
-    @property
-    def initial_state(self):
-        return "standby"
+class ADCSWheelFaultHandler(SingleSatCase, PSimCase):
+    def __init__(self, *args, **kwargs):
+        super(ADCSWheelFaultHandler, self).__init__(*args, **kwargs)
+        self.initial_state = "standby"
+        self.psim_configs += ['truth/standby']
+        self.psim_config_overrides["truth.leader.attitude.w"] = lin.Vector3([0.01,0.0711,-0.01])
```
tanishqaggarwal, line 10: Override required?

ptest/cases/simple_fault_handlers.py
```diff
@@ -48,23 +47,17 @@ def run_case_singlesat(self):
...
-        return "standby"
-
-    @property
-    def fast_boot(self):
-        return True
+class LowBattFaultHandler(SingleSatCase, PSimCase):
+    def __init__(self, *args, **kwargs):
+        super(LowBattFaultHandler, self).__init__(*args, **kwargs)
+        self.psim_configs += ["truth/standby"]
+        self.psim_config_overrides["truth.leader.attitude.w"] = lin.Vector3([0.01,0.0711,-0.01])
```
tanishqaggarwal, line 54: Override required?

ptest/cases/reset_for_flight.py
```diff
@@ -1,36 +1,38 @@
...
-    #make sure spacecraft is shut down right after this test case so mission doesnt start executing
-    self.ws( "cycle.auto", True )
-    self.cycle()
-    self.print_rs("cycle.auto")
+        self.print_header("Starting Reset")
+        self.ws( "cycle.auto", False )
+        self.set_confirm("gomspace.pptmode_cmd", 1)
+        self.set_confirm("pan.bootcount", 0)
+        self.set_confirm("pan.deployment.elapsed", 0)
+        self.set_confirm("pan.deployed", False)
```
tanishqaggarwal, line 29: These flags are redundant...are we still using both of them in FSW? The condition deployed is equivalent to deployment.elapsed >= 30 mins

ptest/cases/reset_for_flight.py
```diff
@@ -1,36 +1,38 @@
...
 
-    self.print_header("Flight Reset Complete")
-    self.finish()
\ No newline at end of file
+        #ensure autocycling is happening 
+        #make sure spacecraft is shut down right after this test case so mission doesnt start executing
```
tanishqaggarwal, line 32: Is shutting down the spacecraft up to the testcase operator? How might we be able to enforce this?

ptest/cases/reset_for_flight.py
```diff
@@ -1,36 +1,38 @@
...
-  def run_case_singlesat(self):
-    self.print_header("Starting Reset")
-    self.ws( "cycle.auto", False )
-    self.set_confirm("gomspace.pptmode_cmd", 1)
-    self.set_confirm("pan.bootcount", 0)
-    self.set_confirm("pan.deployment.elapsed", 0)
-    self.set_confirm("pan.deployed", False)
+    def run(self):
+        self.mission_state = "manual"
+        self.cycle()
```
tanishqaggarwal, line 22: Why call self.cycle() when cycle.auto may be true?

ptest/cases/boot_to.py
kylekrol: No but what we're trying to go for here is booting with the simulation and sensor data more often. These essentially test for regressions in full mission cases.

Also, I'm not sure about the value of having the satellite in something like standby without sensor data -- we replaced all testcases that required it with a `psim` testcase.

ptest/cases/piksi_checkout_case.py
kylekrol: Lot's of people tried to fast boot to manual and this is the equivalent. It worries me tho that github says this is in the `setup_post_bootsetup` function when it should just be `post_boot` @shihaocao?

ptest/cases/prop_fault_handler.py
kylekrol: #720.

ptest/cases/simple_fault_handlers.py
kylekrol: #720.

ptest/cases/simple_fault_handlers.py
kylekrol: #720.

ptest/cases/reset_for_flight.py
kylekrol: Yeah we could change this so that it manually cycles a certain number of times. Separate ticket though I'd say.

kylekrol: Oh wait but we can't because autocycle would be true...

kylekrol: Seems like we may need to open a ticket for this case in general @shihaocao?

tanishqaggarwal: I feel like this is OK just as long as the operator actually shuts down the spacecraft. Spending a few cycles in the startup deployment phase is no biggie really

kylekrol: Ah true true as long as we run it in realtime.

ptest/cases/reset_for_flight.py
kylekrol: Also a good point haha Can take a look at this a little later.

ptest/cases/gomspace_checkout_case.py
```diff
@@ -14,7 +14,7 @@ def str_to_bool(self, string):
         else:
             raise ValueError
 
-    def run_case_singlesat(self):
+    def run(self):
         self.failed = False
```
madeleinemc, line 18: set gomspace.piksi_off to false

madeleinemc (changes requested):
I think it looks good! I looked over the description and new base files, as well as focusing on PiksiCheckoutCase and GomspaceCheckoutCase

ptest/cases/gomspace_checkout_case.py
kylekrol: Is this something that should go in this PR or should that be included somewhere else?

madeleinemc: Ohh yeah but will it be overwritten by this change to gomspace_checkout_case?

shihaocao: Im suprised that CI didn't catch this spooky interaction on the last PR

kylekrol: Seems like it's just because we were hard setting the sat into manual.

ptest/cases/reset_for_flight.py
shihaocao: If it is true, then cycling is fine here anyway

madeleinemc (approved):

18kudlugim (approved):
Looks pretty solid, everything in the PR overview matches what's in the base folder, etc., and the PTest cases now all run locally

Duncan-McD (approved):
Looks good to me

PR #714, opened 2021-04-13
Fix Serializer to Allow Us to Toggle Flows

fatimayousuf:
# Fix Serializer to Allow Us to Toggle Flows

### Summary of changes
- We use a field called `downlink.toggle_id` to deactivate/reactivate flows of telemetry. 
- When testing this feature using ptest, I sent an uplink packet that would set `downlink.toggle_id` to 2, so as to deactivate the second flow. However, the value of `downlink.toggle_id` would not change at all
- This serializer change allows us to properly set `downlink.toggle_id` and toggle flows


src/fsw/FCCode/DownlinkProducer.cpp
```diff
@@ -6,9 +6,9 @@ DownlinkProducer::DownlinkProducer(StateFieldRegistry& r,
     const unsigned int offset) : TimedControlTask<void>(r, "downlink_ct", offset),
                                  snapshot_ptr_f("downlink.ptr"),
                                  snapshot_size_bytes_f("downlink.snap_size"),
-                                 shift_flows_id1_f("downlink.shift_id1", Serializer<unsigned char>(0,10,1)),
-                                 shift_flows_id2_f("downlink.shift_id2", Serializer<unsigned char>(0,10,1)),
-                                 toggle_flow_id_f("downlink.toggle_id", Serializer<unsigned char>(0,10,1))
+                                 shift_flows_id1_f("downlink.shift_id1", Serializer<unsigned char>(37)),
```
tanishqaggarwal, line 9: Yikes, good catch. Might want to consider not hardcoding this value (you should be able to get the total number of uplinkable fields by calling `registry.writable_fields.size()`. One issue with this, though, is that the set of writable fields is not fully populated until the entire flight software has been able to construct fields. So in order to make this work, you might have to defer the creation of these fields to the `init()` function.

tanishqaggarwal: Uhh idk why I said uplinked fields bc these fields are obviously bounded by the number of flows, not uplinkable fields. Sorry, disregard that part, but the thing about moving creation of fields to `init()` in order to dynamically initialize based on the number of flows still holds.

shihaocao: Wouldn't this actually have to revolve around the total number of flows, and not the number of writable fields? Unsure where to get a parameter that describes the total number of flows.

If at all possible, statically determing the number of total flows we have would be nice as opposed to a hard coded constant. Otherwise we'll need to update after the next flows update

shihaocao:
It seems like we have two choices, Merge in this PR as is, or write some code to dynamically load this.

Do we have any unit tests that test reordering flows? That might be useful if we decide to just merge this PR and then manually reconfigure that serializer everytime we update flows

src/fsw/FCCode/DownlinkProducer.cpp
tanishqaggarwal: Yes, suggested changes:
1) Convert these fields into state field pointers (i suggest using `std::unique`)
2) Construct these state fields in the `init_flows` function below. Use `flow_data.size()` to populate the serializer bound. e.g.
```
shift_fp = std::make_unique<ReadableStateField<unsigned char>>(flow_data.size());
add_writable_field(shift_fp.get());
```

tanishqaggarwal:
This looks good to me now, except for the unit test failure. Deferring to @shihaocao on approval

shihaocao (approved):

shihaocao:
This is nice!

PR #716, opened 2021-04-14
AMC propagation

afoarce:
# AMC propagation

Fixes #.

### Summary of changes
- Added simulation to AMC for propagating orbits through psim
- Updated GPSTime to easily convert from sim to uplink field types

### Testing
Tested in HOOTL with running autotelem on dualpsim case. Printed uplink orbits every 10 seconds.
Screen Shot 2021-04-13 at 8 20 45 PM


kylekrol:
Very excited! Were we able to see the relative orbit estimate transition to propagating?

afoarce:
> Very excited! Were we able to see the relative orbit estimate transition to propagating?

I haven't figured out why but I do not see the rel_orbit downlink fields being updated. I can read the uplinked fields through usb session and those seem to be updating.

ptest/cases/autonomous_mission_manager_pure_radio.py
```diff
@@ -1,56 +1,110 @@
...
+        # get default sim configs
+        configs = ["sensors/base", "truth/base", "truth/detumble"]
+        configs = ["lib/common/psim/config/parameters/" + f + ".txt" for f in configs]
+        # should we use the default truth.dt.ns, or should it depend on something like prop_time?
+        config = Configuration(configs)
+
+        # update values to current (sim assumes leader, works equally for follower)
+        config["truth.leader.orbit.r"] = lin.Vector3(orbit.pos)
+        config["truth.leader.orbit.v"] = lin.Vector3(orbit.vel)
+        config["truth.t.ns"] = GPSTime(*(orbit.time)).to_pan_ns()  # what are the units on this
```
kylekrol, line 94: Units are nanoseconds if that wasn't clear by now.

kylekrol (approved):

ptest/cases/autonomous_mission_manager_pure_radio.py
```diff
@@ -65,48 +120,51 @@ def sim_model(self):
...
-
-            #wait for data from both spacecrafts to come down from Iridium
-            orbit_data_fields = ["orbit.pos", "orbit.vel"]
-            while("Unable to find" in self.leader.read_state("orbit.time") or 
-                    "Unable to find" in self.follower.read_state("orbit.time")): 
+        self.comms_time_threshold = 60 * 5  # currently 5 minutes for testing
+
+        # Pass telemetry between spacecraft
+        #while(self.continue_mission()):    #for running mission
+        while 1: #for testing purposes
```
shihaocao, line 140: Looks good to me, so long as we have plans to change it back

shihaocao (approved):
Wow looks solid. Excited to see this in mission operation!

ptest/cases/autonomous_mission_manager_pure_radio.py
afoarce: Yes, with the current testing setup the satellites are always in standby which would make the AMC exit, but once we give them leader/follower designation we can change it back.

PR #717, opened 2021-04-14
Close approach gnc

govindchari:
# Close Approach GNC

### Summary of changes
- Increased number of firing in near-field
- Decreased gains in near-field


src/fsw/FCCode/OrbitController.cpp
```diff
@@ -120,32 +139,57 @@ void OrbitController::execute() {
...
+            }
         }
-    }
+        return min_time;
+    };
 
-    // Return the smallest positive time
-    return min_time;
+    return (rel_orbit_state==static_cast<unsigned char>(rel_orbit_state_t::estimating)) ? 
+            next_node(firing_nodes_near) : next_node(firing_nodes_far);
```
tanishqaggarwal, line 164: This one-liner is difficult to parse, I would suggest breaking it out into several lines

src/fsw/FCCode/OrbitController.cpp
```diff
@@ -120,32 +139,57 @@ void OrbitController::execute() {
...
+        data.p = gnc::constant::K_p / gain_factor;  
+        data.d = gnc::constant::K_d / gain_factor;
+        data.energy_gain = gnc::constant::K_e / gain_factor;   // Energy gain                   (J)
+        data.h_gain = gnc::constant::K_h / gain_factor;        // Angular momentum gain         (kg m^2/sec)
+    } else {
+        data.p = gnc::constant::K_p;
+        data.d = gnc::constant::K_d;
+        data.energy_gain = gnc::constant::K_e;   // Energy gain                   (J)
+        data.h_gain = gnc::constant::K_h;        // Angular momentum gain         (kg m^2/sec)
+    }
```
tanishqaggarwal, line 191:
```suggestion
    data.p = gnc::constant::K_p;
    data.d = gnc::constant::K_d;
    data.energy_gain = gnc::constant::K_e;   // Energy gain                   (J)
    data.h_gain = gnc::constant::K_h;        // Angular momentum gain         (kg m^2/sec)
    
    if (rel_orbit_state==static_cast<unsigned char>(rel_orbit_state_t::estimating)) {
        data.p /= gain_factor;  
        data.d /= gain_factor;
        data.energy_gain /= gain_factor;
        data.h_gain /= gain_factor;
    }
```

tanishqaggarwal (commented):
*read in Bill Clinton's reedy Arizonian voice* y'all call this GNC? back in mah day we wrote e'erything in matlab or ain't worth shit. can't deal with all of this C++ mumbo jumbo y'all call guidance these days. what the HECK is going on

src/fsw/FCCode/OrbitController.cpp
```diff
@@ -1,8 +1,20 @@
...
 
+// Firing nodes
+constexpr double pi = gnc::constant::pi;
+static constexpr std::array<double, 3> firing_nodes_far = {pi/3, pi, -pi/3};
+static constexpr std::array<double, 18> firing_nodes_near = {pi/18, pi/6, pi*(5/18), pi*(7/18), pi/2, pi*(11/18), 
+                                pi*(13/18), pi*(5/6), pi*(17/18), -pi*(17/18),
+                                -pi*(5/6), -pi*(13/18), -pi*(11/18), -pi/2,
+                                -pi*(7/18), -pi*(5/18), -pi/6, -pi/18};
+
+auto gain_factor = (double)(firing_nodes_near.size() / firing_nodes_far.size());
```
kylekrol, line 16: We actually need to cast first otherwise it's still going to do integer math and only then cast to a double. No reason not to make this `static constexpr` either.
```suggestion
static constexpr auto gain_factor = static_cast<double>(firing_nodes_near.size()) / firing_nodes_far.size();
```

src/fsw/FCCode/OrbitController.cpp
```diff
@@ -18,16 +30,20 @@ OrbitController::OrbitController(StateFieldRegistry &r, unsigned int offset) :
     sched_valve1_f("orbit.control.valve1", Serializer<unsigned int>(1000)),
     sched_valve2_f("orbit.control.valve2", Serializer<unsigned int>(1000)),
     sched_valve3_f("orbit.control.valve3", Serializer<unsigned int>(1000)),
-    sched_valve4_f("orbit.control.valve4", Serializer<unsigned int>(1000))
+    sched_valve4_f("orbit.control.valve4", Serializer<unsigned int>(1000)),
+    J_ecef_f("orbit.control.J_ecef", Serializer<lin::Vector3d>(0,10,100))
```
kylekrol, line 34: We should be careful with these serializer values here! We don't need `100` bits resolution on the vector's magnitude. For reference the arguments go:

    Serializer<lin::Vector3d>(min_magnitude, max_magnitude, bits_for_magnitude)

src/fsw/FCCode/OrbitController.cpp
```diff
@@ -18,16 +30,20 @@ OrbitController::OrbitController(StateFieldRegistry &r, unsigned int offset) :
...
     add_writable_field(sched_valve1_f);
     add_writable_field(sched_valve2_f);
     add_writable_field(sched_valve3_f);
     add_writable_field(sched_valve4_f);
+    add_writable_field(J_ecef_f);
     sched_valve1_f.set(0);
     sched_valve2_f.set(0);
     sched_valve3_f.set(0);
     sched_valve4_f.set(0);
+    J_ecef_f.set(0.0);
```
kylekrol, line 46: Perhaps you mean this? That being said, `lin` types are guaranteed to default construct full of zeros -- part of them being `constexpr` types.
```suggestion
    J_ecef_f.set(lin::zeros<lin::Vector3d>());
```

src/fsw/FCCode/OrbitController.cpp
kylekrol: Lolll I was actually the one who suggested doing this instead of having it turn into a mutliple return statement function and what not. Making it an `if` `else` would just be the same number of lines pretty much?

src/fsw/FCCode/OrbitController.cpp
kylekrol: This definitely seems to be the way to go.

kylekrol (commented):
Looks like we're definitely heading in the right direction here -- once we get CI figured out should be good to go pretty quickly.

src/fsw/FCCode/OrbitController.cpp
```diff
@@ -120,32 +139,57 @@ void OrbitController::execute() {
...
-        double time_til_node = (firing_nodes[i] - theta) / ang_vel;
-        if (time_til_node > 0 && time_til_node < min_time) {
-            min_time = time_til_node;
+
+    auto next_node = [&](auto const &arr) -> double {
+        for (auto i = 0; i < arr.size(); i++) {
+            double time_til_node = (arr[i] - theta) / ang_vel;
+            if (time_til_node > 0 && time_til_node < min_time) {
+                min_time = time_til_node;
+            }
```
kylekrol, line 158: Keeping `min_time` outside of the lambda but then still returning it is, at the very least, weird and we should move it into the local scope something like this:
```suggestion
    auto next_node = [&](auto const &arr) -> double {
        auto min_time = std::numeric_limits<double>::max();
        for (auto const node : arr) {
            auto const time_til_node = (node - theta) / ang_vel;
            if (time_til_node > 0 && time_til_node < min_time) {
                min_time = time_til_node;
            }
```
Also, we can use a range based for loop if we want! Just don't forget to remove line 127/151 if you go with this change.

src/fsw/FCCode/OrbitController.cpp
```diff
@@ -18,16 +31,20 @@ OrbitController::OrbitController(StateFieldRegistry &r, unsigned int offset) :
     sched_valve1_f("orbit.control.valve1", Serializer<unsigned int>(1000)),
     sched_valve2_f("orbit.control.valve2", Serializer<unsigned int>(1000)),
     sched_valve3_f("orbit.control.valve3", Serializer<unsigned int>(1000)),
-    sched_valve4_f("orbit.control.valve4", Serializer<unsigned int>(1000))
+    sched_valve4_f("orbit.control.valve4", Serializer<unsigned int>(1000)),
+    J_ecef_f("orbit.control.J_ecef", Serializer<lin::Vector3d>(0,0.1,10))
```
kylekrol, line 35: Is there really not a place we can pull this constant from? Having a random magic number floating around really isn't ideal...

kylekrol (commented):
All the flight software changes look good to me assuming @fatimayousuf is good with them as well. Otherwise, I think you just need to update the orbit controller unit tests and maybe look into adding something to test the close approach behavior if that seems feasible.

fatimayousuf (approved):
Orbit controller changes look fine to me! I'd wait on Kyle's approval though since he's a lot more knowledgeable of GNC stuff than I am.

src/fsw/FCCode/OrbitController.cpp
```diff
@@ -120,31 +143,70 @@ void OrbitController::execute() {
...
+        dr_smoothed = dr;
+    } else {
+        dr_smoothed = alpha * dr + (1.0 - alpha) * dr_smoothed;
+    }
+
+    if (!lin::all(lin::isfinite(dv_smoothed))) {
+        dv_smoothed = dv;
+    } else {
+        dv_smoothed = alpha * dv + (1.0 - alpha) * dv_smoothed;
+    }
```
kylekrol, line 189: I worry about the exponential filtering happening here. We want to be sure that the filtering is happening on every time step regardless of what the orbit controller is doing (perhaps it's disabled at a given time step because the attitude estimator is broken or something)?

I'd suggest moving the exponential filter logic to the very start of execute and have a single check like:

    if (relative orbit estimate is valid) {
        do the exponential filtering and/or initialization if things are NaNs
    } else {
        set things to NaNs because we don't want old filtered values hanging around
    }

src/fsw/FCCode/OrbitController.cpp
```diff
@@ -18,16 +31,23 @@ OrbitController::OrbitController(StateFieldRegistry &r, unsigned int offset) :
...
     add_writable_field(sched_valve3_f);
     add_writable_field(sched_valve4_f);
+    add_writable_field(J_ecef_f);
+    add_writable_field(alpha_f);
     sched_valve1_f.set(0);
     sched_valve2_f.set(0);
     sched_valve3_f.set(0);
     sched_valve4_f.set(0);
+    J_ecef_f.set(lin::zeros<lin::Vector3d>());
+    alpha_f.set(0.4);
```
kylekrol, line 50: We want to initialize the filtered values to `lin::nans<lin::Vector3d>()` here too otherwise the filter is gonna think it was already initialized.

src/fsw/FCCode/OrbitController.cpp
```diff
@@ -18,16 +31,23 @@ OrbitController::OrbitController(StateFieldRegistry &r, unsigned int offset) :
     sched_valve1_f("orbit.control.valve1", Serializer<unsigned int>(1000)),
     sched_valve2_f("orbit.control.valve2", Serializer<unsigned int>(1000)),
     sched_valve3_f("orbit.control.valve3", Serializer<unsigned int>(1000)),
-    sched_valve4_f("orbit.control.valve4", Serializer<unsigned int>(1000))
+    sched_valve4_f("orbit.control.valve4", Serializer<unsigned int>(1000)),
+    J_ecef_f("orbit.control.J_ecef", Serializer<lin::Vector3d>(0,0.025,10)),
+    alpha_f("orbit.control.alpha", Serializer<double>(0,1,10))
+
```
kylekrol, line 37:
```suggestion
    alpha_f("orbit.control.alpha", Serializer<double>(0,1,10))
```

kylekrol (changes requested):
I think we need to shift where the exponential filtering is happening to avoid bugs where the relative orbit estimate was valid and then all of a sudden became invalid again. Otherwise, it looks great.

src/fsw/FCCode/OrbitController.cpp
```diff
@@ -120,31 +163,55 @@ void OrbitController::execute() {
...
 
-    // Return the smallest positive time
-    return min_time;
+    return (rel_orbit_state == static_cast<unsigned char>(rel_orbit_state_t::estimating)) ? 
+            next_node(firing_nodes_near) : next_node(firing_nodes_far);
 }
 
 lin::Vector3d OrbitController::calculate_impulse(double t, const lin::Vector3d &r, const lin::Vector3d &v, 
-    const lin::Vector3d &dr, const lin::Vector3d &dv) {
+    const lin::Vector3d &dr_smoothed, const lin::Vector3d &dv_smoothed) {
```
kylekrol, line 191: I think we should change the name of the input parameters just back to `dr` `dv` here as the function doesn't really care if it's handling smoothed data or not.
```suggestion
    const lin::Vector3d &dr, const lin::Vector3d &dv) {
```

src/fsw/FCCode/OrbitController.cpp
```diff
@@ -120,31 +163,55 @@ void OrbitController::execute() {
...
+    unsigned char rel_orbit_state=rel_orbit_valid_fp->get();
 
     // Assemble the input Orbit Controller data struct
     data.t = t;
     data.r_ecef = r;
     data.v_ecef = v;
-    data.dr_ecef = dr;
-    data.dv_ecef = dv;
+    data.dr_ecef = dr_smoothed;
+    data.dv_ecef = dv_smoothed;
```
kylekrol, line 201:
```suggestion
    data.dr_ecef = dr;
    data.dv_ecef = dv;
```

src/fsw/FCCode/OrbitController.cpp
kylekrol, line 50:
```suggestion
    J_ecef_f.set(lin::zeros<lin::Vector3d>());
    alpha_f.set(0.4);
    dr_smoothed = lin::nans<lin::Vector3d>();
    dv_smoothed = lin::nans<lin::Vector3d>();
```

kylekrol (approved):
Seems good to me assuming CI goes through!

PR #726, opened 2021-04-16
Debug Console Improvements

kylekrol:
Will be rebased once #713 is merged.

### Summary of changes

- Remove all use of the `memory` and `set` headers from the debug console.
- Update a macro.
- Remove an explicit `this` capture from a lambda which is now a deprecated C++ feature (don't agree with the deprecation but it is the way it is haha).

### Testing

Existing unit tests with HOOTL and HITL should be sufficient. No regressions were observed.


src/common/debug_console.hpp
```diff
@@ -63,8 +63,8 @@ class debug_console {
      * newline.
      * @param str The string to be printed.
      */
-    void println(severity s, const char *str);
-    void println(const char *str);
+    static void println(severity s, const char *str);
+    static void println(const char *str);
```
kylekrol, line 67: Not really sure why these weren't static originally?

src/common/debug_console.hpp
```diff
@@ -149,7 +149,7 @@ class debug_console {
      */
     void _reader();
 
-    std::shared_ptr<std::thread> reader_thd;
+    std::thread reader_thd;
```
kylekrol, line 152: No reason to malloc something for this guy - making it a normal class member.

src/common/debug_console.cpp
```diff
@@ -84,7 +92,7 @@ void debug_console::init() {
 #ifdef DESKTOP
         std::cin.tie(nullptr);
         running = true;
-        reader_thd = std::make_shared<std::thread>([this] { this->_reader(); });
+        reader_thd = std::thread([&](){ this->_reader(); });
```
kylekrol, line 95: Capturing `this` by value is deprecated and just using `&` has no additional cost and everything is taken care of at compile time.

src/common/ConstexprMap.hpp
```diff
@@ -0,0 +1,36 @@
...
+    constexpr ConstexprMap &operator=(ConstexprMap const &) = delete;
+    constexpr ConstexprMap &operator=(ConstexprMap &&) = delete;
+
+    constexpr auto operator[](K const &k) const {
+      auto const it = std::find_if(data.cbegin(), data.cend(), [&](auto const &kv) {
+          return (k == kv.first);
+        });
+
+      if (it == data.cend()) throw std::range_error("Key not found");
+      assert(it != data.cend());
```
kylekrol, line 31:
```suggestion
```

src/common/debug_console.cpp
```diff
@@ -9,26 +12,31 @@
...
-    {state_field_error::invalid_mode, "mode value is not 'r' or 'w'"},
-    {state_field_error::missing_field_val, "missing value of field to be written"},
-    {state_field_error::invalid_field_val, "field value was invalid"}};
-
-std::map<debug_console::state_cmd_mode, const char*> debug_console::state_cmd_mode_strs{
-    {state_cmd_mode::unspecified_mode, "perform unspecified operation with"},
-    {state_cmd_mode::read_mode, "read"},
-    {state_cmd_mode::write_mode, "write"},
-};
+static constexpr ConstexprMap<debug_console::severity, char const *, 8> severity_strs {{{
```
tanishqaggarwal, line 15: Wow this is so hot

tanishqaggarwal (approved):
Seems like everything passes so we're good?

src/common/debug_console.cpp
kylekrol: > Seems like everything passes so we're good?

If it breaks we'll fix it haha But HOOTL is happy and most of the changes were related to HOOTL.

PR #728, opened 2021-04-16
Sorted State Field Registry

kylekrol:
### Summary of changes

- All `std::vector<T *>`'s are now sorted in the state field registry.
- Move `add` and `find` implementations to mainly templated functions for simplicity.
- Use `DYNAMIC_CAST` macro in the state field registry mock -- found some broken casts in the GOMSpace unit test because of this.

### Testing

Existing unit tests and booting HOOTLs should be sufficient unless we're worried about downlink/uplink stuff and the fact that field indices can change when adding a field.


src/fsw/FCCode/UplinkCommon.h
```diff
@@ -26,7 +26,7 @@ class Uplink {
   /**
    * @brief Reference to registry so that we can update and read fields
    */
-   StateFieldRegistry& registry;
+   StateFieldRegistry const& registry;
```
kylekrol, line 29:
```suggestion
   StateFieldRegistry const &registry;
```

src/fsw/FCCode/UplinkCommon.h
```diff
@@ -26,7 +26,7 @@ class Uplink {
   /**
    * @brief Reference to registry so that we can update and read fields
    */
-   StateFieldRegistry& registry;
+   StateFieldRegistry const &registry;
```
kylekrol, line 29: No reason for this to not be `const`.

test/test_fsw_uplink_consumer/test_fsw_uplink_consumer.cpp
```diff
@@ -63,17 +63,15 @@ class TestFixture {
         uplink_consumer = std::make_unique<UplinkConsumer>(registry, 0);
 
         radio_mt_packet_fp->set(mt_buffer);
-        field_map = std::map<std::string, size_t>();
```
kylekrol, line 66: This is already default initialized as a class member.

test/test_fsw_uplink_consumer/test_fsw_uplink_consumer.cpp
```diff
@@ -63,17 +63,15 @@ class TestFixture {
...
+        uplink_consumer->init_uplink();
 
         // Setup field_map and assign some values
         for (size_t i = 0; i < registry.writable_fields.size(); ++i)
         {
             auto w = registry.writable_fields[i];
-            field_map[w->name().c_str()] = i;
-            w->name();
```
kylekrol, line 73: `c_str()` causes an extra `std::string` construction and the second line doesn't do anything.

src/fsw/FCCode/UplinkCommon.h
kylekrol: Scratch that apparently there is a reason lmao

test/test_fsw_gomspace/test_fsw_gomspace.cpp
```diff
@@ -154,12 +154,12 @@ class TestFixture {
...
-        power_cycle_output3_cmd_fp = registry.find_writable_field_t<unsigned char>("gomspace.power_cycle_output3_cmd");
-        power_cycle_output4_cmd_fp = registry.find_writable_field_t<unsigned char>("gomspace.power_cycle_output4_cmd");
-        power_cycle_output5_cmd_fp = registry.find_writable_field_t<unsigned char>("gomspace.power_cycle_output5_cmd");
-        power_cycle_output6_cmd_fp = registry.find_writable_field_t<unsigned char>("gomspace.power_cycle_output6_cmd");
+        power_cycle_output1_cmd_fp = registry.find_writable_field_t<bool>("gomspace.power_cycle_output1_cmd");
+        power_cycle_output2_cmd_fp = registry.find_writable_field_t<bool>("gomspace.power_cycle_output2_cmd");
+        power_cycle_output3_cmd_fp = registry.find_writable_field_t<bool>("gomspace.power_cycle_output3_cmd");
+        power_cycle_output4_cmd_fp = registry.find_writable_field_t<bool>("gomspace.power_cycle_output4_cmd");
+        power_cycle_output5_cmd_fp = registry.find_writable_field_t<bool>("gomspace.power_cycle_output5_cmd");
+        power_cycle_output6_cmd_fp = registry.find_writable_field_t<bool>("gomspace.power_cycle_output6_cmd");
```
kylekrol, line 162: This issue was caught because of the move to dynamic casting in the mocked state field registry.

test/test_fsw_uplink_consumer/test_fsw_uplink_consumer.cpp
```diff
@@ -129,7 +127,7 @@ void test_create_uplink()
     size_t packet_size = tf.uplink_consumer->index_size + tf.uplink_consumer->get_field_length(idx);
 
     std::vector<bool> expect(packet_size, 0);
-    expect[0] = 0;
+    expect[0] = 1;
```
shihaocao, line 130: why did these change

kylekrol: Because the ordering of the state fields within the state field registry is different after being sorted. This shouldn't break things that aren't hard coded like this but I would like @tanishqaggarwal to weigh in.

kylekrol:
Alright, so there is far far too much code that depends on the state field registry not moving things once added it seems like. This shouldn't be an issue because we don't add fields after initialization but, nevertheless, EEPROM controller is still failing and I don't really know why. I'm going to remove the sorted state field registry changes and just bring in the dynamics casting updates.

PR #730, opened 2021-04-16
Attitude Estimator Validity Fault

kylekrol:
### Summary of changes

If the attitude estimator isn't valid while we're trying to control attitude (pretty much in standby, leader, follower, or a close approach phase) we need to return to standby. Otherwise, the wheels could spin up to unreasonable speeds and we may just waste power.

### Testing

See the added `ptest` case.


shihaocao (approved):
nice

PR #734, opened 2021-04-17
Control Task Timing Fixes

### LINKED ISSUE #624: Verification of control cycle times
2 main issues:
-We need to verify control cycle times to make sure that they are not consistently late or displaying "cascading" lateness (where one task is late and causes all subsequent tasks to also be late)
-We need to make sure that the num_lates field in the timed control task is accurately counting the number of late tasks.

Some issues were observed when reading multiple gomspace state fields over a period of two minutes which causes an excessive number of tasks to be late in HITL. In HOOTL, the quantity of late tasks is lower but still consistently increasing which is a problem. Shown Below:

HITL:
Screen Shot 2021-02-05 at 4 50 39 PM

HOOTL:
Screen Shot 2021-02-05 at 4 51 11 PM

These times were found after running the gomspace logger ptest case which just reads a collection of gomspace fields for two minutes. However, when running the Magnetorquer test case in HOOTL similar behavior was found, the difference is the late tasks weren't increasing, which is due to the nature of the test. 
Screen Shot 2021-02-05 at 5 06 33 PM

knk38:
# Control Task Timing Fixes

Fixes #624 

### Summary of changes
- Added a new field that represents the duration of a control task. (timing." control task name ".duration)
- Changed offset representation to update the time after the end of each task rather than using a raw offset from the start of the control cycle.
- Shifted allotted time for each control task 
- Added a new field that represents the duration of a control cycle (pan.cc_duration)

### List of Timing Values Spreadsheet
https://docs.google.com/spreadsheets/d/1jCz0EPqmQYzXp4ZEQNZETDVkG-WAafRNskDdm0EUKvc/edit#gid=0

### Testing
Below are graphs for each control task representing its duration over time. 
adcs_commander_time
adcs_monitor_time
attitude_controller_time
dcdc_time
debug_time
docking_time
downlink_time
eeprom_ct_time
gomspace_time
mission_time
orbit_time
piksi_time
prop_time
image
uplink_time
Screen Shot 2021-04-19 at 10 17 55 AM
Screen Shot 2021-04-19 at 10 33 03 AM

### Constants
The values of constants for allotted time for each control task were adjusted to represent 1.4 * control task duration


src/fsw/FCCode/MainControlLoop.cpp
```diff
@@ -100,33 +100,49 @@ void MainControlLoop::execute() {
...
+    TRACKED_CONSTANT_SC(unsigned int, dcdc_duration, 1000);
+    TRACKED_CONSTANT_SC(unsigned int, attitude_controller_duration, 1000);
+    TRACKED_CONSTANT_SC(unsigned int, adcs_commander_duration, 1000);
+    TRACKED_CONSTANT_SC(unsigned int, adcs_box_controller_duration, 10000);
+    TRACKED_CONSTANT_SC(unsigned int, orbit_duration, 5000);
+    TRACKED_CONSTANT_SC(unsigned int, prop_duration, 1000);
+    TRACKED_CONSTANT_SC(unsigned int, downlink_duration, 1000);
+    TRACKED_CONSTANT_SC(unsigned int, quake_duration, 30000);
+    TRACKED_CONSTANT_SC(unsigned int, docking_duration, 10000);
+    TRACKED_CONSTANT_SC(unsigned int, eeprom_duration, 10000);
```
tanishqaggarwal, line 119: This is a total control cycle length of 185.4 ms. Some things around the software stack depend explicitly on a control cycle length provided by `pan.cc_ms`. So, two things:
1) Is there any opportunity to cut the cycle length down? 
2) Can we dynamically set `pan.cc_ms` based on the values of these durations? ConstexprMap might help you in storing all of the durations in some kind of container that can be iterated over in order to sum the durations.

tanishqaggarwal (commented):
I think you forgot to upload the graphs to the PR description

src/fsw/FCCode/MissionManager.cpp
```diff
@@ -16,30 +16,30 @@ const constexpr unsigned char MissionManager::kill_switch_value;
...
+        adcs_functional_fault_fp(FIND_FAULT(adcs_monitor.functional_fault.base)),
+        wheel1_adc_fault_fp(FIND_FAULT(adcs_monitor.wheel1_fault.base)),
+        wheel2_adc_fault_fp(FIND_FAULT(adcs_monitor.wheel2_fault.base)),
+        wheel3_adc_fault_fp(FIND_FAULT(adcs_monitor.wheel3_fault.base)),
+        wheel_pot_fault_fp(FIND_FAULT(adcs_monitor.wheel_pot_fault.base)),
+        pressurize_fail_fp(FIND_FAULT(prop.pressurize_fail.base)),
+        mission_state_f("pan.state", Serializer<unsigned char>(12), 1),
+        is_deployed_f("pan.deployed", Serializer<bool>(), 1000),
+        deployment_wait_elapsed_f("pan.deployment.elapsed", Serializer<unsigned int>(15000), 500),
+        sat_designation_f("pan.sat_designation", Serializer<unsigned char>(2), 1),
```
shihaocao, line 40: probably undo this change to make it EEPROM saved haha

src/fsw/FCCode/MainControlLoop.cpp
```diff
@@ -26,30 +26,30 @@ MainControlLoop::MainControlLoop(StateFieldRegistry& registry,
...
       PIKSI_INITIALIZATION,
-      piksi_control_task(registry, piksi_control_task_offset, piksi),
+      piksi_control_task(registry, piksi),
       ADCS_INITIALIZATION,
-      adcs_monitor(registry, adcs_monitor_offset, adcs),
-      debug_task(registry, debug_task_offset),
-      estimators(registry, estimators_offset),
+      adcs_monitor(registry, adcs),
+      debug_task(registry),
+       estimators(registry),
```
tanishqaggarwal, line 33:
```suggestion
      estimators(registry),
```

src/fsw/FCCode/TimedControlTask.hpp
```diff
@@ -167,22 +176,22 @@ class TimedControlTask : public ControlTask<T>, public TimedControlTaskBase {
...
         ControlTask<T>(registry),
-        offset(us_to_duration(_offset + 1)),
         num_lates_field_name("timing." + name + ".num_lates"),
         num_lates_f(num_lates_field_name, Serializer<unsigned int>()),
         avg_wait_field_name("timing." + name + ".avg_wait"),
-
-        avg_wait_f(avg_wait_field_name, Serializer<float>(0,PAN::control_cycle_time_us, 18))
+        avg_wait_f(avg_wait_field_name, Serializer<float>(0,PAN::control_cycle_time_us, 18)),
+        ct_duration_field_name("timing." + name + ".duration"),
+        ct_duration_f(ct_duration_field_name, Serializer<unsigned int>() )
```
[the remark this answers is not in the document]
knk38: I just copied how the other two fields were made but this makes sense.

src/fsw/FCCode/MissionManager.cpp
knk38: > I think you forgot to upload the graphs to the PR description

Yup, I will add them all today

knk38:
> I think you forgot to upload the graphs to the PR description

Yes, I will upload them today. I wanted to run them on the EDU rather than the teensy I have at home just in case.

src/fsw/FCCode/MainControlLoop.cpp
knk38: With respect to this, I think it is possible fo update pan.cc_ms however I feel that it would be simpler just to adhere to the 170 ms limit. Especially because I can easily remove 15.4 ms of time among all of the tasks so that none of them are late as well.

knk38: However, now that I see that debug ct is taking 150 ms on average there may have to be a change to `pan.cc_ms`

shihaocao:
@knk38 I think our path forward is we need to then also make FSW configuration changes to truly, and completely ignore the execution of debug task (ie allocate no time to it). We really want to make FSW run with the 170 ms mark as we've done tons of testing with 170ms.

If native/hootl testing runs beyond 170 because of debug task, then that's fine. Timing matters less on HOOTL (since no hardware) It is absolutely critical on HITL though.

shihaocao:
We should have a 2.1x factor of safety on Estimators @knk38

src/fsw/FCCode/TimedControlTask.hpp
```diff
@@ -134,11 +136,18 @@ class TimedControlTask : public ControlTask<T>, public TimedControlTaskBase {
      * @param control_cycle_start_time System time for the start of the control task.
      * @return T Value returned by execute().
      */
-    T execute_on_time() {
-      sys_time_t earliest_start_time = 
-        TimedControlTaskBase::control_cycle_start_time + offset;
-      wait_until_time(earliest_start_time);
-      return this->execute();
+    T execute_on_time(unsigned int duration_us) {
```
shihaocao, line 139: we should make this void

shihaocao: a void return

knk38 (commented):
should be final

src/fsw/FCCode/MainControlLoop.cpp
```diff
@@ -100,35 +120,54 @@ void MainControlLoop::execute() {
...
+    TRACKED_CONSTANT_SC(unsigned int, uplink_duration, 10000);
+    TRACKED_CONSTANT_SC(unsigned int, attitude_estimator_duration, 5000);
+    TRACKED_CONSTANT_SC(unsigned int, mission_duration, 1000);
+    TRACKED_CONSTANT_SC(unsigned int, dcdc_duration, 1000);
+    TRACKED_CONSTANT_SC(unsigned int, attitude_controller_duration, 1000);
+    TRACKED_CONSTANT_SC(unsigned int, adcs_commander_duration, 1000);
+    TRACKED_CONSTANT_SC(unsigned int, adcs_box_controller_duration, 10000);
+    TRACKED_CONSTANT_SC(unsigned int, orbit_duration, 5000);
+    TRACKED_CONSTANT_SC(unsigned int, prop_duration, 1000);
+    TRACKED_CONSTANT_SC(unsigned int, downlink_duration, 1000);
```
tanishqaggarwal, line 136: Didn't downlink take a long time when it was being tested?

knk38: Yup, fixed.

knk38: Also any idea why the FSW HOOTL tests would be frozen in CI but working fine locally?

tanishqaggarwal: Idk this happened a lot weirdly when I worked on the code, I never figured it out

Sometimes I was able to replicate the issue using Ubuntu Multipass (neat little virtualization software that makes a small Linux container on your machine). When I fixed the issue on Multipass it usually went away, so maybe try that

src/fsw/FCCode/DockingController.hpp
```diff
@@ -40,9 +39,12 @@ class DockingController : public TimedControlTask<void> {
     //shared pointer set by mission manager - tells control task to dock or undock motor
     const WritableStateField<bool>* docking_config_cmd_fp;
 
-    //fields for step angle and step delay to change the speed/torque of the motor and how long it will turn
+    //field for step angle which should be constant based on its setup
```
afoarce, line 42: The step angle should be constant but we don't know the theoretical value for the motor so that's why this is a writeable state field. Changing it would affect turning based on the checkout case.

src/fsw/FCCode/DownlinkProducer.hpp
```diff
@@ -30,13 +30,12 @@ class DownlinkProducer : public TimedControlTask<void> {
      * @brief Construct a new Downlink Producer.
      * 
      * @param registry State field registry.
-     * @param offset Offset, in microseconds, from the beginning of the control
+      Offset, in microseconds, from the beginning of the control
```
afoarce, line 33: do we need this comment?

src/fsw/FCCode/MainControlLoop.hpp
```diff
@@ -83,38 +83,18 @@ class MainControlLoop : public ControlTask<void> {
...
     ReadableStateField<unsigned int> control_cycle_ms_f;
 
+    ReadableStateField<unsigned int> control_cycle_duration_f;
+    sys_time_t prev_sys_time;
+
     PropController prop_controller;
     MissionManager mission_manager;
     AttitudeController attitude_controller; // needs adcs.state from MissionManager
-    ADCSCommander adcs_commander; // will need inputs from computer
+    ADCSCommander adcs_commander; // will need inputs from computer++
```
afoarce, line 92: what is computer++

src/fsw/FCCode/UplinkConsumer.h
```diff
@@ -46,9 +46,9 @@ class UplinkConsumer : public TimedControlTask<void>, public Uplink {
     
     /**
      * @param registry State field registry
-     * @param offset Offset from the start of the cycle (TimedControlTask)
+      Offset from the start of the cycle (TimedControlTask)
```
afoarce, line 49: what does this comment refer to

afoarce (approved):
Overall changes look good, left some comments/questions.

src/fsw/FCCode/ClockManager.cpp
```diff
@@ -14,14 +14,7 @@ ClockManager::ClockManager(StateFieldRegistry &registry,
...
 void ClockManager::execute() {
-    if (has_executed) {
-        sys_time_t earliest_start_time =
-            TimedControlTaskBase::control_cycle_start_time + control_cycle_size;
-        wait_until_time(earliest_start_time);
-    }
-
-    has_executed = true;
-    TimedControlTaskBase::control_cycle_start_time = get_system_time();
+    TimedControlTaskBase::control_task_end_time = get_system_time() + us_to_duration(1100);
```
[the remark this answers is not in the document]
knk38: Ah so that is the duration of clockmanager that we have allotted, however I think this is going to change soon as clock manager needs to be restructured so it is more adaptive and we can actually enforce a control cycle time

src/fsw/FCCode/DockingController.hpp
knk38: Yeah makes sense, I just split up the comment so that one is for the step angle and one is for the step delay, also whoops this was not supposed to be in this PR but oh well.

knk38:
just as a comment I had this to squash so it merged before I could address the extra comments, but @afoarce your points are valid and I will remove the offset parameter headers in the next PR

PR #735, opened 2021-04-19
Remove Angular Rate Overrides from PTest Cases

### LINKED ISSUE #720: Remove Angular Rate PSim Config Override
There were some PSim config overrides in #713 that introduce the following line

    self.psim_config_overrides["truth.leader.attitude.w"] = lin.Vector3([0.01,0.0711,-0.01])

This won't be needed once the updated `psim` initial conditions get merged in where the satellite's initial angular rate will be below the standby threshold.

kylekrol:
Closes #720.

We actually have CI to test these things now so everything should be alright!

Future PRs should no longer need to override PSim initial conditions to make it into standby quickly.


shihaocao (approved):
Nice

PR #738, opened 2021-04-19
Update Configs

shihaocao:
# Update Configs

Tragically, this PR is necessary as moving forward with testing, we will be potentially all using the email server at once, so we cannot have identical IMEI numbers, and thus must git ignore all files running `autotelem`.

### Summary of changes
- Update git ignore
- Update discord files
- Load in amc.json as a config with two radios so that you can run AMC, you will need to fill in the IMEI numbers properly per your own `hootl_hootl_autotelem.json` or `hootl_hitl_autotelem.json`!


Duncan-McD (approved):
makes sense to me

Issue #691, opened 2021-03-30
Telemetry and Commanding Infrastructure Verification


shihaocao:
Need to come up with a procedure to connect to satellite and how to send commands (especially for people without much experience in telemetry)
Come up with mission procedures: what should we do if some fault happens or something goes wrong?
Send commands and check that appropriate states on the satellite change/proper response occurs
i.e if we see a fault occur, and we send a command to suppress the fault, then we should get downlinked telemetry showing that fault is suppressed
Critical command fields: resetting fields (i.e gomspace), fault suppressing commands, etc.
Check that we can get all the telemetry in the flows
Check that we can reorder flows of telemetry

shihaocao:
#598 parallel track

kylekrol:
Related to #597 and #599.

Issue #690, opened 2021-03-30
Nominal Mission Procedures


shihaocao:
Startup to Standby Transition
Verify that spacecraft is detumbling
Specify what fields we need to see and how they should change
Once spacecraft is in Standby
Gomspace data, Piksi data, Propulsion data, any faults, all GNC estimators ,the docking button isn’t pressed
ADCS HAVT Table
Standby to Leader/Follower
Leader
Check follower location from ground
Follower
Location
Use GNC estimators to locate/propagate Leader
Once in Leader/Follower we should periodically check general satellite “health” : battery voltage, attitude, faults, etc.
Close Approach
Verify Piksi mode and that relative orbit estimator is in “estimating” mode
Downlink orbit estimator fields to ensure satellites are actually “close”
Docking to Docked
Verify they are right next to each other
Check the button was pressed (state field)

Issue #689, opened 2021-03-30
Full Mission Dress Rehearsal


shihaocao:
Start a dual hootl ptest environment with simulation to serve as the base of the Dress Rehearsal
Connect to ground software on server and connect to the flight units (maybe HOOTL)
Displaying state changes or data on the ground
MCT plots (from startup to docking) 
Mission Milestone tests
1. start up → detumbling → standby → leader/follower
2. Exhaustive leader/follower
3. End of leader/follower→ close approach →  docking
Show the effects of each of these in MCT
Todo: 
Psim initial conditions
Set up server with GSW
Skipping deployment hold

PR #740, opened 2021-04-19
Near Field Case

kylekrol:
Relates to #689 #690 #691

### Summary of Changes

* Added a testcase that nominally boots both spacecraft into leader and follow close approach with CDGPS data.
* Updated the far field to near field transition distance from `100` meters to `2000` meters.

### Testing

Sufficient testing was added to CI.


shihaocao (approved):
Sweet

PR #742, opened 2021-04-24
Fix USB Session Synchronization

### LINKED ISSUE #557: USB Session Exceptions
After running some long duration HITLs on the flight units, I've hit the following stack trace somewhat frequently after cases have been running for about fifteen minutes or so:

```
Traceback (most recent call last):
  File "/home/kkrol/git/pan/FlightSoftware/ptest/usb_session.py", line 111, in check_console_msgs
    if self.console.inWaiting() > 0:
  File "/home/kkrol/git/pan/FlightSoftware/venv/lib/python3.8/site-packages/serial/serialutil.py", line 590, in inWaiting
    return self.in_waiting
  File "/home/kkrol/git/pan/FlightSoftware/venv/lib/python3.8/site-packages/serial/serialposix.py", line 467, in in_waiting
    s = fcntl.ioctl(self.fd, TIOCINQ, TIOCM_zero_str)
OSError: [Errno 5] Input/output error
Unspecified error. Exiting.
 - Terminating console connection to and saving logging/telemetry data for FlightController.
Exception in thread FlightController logger thread:
Traceback (most recent call last):
  File "/home/kkrol/git/pan/FlightSoftware/ptest/usb_session.py", line 111, in check_console_msgs
    if self.console.inWaiting() > 0:
  File "/home/kkrol/git/pan/FlightSoftware/venv/lib/python3.8/site-packages/serial/serialutil.py", line 590, in inWaiting
    return self.in_waiting
  File "/home/kkrol/git/pan/FlightSoftware/venv/lib/python3.8/site-packages/serial/serialposix.py", line 467, in in_waiting
    s = fcntl.ioctl(self.fd, TIOCINQ, TIOCM_zero_str)
OSError: [Errno 5] Input/output error
During handling of the above exception, another exception occurred:
Traceback (most recent call last):
  File "/usr/lib/python3.8/threading.py", line 932, in _bootstrap_inner
    self.run()
  File "/usr/lib/python3.8/threading.py", line 870, in run
    self._target(*self._args, **self._kwargs)
  File "/home/kkrol/git/pan/FlightSoftware/ptest/usb_session.py", line 169, in check_console_msgs
    self.disconnect()
  File "/home/kkrol/git/pan/FlightSoftware/ptest/usb_session.py", line 447, in disconnect
    self.check_msgs_thread.join()
  File "/usr/lib/python3.8/threading.py", line 1008, in join
    raise RuntimeError("cannot join current thread")
RuntimeError: cannot join current thread
```

This has happened while running the `DitlCase` which is very similiar to the `SpinMotorsCase` except it continues to log telemetry over the course of the simulation.

Unfortunately, the ptest logs weren't saved except for the raw log which is attached here:
[FlightController_raw-log.txt](https://github.com/pathfinder-for-autonomous-navigation/FlightSoftware/files/5587196/FlightController_raw-log.txt)

### LINKED ISSUE #743: Bazel Failing to Install of Github Actions' Servers
Per the title. Not really sure but this may have been related to the recent update to Ubuntu 20.xx?

kylekrol:
Fixes #557 
Fixes #743

### Summary of changes

- Update synchronization primitives around USB Session read and write state.
- USB session's HTTP endpoint was also removed to try and reduce the general overhead of `ptest`.

There were two main issues with the previous implementation. First, the Python `queue.Queue` doesn't guarantee a FIFO behavior in terms of the threads that are blocking on the queue (spurious wakeups), just FIFO behavior with the items in the queue:

https://github.com/python/cpython/blob/196983563d05e32d2dcf217e955a919f9e0c25e1/Lib/queue.py#L154-L183

Secondly, we have no synchronization in terms of when a physical request was actually sent over and who blocked on the requests queue:

https://github.com/pathfinder-for-autonomous-navigation/FlightSoftware/blob/a0078703eb03bf9aef6e5ffaf8f27ff56ec188b2/ptest/usb_session.py#L204-L214

https://github.com/pathfinder-for-autonomous-navigation/FlightSoftware/blob/a0078703eb03bf9aef6e5ffaf8f27ff56ec188b2/ptest/usb_session.py#L226-L231

https://github.com/pathfinder-for-autonomous-navigation/FlightSoftware/blob/a0078703eb03bf9aef6e5ffaf8f27ff56ec188b2/ptest/usb_session.py#L272-L279

To fix this, the new design works as follows:

1. A read state or write state function will grab the requests lock that ensure they're the only thread adding requests to the queue.
2. The requests are actually added to the queue.
3. The thread with the requests queue will then grad the device lock to send the request over the serial line.
4. Both the requests and device lock are released.
5. The thread reading or writing state will block on each requests `data` property which will happen until the thread processing incoming has fulfilled the request.

### Testing

Threading thing are inherently difficult to test but HOOTL worked very well on much system spamming read state commands alongside a `psim` simulation. Hopefully that means the issues are cleared up!

### DIFF
--- a/.github/workflows/hootl.yml
+++ b/.github/workflows/hootl.yml
@@ -31,21 +31,18 @@ jobs:
     - uses: actions/setup-go@v2
       with:
         go-version: '^1.13.1'
+    - uses: bazelbuild/setup-bazelisk@v1
+    - name: Mount bazel cache
+      uses: actions/cache@v2
+      with:
+        path: "~/.cache/bazel"
+        key: bazel
 
     - name: Install dependencies
       run: |
         python -m pip install --upgrade pip
         pip install -r requirements.txt
         platformio update
-
-    - name: Install Bazel
-      run: |
-        sudo apt install apt-transport-https curl gnupg
-        curl -fsSL https://bazel.build/bazel-release.pub.gpg | gpg --dearmor > bazel.gpg
-        sudo mv bazel.gpg /etc/apt/trusted.gpg.d/
-        echo "deb [arch=amd64] https://storage.googleapis.com/bazel-apt stable jdk1.8" | sudo tee /etc/apt/sources.list.d/bazel.list
-        sudo apt upgrade
-        sudo apt install bazel
 
     - name: Install Psim
       run: |
--- a/.github/workflows/hootl.yml
+++ b/.github/workflows/hootl.yml
@@ -40,10 +40,11 @@ jobs:
 
     - name: Install Bazel
       run: |
-        sudo apt install curl gnupg
+        sudo apt install apt-transport-https curl gnupg
         curl -fsSL https://bazel.build/bazel-release.pub.gpg | gpg --dearmor > bazel.gpg
         sudo mv bazel.gpg /etc/apt/trusted.gpg.d/
         echo "deb [arch=amd64] https://storage.googleapis.com/bazel-apt stable jdk1.8" | sudo tee /etc/apt/sources.list.d/bazel.list
+        sudo apt upgrade
         sudo apt install bazel
 
     - name: Install Psim
--- a/ptest/usb_session.py
+++ b/ptest/usb_session.py
@@ -7,14 +7,12 @@ import queue
 import os
 import pty
 import subprocess
-from multiprocessing import Process
 import glob
 from elasticsearch import Elasticsearch
 import email
 import imaplib
 
 from .data_consumers import Datastore, Logger
-from .http_cmd import create_usb_session_endpoint
 from . import get_pio_asset
 from .cases.utils import str_to_val
 import lin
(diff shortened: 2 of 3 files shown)


.github/workflows/hootl.yml
```diff
@@ -31,21 +31,19 @@ jobs:
     - uses: actions/setup-go@v2
       with:
         go-version: '^1.13.1'
+    - uses: bazelbuild/setup-bazelisk@v1
+    - name: Mount bazel cache
+      uses: actions/cache@v2
+      with:
+        path: "~/.cache/bazel"
+        key: bazel
```
kylekrol, line 39: This is apparently the more Github Actions way of doing things?
https://github.com/marketplace/actions/setup-bazelisk

ptest/usb_session.py
```diff
@@ -30,6 +28,38 @@ class USBSession(object):
...
+            with self.__lock:
+                assert self.__data is None, "Error; the request has already been filled"
+                assert data is not None, "Error; a request must be filled with a value other than None"
+
+                self.__data = data
+                self.__has_reply.notify_all()
+
+        @property
+        def field(self) -> str:
+            return self.__field
```
kylekrol, line 62: This is the new synchronization mechanism that allows the requesting thread to block on a queued condition variables (i.e. FIFO ordered is now enforced for the threads themselves) until the USB session parsing thread populates data.

ptest/usb_session.py
```diff
@@ -221,15 +230,21 @@ def read_state(self, field, timeout = None):
...
-        return self._wait_for_state(field)
+        request = USBSession.Request(field)
+        with self.request_lock:
+            self.requests.put(request, block=False)
+            with self.device_lock:
+                self.console.write(json_cmd.encode())
 
+        self.raw_logger.put("Sent:     " + json_cmd.rstrip())
+
+        return request.data['val']
```
kylekrol, line 247: The order of the operations here is incredibly important! Event down to adding the request to the queue prior to writing it over the console tty!

ptest/usb_session.py
```diff
@@ -269,15 +284,16 @@ def _write_state_basic(self, fields, vals, timeout = None):
...
+        with self.request_lock:
+            for request in requests:
+                self.requests.put(request, block=False)
+            with self.device_lock:
+                self.console.write(json_cmds.encode())
 
-        returned_vals = []
-        for field in fields:
-            returned_vals.append(self._wait_for_state(field, timeout))
+        self.raw_logger.put("Sent:     " + json_cmds)
```
kylekrol, line 294: Order of operations is super important here as well!

tanishqaggarwal, line 294: Should we remove this print?

tanishqaggarwal (approved):
Nice!

ptest/usb_session.py
kylekrol: This is in the raw logger? Do we not want that recorded in the raw logging file?

shihaocao (approved):
solid

ptest/usb_session.py
tanishqaggarwal: Oh ur right it's part of the raw logger, oops

PR #744, opened 2021-04-26
Fancy Testcase Flight Controllers

kylekrol:
### Summary of changes

This is really just a quality of life thing. Essentially, this PR attempts to centralize the way in which we interact with a flight controller across dual and single sat cases. Much like `psim`, there is now a `self.leader: FancyFlightController` in all test cases and a `self.follower: FancyFlightController` in dual sat cases.

The PR is pretty straightforward, but now we have the follower abilities:

```
self.[leader|follower].rs(<field>, print=[True|False])
self.[leader|follower].ws(<field>, <value>, print=[True|False])
```

along with the `mission_state` property (we could add more of these and even come up with a fancy decorate to generate them for us!)

Most changes are intended to be backwards compatible; however, some write state functions were removed that did a `ws` immediately followed by a read to reduce test case overhead.

### DIFF
--- a/ptest/cases/base/dual_sat_case.py
+++ b/ptest/cases/base/dual_sat_case.py
@@ -1,13 +1,7 @@
 from ...data_consumers import Logger
-import time
-import math
-import threading
-import traceback
 from ..utils import Enums, TestCaseFailure, suppress_faults
-import psim # the actual python psim repo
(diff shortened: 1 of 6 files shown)


shihaocao (approved):
Sweet

PR #747, opened 2021-04-27
Add Sleep on Desktop Platforms

kylekrol:
HOOTLs no longer pin a core to 100% for each fsw binary. Can someone verify this still builds on Mac?

I also suspect this could help with performance issues on Linux. Most likely, the scheduler was demoting the fsw binary as a long running program without many interrupts/system calls.


shihaocao (approved):

tanishqaggarwal:
I strongly suggest reversion until we verify that the timing is still okay.

Issue #695, opened 2021-04-01
Day-In-The-Life Test


kylekrol:
We want to run flight software for an extended period of time (on the order of a day or more) to verify flight software is actually stable enough - i.e. no subtle memory leaks and execution time remains constant over time.

This ideally would be run on an actual Teensy but we could think about running a long-duration HOOTL case as well if we monitor memory usage.

DITL Logging

24 Hr HOOTL

24 HR HITL QuakeBox

3 HR HITL EDU 

Relates to #604 and #624.

We may have issues with #557.

shihaocao:
Want to track memory usage over Day in the Life, as well as cycles passed (indirect verification of timing)

angelaloh09:
Goals of Ditl case were to fsw can run for long period of time, check for regular control cycle times & to ensure no memory leaks. 
- Verification of control cycle times with EDU 70 min. test attached in #624
- Verification of no memory leak plots/log screenshots for HOOTL Logging, 3HR Quake, and 70 min. Edusat attached in #604

Issue #604, opened 2021-01-03
Ensure no memory leaks of FlightSoftware


tanishqaggarwal:
We should run FSW for a long, long time and check if memory usage keeps increasing.

samuelqian01:
![Screenshot from 2021-04-17 10-30-30](https://user-images.githubusercontent.com/65980474/115117257-a6948380-9f6b-11eb-9ed5-a416ee85c001.png)
![Screenshot from 2021-04-17 10-35-44](https://user-images.githubusercontent.com/65980474/115117260-a8f6dd80-9f6b-11eb-8f75-af0308344e49.png)

No memory leaks after many many many hours of running.

angelaloh09:
3 Hr Quake Test shows no memory leaks:
<img width="523" alt="Screen Shot 2021-05-08 at 10 18 58 AM" src="https://user-images.githubusercontent.com/44074422/117542486-d40ea300-afe6-11eb-970b-3ce8c6eb4783.png">

angelaloh09:
70 min EduSat shows no memory leaks and normal cycling: (detumble state)

No Memory Leaks:
<img width="639" alt="pan_memory_use_ditl" src="https://user-images.githubusercontent.com/44074422/117542541-0f10d680-afe7-11eb-8c35-06f9957f4b2f.png">


Cycling: 
![pan_cycle_no_detumble_ditl](https://user-images.githubusercontent.com/44074422/117542519-f43e6200-afe6-11eb-896c-0498197992c5.png)

PR #764, opened 2021-05-08
Ditl

angelaloh09:
# Ditl Case

Fixes #. 
- #695 
- #604

### Summary of changes
- Added Day in the Life Test Case which inherits from SingleSatDetumbleCase

### Testing
- Tested: DITL Logging, 24 Hr HOOTL, 3 Hr Quake Test, 70 min. Edu Sat Test
- All tests logged data and showed no memory leaks and stable cycling

Logs plot outputs for cycling uploaded to issue #624 and plots for no mem leaks have been attached to issue #695


configs/flight_and_test_envs.ini
```diff
@@ -44,6 +44,10 @@ build_flags = ${fsw_teensy_hitl.build_flags} ${follower.build_flags}
 extends = fsw_teensy36
 build_flags = ${fsw_teensy_hitl.build_flags} ${leader.build_flags}
 
+[env:fsw_teensy36_hitl_leader_autotelem]
+extends = env:fsw_teensy36_hitl_leader
+build_flags = ${env:fsw_teensy36_hitl_leader.build_flags} -D AUTOTELEM
+
```
shihaocao, line 50: To confirm these two configs were used for the 24 HITL with no quake?

ptest/cases/ditl.py
```diff
@@ -0,0 +1,30 @@
+# Runs mission from startup state to standby state.
+from psim.sims import SingleAttitudeOrbitGnc
+from .utils import Enums, mag_of, sum_of_differentials
+from .mission import SingleSatDetumbleCase
```
kylekrol, line 4:
```suggestion
from .mission import SingleSatDetumbleCase
```

kylekrol: Don't think we need these extra imports either.

kylekrol (approved):
I'd just recommend removing the extra code in the suggestions I have but looks good!

configs/flight_and_test_envs.ini
angelaloh09: Yeah, those were used for 24 HITL, but I don't think they were committed because they were included in the gitignore file. @samuelqian01 do you wanna push the configs to the branch? I don't have them on my end

knk38 (approved):
looks good to me, pending kyle's suggestions

PR #767, opened 2021-05-11
Reduce Debug Console Cpu Overhead with Condition Variables

kylekrol:
### Summary of Changes

* Removed the lockless queue in favor of a deque and condition variables from the stl.
* This also prevents each flight software binary from pinning a Cpu core in Hootl which will hopefully help with performance issues with our big integrated testing campaign.

### Testing

Timing performance with updates:
mutex

and without updates:
mutexless

I don't see any noticeable regressions?


src/fsw/FCCode/DebugTask.cpp
```diff
@@ -20,12 +20,20 @@ void DebugTask::execute() {
 #ifndef FLIGHT
   start_cycle_f.set(false);
   if(auto_cycle_f.get()){
-    process_commands(_registry);
+    process_commands(_registry
+#ifdef DESKTOP
+    , false
+#endif
+    );
```
tanishqaggarwal, line 27:
```suggestion
    #ifdef DESKTOP
       const bool blocking = true;
    #else
       const bool blocking = false;
    #endif
    process_commands(_registry, blocking);
```

Just a little cleaner, IMO

tanishqaggarwal (approved):
Gud shit

src/fsw/FCCode/DebugTask.cpp
kylekrol: Yeah no good move for sure!

kylekrol: Oh wait actually can't do this because the argument doesn't exist in HOOTL.

tanishqaggarwal: Oh why not just let the argument exist and be useless in HITL...oh the compiler will complain?

tanishqaggarwal: Whatever it doesn't matter

kylekrol: Compiler errors for unused argument yeah.

PR #778, opened 2021-05-12
don't fire unless in follower/follower close approach

fatimayousuf:
# don't fire unless in follower/follower close approach

### Summary of changes
- Add another condition to `validate_schedule()` in `PropController.cpp` to check if we are in follower or close approach before firing the prop system


shihaocao (approved):
Looks good to me!

tanishqaggarwal (commented):
I think I'm missing context on this PR, but this doesn't seem like a good idea--what if you want to test the prop system in standby?

EDIT: as a followup, I see that manual is an allowable mode for the prop controller--was the intent of this change to allow testability? But we should never run manual mode in flight because it turns off all fault handlers, so we would still have issues with being able to test prop in flight.

fatimayousuf:
> I think I'm missing context on this PR, but this doesn't seem like a good idea--what if you want to test the prop system in standby?
> 
> EDIT: as a followup, I see that manual is an allowable mode for the prop controller--was the intent of this change to allow testability? But we should never run manual mode in flight because it turns off all fault handlers, so we would still have issues with being able to test prop in flight.

I believe we only have a week and a half to finish flight software, but @shihaocao can correct me on that. So I don't know if we would really have time to do much hardware testing with the prop system anymore. In flight, I believe those are these only two modes (follower/close approach) we should fire in anyways. We shouldn't be firing the prop system in standby, and it's not hard to go from standby to follower with USB session. We added manual just in case we wanted to run some sorta ptest case idk

kylekrol:
I agree with @fatimayousuf here that we just need to be done and I don't think test firing the propulsion system is a reasonable ask. For starters, it'd be very difficult to decipher whether or not the appropriate delta V was applied and that basically means doing the test firing is useless imo. We can just watch temperatures and pressure in follower mode and check to see the orbits are converging over time.

tanishqaggarwal:
I'm just saying, why change the state machine to restrict functionality, especially if there's only a week and a half left? This just introduces a test gap (a small one to be sure, but why even add one?) What's the motivation behind this change?

fatimayousuf:
> I'm just saying, why change the state machine to restrict functionality, especially if there's only a week and a half left? This just introduces a test gap (a small one to be sure, but why even add one?) What's the motivation behind this change?

Because without this change, during the actual mission, there's nothing stopping the prop system from firing in startup, standby or safe hold.

knk38:
Based on Tanishq's comments I would have to agree that we should leave the prop state machine alone, I think a better place for these changes is to add a few lines of code in the orbit controller which is already doing the scheduling anyway. Echoing @kylekrol 's comment, prop should remain a sole hardware controller while orbit should be doing the higher level thinking

fatimayousuf:
> Based on Tanishq's comments I would have to agree that we should leave the prop state machine alone, I think a better place for these changes is to add a few lines of code in the orbit controller which is already doing the scheduling anyway. Echoing @kylekrol 's comment, prop should remain a sole hardware controller while orbit should be doing the higher level thinking

Ok yeah, I can move this condition to the orbit controller!

test/test_fsw_orbit_controller/test_orbit_controller.cpp
```diff
@@ -24,6 +24,7 @@ class TestFixture {
```
tanishqaggarwal, line 24: Should we also add a negative unit test, e.g. we don't cause a firing if we're not in follower/close approach?

govindchari (approved):

PR #783, opened 2021-05-21
Changed near field regression slope

govindchari:
# Near-Field Regression Constant
Tank2 pressure was lowered to 15psi from 25psi in near field. As a result, we need to change the regression slope that allows us to go from impulse to valve opening times. With a lower tank pressure, we will have a lower thrust. Thrust and pressure vary linearly ideally, so the slope for 15psi is (15/25) multiplied by the slope for 25psi.

### Constants
Changes valve_time_lin_reg_slope_near


shihaocao (approved):
ok

PR #785, opened 2021-05-25
updated reset case

angelaloh09:
# updated reset_for_flight case

### Summary of changes
- ptest writes eeprom saved fields to initial conditions each cycle for 1010 cycles.
- prints init and final states 

### Testing
- ran test on teensy
Screen Shot 2021-05-24 at 11 13 59 PM


ptest/cases/reset_for_flight.py
```diff
@@ -3,36 +3,48 @@
...
+        #additional 10 cycles to make sure fields are saved to EEPROM
+        cycle_duration = cycle_no + 1000 + 10 
+
+        while (cycle_no < cycle_duration):
+            self.print_header("cycle_no: " + str(self.rs("pan.cycle_no")))
+            
+            self.ws("gomspace.pptmode_cmd", 1)
+            self.ws("pan.bootcount", 0)
+            self.ws("pan.deployment.elapsed", 0)
+            self.ws("pan.deployed", False)
```
shihaocao, line 34: kill switch should also be in here, and we can remove gomspace.pptmode_cmd

shihaocao: "eeprom_saved_fields": {
        "cycle.auto": 1,
        "cycle.start": 1,
        "pan.bootcount": 1000,
        "pan.deployed": 1000,
        "pan.deployment.elapsed": 500,
        "pan.kill_switch": 100,
        "pan.state": 1
    },

shihaocao: Eeprom saved fields are above, per the `telemetry` file

knk38: PPT mode for gomspace isnt EEPROM tracked but I feel like it is still important to ensure that it is set to MPPT?

shihaocao: image

shihaocao: image

shihaocao: I believe it has already defaulted to 1, so that should be fine

shihaocao:
I think case looks good now, my question would just be, what should we be looking for outputs to know that the case worked properly? And or has the `update` branch, and/or has `pprocs` been updated?

knk38:
> I think case looks good now, my question would just be, what should we be looking for outputs to know that the case worked properly? And or has the `update` branch, and/or has `pprocs` been updated?

I think we can update pprocs to add  verification of the final values of each field. Since this case prints them out, operator can just visually check that they are what we expect. We would have to add a small chart or something in pprocs that indicates what should be expected

angelaloh09:
> > I think case looks good now, my question would just be, what should we be looking for outputs to know that the case worked properly? And or has the `update` branch, and/or has `pprocs` been updated?
> 
> I think we can update pprocs to add verification of the final values of each field. Since this case prints them out, operator can just visually check that they are what we expect. We would have to add a small chart or something in pprocs that indicates what should be expected

Just updated pprocs with testing procedure + expected conditions, and made a pr for that

shihaocao (approved):
sick

shihaocao:
For each EEPROM saved field
- What is the correct value to save as. Yes correct
- Do we know it been saved at the end of the procedure. Yes
- We have to turn it on one more time after this procedure, does that have adverse effects. No* if done properly

PR #786, opened 2021-05-25
Quake fault handler HITL

18kudlugim:
QFH PTest changes following HITL Testing:
- Needed to change the number of control cycles in a day to allow power cycling in HITL to complete in the smallest amount of time possible. Power cycling takes approximately 45 seconds in HITL to complete, and this must happen before the next power cycle is to take place. 
- For the above reasons, we created a QFH_SPEEDUP build flag that sets the tracked constant one_day_ccno to 1400.
- quake_power_cycling.py PTest formatting cleanup
- need to make a note to change build flag from SPEEDUP to QFH_SPEEDUP when conducting tests


ptest/cases/quake_power_cycling.py
```diff
@@ -26,51 +27,56 @@ def qfh_state(self):
...
+    def advance_to_next_qfh_state(self, time):
         for i in range(time):
             self.cycle()
             self.diagnostics()
             self.cycles_since_blackout_start += 1
 
             # Check mission state when in QFH forced standby
-            if self.qfh_state() == Enums.qfh_states['forced_standby'] and not self.mission_state == "standby":
+            if self.qfh_state() == Enums.qfh_states[
+                    'forced_standby'] and not self.mission_state == "standby":
```
knk38, line 48: maybe move this back to be one line for readability sake

knk38 (approved):
Looks great!

src/fsw/FCCode/constants.hpp
```diff
@@ -23,6 +23,10 @@ namespace PAN {
         #endif
     #endif
 
+    # ifdef QFH_SPEEDUP
```
fatimayousuf, line 26: Can you make a new teensy36 pio environment that has the QFH SPEEDUP build flag? There's already one for the SPEEDUP flag, so you can mostly just copy that.

Then, in the mission procedures, we can tell them to upload this environment to the satellite before running this checkout case

fatimayousuf (changes requested):

src/fsw/FCCode/constants.hpp
18kudlugim: sounds good, I pushed the changes just now

ptest/cases/quake_power_cycling.py
```diff
@@ -81,14 +86,16 @@ def diagnostics(self):
     def run(self):
         # The satellite has been in a blackout since startup. Cycle count starts at 1.
         self.cycles_since_blackout_start = self.rs("pan.cycle_no") - 1
-
-        # These steps necessary in HOOTL, unsure about how it will affect HITL
         self.ws("radio.state", Enums.radio_states["wait"])
-        self.ws("gomspace.power_cycle_output3_cmd",True)
+
+        # This step is necessary in HOOTL, must be commented out in HITL
```
fatimayousuf, line 91: Why is this step not necessary in HITL?

fatimayousuf (approved):

ptest/cases/quake_power_cycling.py
18kudlugim: I think it has something to do with an external subsystem setting a value in HOOTL. I discussed this with @kylekrol earlier in the sem and it was determined to be HOOTL-only. Wasn't an issue when we ran it in HITL, though. If you run the qfh test in HOOTL and plot it you would probably see the issue with the first power cycle not happening, and only subsequent ones following expected behavior

knk38:
Since this was hitl tested a few times it is good to go

PR #791, opened 2021-07-08
update constants control cycle time 120 to 170

shihaocao:
This PR updates the control cycle time for FLIGHT from 120 ms to 170 ms.
A critical bug, and fix.


afoarce (approved):
seems legit

PR #796, opened 2021-08-09
Gnc upgrade

shihaocao:
On this ticket:
- when we shouldn't fire do not zero out any valves, this lets us maintain manual control of the valves, even when the orbit controller doesn't autonomously request a valve firing
- move the condition of only firing when in follower or follower close approach as an early return statement, if we are not in those two states. This prevents us from scheduling a firing or setting a valve schedule when we are not in those states
- Ensure all transitions out of Pressurizing have a reset call to ensure all valves are closed
- Allow entry into the pressurizing state if the firing is closer than the minimum required number of cycles to "ensure" we hit target pressure
- Allow the pressurizing state to exit to idle if the firing is too close (possibly meaning we don't hit target pressure).
- Add test case to test the above

### DIFF
--- a/src/fsw/FCCode/OrbitController.cpp
+++ b/src/fsw/FCCode/OrbitController.cpp
@@ -88,10 +88,12 @@ void OrbitController::execute() {
     // If we don't have all the information we need, don't calculate a firing
     if (!time_valid_fp->get() || !orbit_valid_fp->get() || !rel_orbit_valid_fp->get() ||
             !attitude_estimator_valid_fp->get()) {
-        sched_valve1_f.set(0);
-        sched_valve2_f.set(0);
-        sched_valve3_f.set(0);
-        sched_valve4_f.set(0);
+        return;
+    }
+    
+    // Don't fire unless we're in follower and follower close approach.
+    mission_state_t mission_state = static_cast<mission_state_t>(mission_state_fp->get());
+    if (!(mission_state == mission_state_t::follower || mission_state == mission_state_t::follower_close_approach)){
         return;
     }
 
@@ -137,42 +139,24 @@ void OrbitController::execute() {
         prop_cycles_until_firing_fp->set(time_till_firing_cc);
     }
 
-    // Check if the satellite is around a firing point and the prop system is ready to fire
-    // and if the time and orbit data is valid
-    if ( time_till_firing_cc < 20 && static_cast<prop_state_t>(prop_state_fp->get()) == prop_state_t::await_firing) {
+    // Collect the output of the PD controller and get the needed impulse
+    lin::Vector3d J_ecef = calculate_impulse(t, r, v, dr_smoothed, dv_smoothed);
 
-        // Collect the output of the PD controller and get the needed impulse
-        lin::Vector3d J_ecef = calculate_impulse(t, r, v, dr_smoothed, dv_smoothed);
+    // Save J_ecef to statefield
+    J_ecef_f.set(J_ecef);
 
-        // Save J_ecef to statefield
-        J_ecef_f.set(J_ecef);
+    // Transform the impulse from ecef frame to the eci frame
+    lin::Vector4d q_eci_ecef;
+    gnc::utl::quat_conj(q_ecef_eci, q_eci_ecef);
+    lin::Vector3d J_eci;
+    gnc::utl::rotate_frame(q_eci_ecef, J_ecef, J_eci);
 
-        // Transform the impulse from ecef frame to the eci frame
-        lin::Vector4d q_eci_ecef;
-        gnc::utl::quat_conj(q_ecef_eci, q_eci_ecef);
-        lin::Vector3d J_eci;
-        gnc::utl::rotate_frame(q_eci_ecef, J_ecef, J_eci);
+    // Transform the impulse from eci frame to the body frame of the spacecraft
+    lin::Vector4f q_body_eci = q_body_eci_fp->get();
+    lin::Vector3d J_body;
+    gnc::utl::rotate_frame(lin::cast<double>(q_body_eci).eval(), J_eci, J_body);
 
-        // Transform the impulse from eci frame to the body frame of the spacecraft
-        lin::Vector4f q_body_eci = q_body_eci_fp->get();
-        lin::Vector3d J_body;
-        gnc::utl::rotate_frame(lin::cast<double>(q_body_eci).eval(), J_eci, J_body);
-
-        // Communicate desired impulse to the prop controller.
-        mission_state_t mission_state = static_cast<mission_state_t>(mission_state_fp->get());
-        if (mission_state == mission_state_t::follower || mission_state == mission_state_t::follower_close_approach) {
-            schedule_valves(J_body);
-        }
-        else{
-            // 0 out schedules
-            schedule_valves(lin::zeros<lin::Vector3d>());
-        }
-    }
-    else{
-        // 0 out schedules
-        schedule_valves(lin::zeros<lin::Vector3d>());
-    }
-
+    schedule_valves(J_body);
 }
 
 double OrbitController::time_till_node(double theta, const lin::Vector3d &pos, const lin::Vector3d &vel) {
--- a/src/fsw/FCCode/OrbitController.cpp
+++ b/src/fsw/FCCode/OrbitController.cpp
@@ -133,7 +133,7 @@ void OrbitController::execute() {
     double time_till_firing_cc = time_till_firing * 1000 / PAN::control_cycle_time_ms;
 
     // Schedule the valves for firing soon if the prop system is idle
-    if (time_till_firing_cc <= (prop_min_cycles_needed() + 10) && static_cast<prop_state_t>(prop_state_fp->get()) == prop_state_t::idle) {
+    if (time_till_firing_cc <= (prop_min_cycles_needed() + 10) && static_cast<prop_state_t>(prop_state_fp->get()) == prop_state_t::idle) { // should we keep the <=
         prop_cycles_until_firing_fp->set(time_till_firing_cc);
     }
 
@@ -163,7 +163,10 @@ void OrbitController::execute() {
         if (mission_state == mission_state_t::follower || mission_state == mission_state_t::follower_close_approach) {
             schedule_valves(J_body);
         }
-
+    }
+    else{
+        // 0 out schedules
+        schedule_valves(lin::zeros<lin::Vector3d>());
     }
 
 }
--- a/src/fsw/FCCode/OrbitController.cpp
+++ b/src/fsw/FCCode/OrbitController.cpp
@@ -163,6 +163,10 @@ void OrbitController::execute() {
         if (mission_state == mission_state_t::follower || mission_state == mission_state_t::follower_close_approach) {
             schedule_valves(J_body);
         }
+        else{
+            // 0 out schedules
+            schedule_valves(lin::zeros<lin::Vector3d>());
+        }
     }
     else{
         // 0 out schedules
--- a/src/fsw/FCCode/PropController.cpp
+++ b/src/fsw/FCCode/PropController.cpp
@@ -390,7 +390,7 @@ bool PropState_Pressurizing::can_enter() const
 
     // min_cycles_needed is actually an upper bound on how long we could need
     bool is_time_to_pressurize = (
-            (controller->cycles_until_firing.get() <= controller->min_cycles_needed())
+            (controller->cycles_until_firing.get() <= controller->min_cycles_needed() - 1)
         );
 
     return ((was_await_pressurizing || was_idle) && is_time_to_pressurize && is_schedule_valid && is_functional);
--- a/src/fsw/FCCode/PropController.cpp
+++ b/src/fsw/FCCode/PropController.cpp
@@ -281,12 +281,9 @@ bool PropState_AwaitPressurizing::can_enter() const
 {
     bool was_idle = controller->check_current_state(prop_state_t::idle);
     bool is_schedule_valid = controller->validate_schedule();
-    // enter await pressurizing if orbit controller requests a firing.
-    bool requested =
-        controller->cycles_until_firing.get() > 0;
     bool is_functional = PropulsionSystem.is_functional();
 
-    return (was_idle && is_schedule_valid && is_functional && requested);
+    return (was_idle && is_schedule_valid && is_functional);
 }
 
 void PropState_AwaitPressurizing::enter()
@@ -394,7 +391,6 @@ bool PropState_Pressurizing::can_enter() const
     // min_cycles_needed is actually an upper bound on how long we could need
     bool is_time_to_pressurize = (
             (controller->cycles_until_firing.get() <= controller->min_cycles_needed())
-            && (controller->cycles_until_firing.get() > 0)
         );
 
     return ((was_await_pressurizing || was_idle) && is_time_to_pressurize && is_schedule_valid && is_functional);
--- a/src/fsw/FCCode/PropController.cpp
+++ b/src/fsw/FCCode/PropController.cpp
@@ -282,11 +282,11 @@ bool PropState_AwaitPressurizing::can_enter() const
     bool was_idle = controller->check_current_state(prop_state_t::idle);
     bool is_schedule_valid = controller->validate_schedule();
     // Enter Await Pressurizing rather than Pressurizing if we have MORE than enough time
-    bool more_than_enough_time =
-        controller->cycles_until_firing.get() >= controller->min_cycles_needed();
+    // bool more_than_enough_time =
+    //     controller->cycles_until_firing.get() >= controller->min_cycles_needed();
     bool is_functional = PropulsionSystem.is_functional();
 
-    return (was_idle && is_schedule_valid && more_than_enough_time && is_functional);
+    return (was_idle && is_schedule_valid && is_functional);
 }
 
 void PropState_AwaitPressurizing::enter()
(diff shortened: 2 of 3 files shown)


src/fsw/FCCode/PropController.cpp
```diff
@@ -391,15 +391,14 @@ bool PropState_Pressurizing::can_enter() const
     bool was_idle = controller->check_current_state(prop_state_t::idle);
     bool is_functional = PropulsionSystem.is_functional();
 
-    // It is time to pressurize when we have min_cycles_needed - 1 cycles left
-    bool is_time_to_pressurize =
-        controller->cycles_until_firing.get() == controller->min_cycles_needed() - 1;
-    
-    return ((was_await_pressurizing || was_idle) && is_time_to_pressurize && is_schedule_valid && is_functional);
+    return ((was_await_pressurizing || was_idle) && is_schedule_valid && is_functional);
```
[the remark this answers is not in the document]
athena255: I think this change should be fine though as long as you are not concerned about potentially spending a long time in AwaitFiring with a pressurized outer tank. The things we were concerned about was leaking propellant and overheating. In general, if you are ok with pressurizing immediately, then this change should be ok.

src/fsw/FCCode/PropController.cpp
```diff
@@ -391,15 +391,14 @@ bool PropState_Pressurizing::can_enter() const
...
 }
 
 prop_state_t PropState_Pressurizing::evaluate()
 {
+    // if our desired time to fire has already passed, early exit to idle to try again
+    if (controller->cycles_until_firing.get() == 0)
+        return prop_state_t::idle;
     if (controller->can_enter_state(prop_state_t::handling_fault))
         return prop_state_t::handling_fault;
     return ActionCycleOpenClose::evaluate();
```
athena255, line 404: ActionCycleOpenClose::evaluate() does the actual opening and closing of the valves.  The thing to be careful of is not accidentally leaving a valve open...I don't recall if that is actually a possibility. That will consume a lot of power very fast.

src/fsw/FCCode/OrbitController.cpp
```diff
@@ -10,10 +10,38 @@ const constexpr double OrbitController::valve_time_lin_reg_intercept;
...
+    -1.884955592,-1.850049007,-1.815142422,-1.780235837,-1.745329252,-1.710422667,
+    -1.675516082,-1.640609497,-1.605702912,-1.570796327,-1.535889742,-1.500983157,
+    -1.466076572,-1.431169987,-1.396263402,-1.361356817,-1.326450232,-1.291543646,
+    -1.256637061,-1.221730476,-1.186823891,-1.151917306,-1.117010721,-1.082104136,
+    -1.047197551,-1.012290966,-0.977384381,-0.942477796,-0.907571211,-0.872664626,
+    -0.837758041,-0.802851456,-0.767944871,-0.733038286,-0.698131701,-0.663225116,
+    -0.628318531,-0.593411946,-0.558505361,-0.523598776,-0.488692191,-0.453785606,
+    -0.41887902,-0.383972435,-0.34906585,-0.314159265,-0.27925268,-0.244346095,
+    -0.20943951,-0.174532925,-0.13962634,-0.104719755,-0.06981317,-0.034906585
+    };
```
kylekrol, line 44: How did we end up generating this? Perhaps if you used code leave a little snippet as a comment in case we changed it later.

src/fsw/FCCode/OrbitController.cpp
```diff
@@ -135,7 +163,7 @@ void OrbitController::execute() {
     double time_till_firing_cc = time_till_firing * 1000 / PAN::control_cycle_time_ms;
 
     // Schedule the valves for firing soon if the prop system is idle
-    if (time_till_firing_cc <= (prop_min_cycles_needed() + 10) && static_cast<prop_state_t>(prop_state_fp->get()) == prop_state_t::idle) {
+    if (time_till_firing_cc <= (prop_min_cycles_needed() + 10) && static_cast<prop_state_t>(prop_state_fp->get()) == prop_state_t::idle) { // should we keep the <=
         prop_cycles_until_firing_fp->set(time_till_firing_cc);
     }
```
kylekrol, line 168: Why do we even care about how long the prop system is gonna need -- I thought we were just gonna have the orbit controller schedule stuff and it was up to prop ultimately whether or not it gets executed eventually?

src/fsw/FCCode/OrbitController.cpp
```diff
@@ -165,7 +193,10 @@ void OrbitController::execute() {
         if (mission_state == mission_state_t::follower || mission_state == mission_state_t::follower_close_approach) {
             schedule_valves(J_body);
         }
-
+    }
+    else{
+        // 0 out schedules
+        schedule_valves(lin::zeros<lin::Vector3d>());
     }
```
kylekrol, line 200: Why do we need this if the existing schedule would be in the past?

athena255 (commented):
In general, the main thing to be careful of is to not forget to close the valves. 
Since controlcycles are like 170 ms, and we always open the valve for 1s during a single pressurizing cycle, then that means the valves will be open for several controlcycles. If you exit Pressurizing early during the  1000/170 control cycles in which the valve is open, it is possible that the state machine will not close the valves. 

I do, however, recall making the valves force close upon entering idle...or something. I believe there are unit tests for PropFaultHandler for testing the case where we exit Firing early, but I'll have to look into it a bit later.

athena255:
Also, if you need the state machine to allow firings at a lower pressurization threshold, you can set the statefield for that. It is a 25psi default, but this can be set from the ground.

kylekrol:
This is just a loose first pass -- I'll be back!

kylekrol:
> Also, if you need the state machine to allow firings at a lower pressurization threshold, you can set the statefield for that. It is a 25psi default, but this can be set from the ground.

I think the main reason we didn't want to do this is because the tank wouldn't attempt to pressurize further? Or is that not the case? Perhaps no changes/minimal changes are actually required...

src/fsw/FCCode/PropController.cpp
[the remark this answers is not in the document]
kylekrol: Wait... Yeah @shihaocao this doesn't seem like a condition we want to get rid of. We don't need to rush pressurization for no reason.

src/fsw/FCCode/PropController.cpp
```diff
@@ -391,15 +391,14 @@ bool PropState_Pressurizing::can_enter() const
...
-    bool is_time_to_pressurize =
-        controller->cycles_until_firing.get() == controller->min_cycles_needed() - 1;
-    
-    return ((was_await_pressurizing || was_idle) && is_time_to_pressurize && is_schedule_valid && is_functional);
+    return ((was_await_pressurizing || was_idle) && is_schedule_valid && is_functional);
 }
 
 prop_state_t PropState_Pressurizing::evaluate()
 {
+    // if our desired time to fire has already passed, early exit to idle to try again
```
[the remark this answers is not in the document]
tanishqaggarwal: I don't think this change is necessary. If cycles_until_firing is zero, then it just means that the prop state machine will pick up the next OrbitController-specified firing as soon as it is ready to fire (i.e. after it's done pressurizing.) Since the OrbitController should be continuously updating the requested firing, we should be good.

In general, I think a lot of the prop SM trouble is coming from the OrbitController having knowledge of the state machine.

src/fsw/FCCode/PropController.cpp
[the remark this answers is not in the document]
tanishqaggarwal: @kylekrol This change is required if we're removing a similar guard in AwaitPressurize::can_enter above.

I think the change is fine because the prop state machine simply won't fire until pressures are ready, so it's fine to let it at least _try_ to pressurize. Once it is pressurized, future pressurizations will be fast and don't need all of the cycles. So this change makes sense.

tanishqaggarwal (commented):
Basically confirming some of what Athena said + one additional suggestion

src/fsw/FCCode/OrbitController.cpp
shihaocao: @govindchari can you paste your code here?

govindchari: x=30;
y=6;
list=zeros(x,y);
for i=1:length(y)
    
    list(i)=(i-1)*(pi/90);
    if list(i)>pi
        list(i)=list(i)-2*pi;
    end
end
list

src/fsw/FCCode/OrbitController.cpp
shihaocao: I put this in because all the transitions from idle -> await_pressurizing -> pressurizing -> await_firing -> firing, require having a valid_schedule (meaning that all schedule elements are <1000. If we ever schedule something that has greater than 1000, then we'll never pass that check. But if we never pass that check, we'll never get to await_firing, so we'll never reschedule.

Thus. I added this else section to make sure we dont get stuck there, does this make sense?

shihaocao: This change is unrelated to the GNC push and is a "fix while here"

shihaocao: If anything, I might actually be inclined to remove this prop state is in await firing guard...? On line 172?
@govindchari @fatimayousuf

src/fsw/FCCode/PropController.cpp
```diff
@@ -282,11 +282,11 @@ bool PropState_AwaitPressurizing::can_enter() const
...
     bool is_schedule_valid = controller->validate_schedule();
     // Enter Await Pressurizing rather than Pressurizing if we have MORE than enough time
-    bool more_than_enough_time =
-        controller->cycles_until_firing.get() >= controller->min_cycles_needed();
+    // bool more_than_enough_time =
+    //     controller->cycles_until_firing.get() >= controller->min_cycles_needed();
     bool is_functional = PropulsionSystem.is_functional();
 
-    return (was_idle && is_schedule_valid && more_than_enough_time && is_functional);
+    return (was_idle && is_schedule_valid && is_functional);
```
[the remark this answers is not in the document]
shihaocao: Ah I see what you mean about how it will always be in await pressurizing. Should we perhaps add an additional check that cycles until firing are greater than zero or that the scheduled valve commands are greater than zero?

src/fsw/FCCode/PropController.cpp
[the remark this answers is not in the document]
shihaocao: I agree about looking down into the prop state machine, looking to see if we can clear those.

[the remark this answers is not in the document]
shihaocao: @tanishqaggarwal @athena255 we are worried about spending too long in pressurizing, and running out of cycles allocated to Pressurizing in ActionCycleOpenClose. This check puts us back in idle where we can wait for the next firing.

src/fsw/FCCode/OrbitController.cpp
shihaocao: @kylekrol @stewartaslan whats the longest duration a valve can be open? Can bad things happen if we reschedule valves once they've already been going?

src/fsw/FCCode/OrbitController.cpp
tanishqaggarwal: Suggestion: add a quick comment before the numbers containing a brief explanation

src/fsw/FCCode/OrbitController.cpp
tanishqaggarwal: Valves are open for max 1 s for some hardware reasons I can't remember. Likely has to do with S&H shenanigans

No, the hardware underneath is fundamentally stateless (it's either on or off)--but the valve schedule is copied over to the hardware driver (`PropulsionSystem`) during the await_firing state, and the driver works on this copy. So changing the state registry values mid-firing would do nothing. Why are you seeking this feature?

src/fsw/FCCode/PropController.cpp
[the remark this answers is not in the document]
tanishqaggarwal: If we're "almost" pressurized, ActionCycleOpenClose::evaluate shouldn't take too long to actually achieve the desired pressure. _We will not be losing cycles if the prop system hardware is still operational,_ since the only reason for losing cycles is a leak or reduced inner tank pressure due to significant use of propellant.

Is the concern something like this:
- The firing fed into the orbit controller is for node n
- We miss the firing at node n because the prop system isn't pressurized, so we should wait until the orbit controller assigns the requisite firing for node n+1, rather than getting "almost pressurized" quickly and then accidentally firing the impulse computed for node n

If so, (forgive my ignorance if this doesn't work) why isn't the OrbitController just continuously computing optimal firings at each timestep? What makes this difficult? And is the node n firing good enough that we could get by without a state machine change and still achieve rendezvous?

tanishqaggarwal (commented):
Sorry, I think I'm writing a little defensively--but I just want to be extra, extra careful about state machine changes when we're literally a week out from integration. I personally feel that it is better to reach for success in secondary mission objectives (demonstration of low-cost subsystems) than accidentally break the prop system in the hopes of achieving the primary objective (rendezvous).

src/fsw/FCCode/OrbitController.cpp
tanishqaggarwal: > I put this in because all the transitions from idle -> await_pressurizing -> pressurizing -> await_firing -> firing, require having a valid_schedule (meaning that all schedule elements are <1000. If we ever schedule something that has greater than 1000, then we'll never pass that check. But if we never pass that check, we'll never get to await_firing, so we'll never reschedule.
> 
> Thus. I added this else section to make sure we dont get stuck there, does this make sense?

Suppose the limit on valve firing is $L$, as long as it doesn't exceeed $L$ then this if-statement is unnecessary. I think the onus is on the orbit controller to saturate the firings so that this if statement is not required.

src/fsw/FCCode/OrbitController.cpp
shihaocao: @govindchari Can you implement this as a live function with a tunable number of firings per orbit or something to that nature

src/fsw/FCCode/OrbitController.cpp
```diff
@@ -90,10 +90,12 @@ void OrbitController::execute() {
     // If we don't have all the information we need, don't calculate a firing
     if (!time_valid_fp->get() || !orbit_valid_fp->get() || !rel_orbit_valid_fp->get() ||
             !attitude_estimator_valid_fp->get()) {
-        sched_valve1_f.set(0);
-        sched_valve2_f.set(0);
-        sched_valve3_f.set(0);
-        sched_valve4_f.set(0);
+        return;
```
shihaocao, line 93: "if the OrbitController doesn't have enough information to schedule valves properly, don't set them to 0, leave the values as is, early return"

src/fsw/FCCode/OrbitController.cpp
```diff
@@ -90,10 +90,12 @@ void OrbitController::execute() {
...
+    }
+    
+    // Don't fire unless we're in follower and follower close approach.
+    mission_state_t mission_state = static_cast<mission_state_t>(mission_state_fp->get());
+    if (!(mission_state == mission_state_t::follower || mission_state == mission_state_t::follower_close_approach)){
```
shihaocao, line 98: dont schedule if we're not in the right mission state

src/fsw/FCCode/PropController.cpp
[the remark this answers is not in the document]
shihaocao: Resolving this because, we evaluate pressurizing first, which has the masked effect that this block is still here.

src/fsw/FCCode/OrbitController.cpp
shihaocao: The best reason is to not schedule something too far into the future, as we get closer to the node our estimates of how far the node is in the future become better and more accurate.

src/fsw/FCCode/PropController.cpp
shihaocao: We added the above reset calls to ensure we are closing valves when leaving the pressurizing state.

src/fsw/FCCode/PropController.cpp
```diff
@@ -413,6 +418,7 @@ prop_state_t PropState_Pressurizing::handle_out_of_cycles()
     if (controller->pressurize_fail_fault_f.is_faulted())
     {
         DD("[!] --> transitioning to handling_fault\n");
+        PropulsionSystem.reset();
```
shihaocao, line 421: Adding another one just in case

src/fsw/FCCode/PropController.cpp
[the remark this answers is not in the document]
shihaocao: Marking as outdated. Changed condition to only allow transition into pressurizing if we're at or below threshold cycles. This means we don't pressurize too early, as before, but also allows us to pressurize even in rushed scenarios where we are trying to fire towards a node that is closer than the min cycles.

src/fsw/FCCode/PropController.cpp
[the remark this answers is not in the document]
shihaocao: Marking as a desired and tested feature. We do want the Pressurizing state to be able to leave to idle early if the firing node has passed or is too early.

src/fsw/FCCode/OrbitController.cpp
shihaocao: Removed this line change about zeroing out schedules. Marking as invalid comment.

src/fsw/FCCode/PropController.cpp
shihaocao: Will confirm with Athena that reset() closes all valves intertank and thrust valves

shihaocao: Marking as closed. Documentation says all valves will be closed

Duncan-McD (approved):
:+1:  looks good 2 me

PR #798, opened 2021-08-28
Node generator

govindchari:
# Node Generator

### Summary of changes
- Allows near field firing frequency to be controlled from the ground
- Adds node_generator function which generates a list of firing nodes
- Closes issue #799 
- Since the size of node list is 360, we can fire no faster than every 15 seconds, if we want to fire more frequently, the node list size needs to increase

### Telemetry
- Adds num_near_field_nodes as a writable state field


knk38 (commented):
Looks good to me and passes CI, pending @fatimayousuf's approval

fatimayousuf (approved):
:)

Issue #797, opened 2021-08-21
CI Timing Out Due to Persistant Flight Software Binary Threads


kylekrol:
TL;DR; PTest/thread handling in flight software sucks and isn't shutting down the flight software binary processes properly. Running all the mission checkouts locally on my desktop eventually pinned all 24 cores and grabbed 32 gigs of RAM plus 38 of swap crashing other processes running in the background -- like my chrome tabs.

Upgrade PTest to properly shut things down.

kylekrol:
So what I think something like this is happening once I added changes to handle signals properly:

1. Upon PTest termination we send `SIGTERM` to the main flight software process thread.
2. The main thread then tells the reader thread in `debug_console` to stop and blocks until that thread exits.
3. The reader thread however, is blocked waiting for another line of input and therefore never exits.
4. This means the main thread doesn't exit, and we just get stuck with extra processes left around the system.
5. Not really sure why this grabs more and more memory and pins the CPU however.

kylekrol:
Hmmm... so perhaps this is more of a problem on Linux than mac? I don't see the phantom flight software processes on MacOS -- just the downlink parser.

<img width="1124" alt="Screen Shot 2021-08-29 at 1 06 58 PM" src="https://user-images.githubusercontent.com/33558436/131258996-5aaa1155-8bd2-4ce9-bd2d-12de00a2bd81.png">

shihaocao:
Merging this should hopefully unblock other CI blocked PR's

shihaocao:
Implementation for PR copied from here:
https://stackoverflow.com/questions/15524122/how-to-implement-timeout-for-getline

shihaocao:
Added this BLAS limiter to pass CI, unsure if its spurred by this PR but it should be fine to add:
https://stackoverflow.com/questions/52026652/openblas-blas-thread-init-pthread-create-resource-temporarily-unavailable

shihaocao:
Closing per #801

PR #801, opened 2021-08-30
Add timeout in reader thread in debug_console

shihaocao:
Addresses: #797 Tested on Linux

### DIFF
--- a/src/common/debug_console.cpp
+++ b/src/common/debug_console.cpp
@@ -152,7 +152,7 @@ void debug_console::open() {
             // a timer that will kill the reader thread if we are no longer running the reader
             std::thread([&]
             {
-                std::this_thread::sleep_for(std::chrono::seconds(1));
+                std::this_thread::sleep_for(std::chrono::milliseconds(500));
 
                 if (!reader_thread_is_running){
                     std::terminate();
--- a/src/common/debug_console.cpp
+++ b/src/common/debug_console.cpp
@@ -14,6 +14,7 @@
     #include <iostream>
     #include <mutex>
     #include <thread>
+    #include <atomic>
 #else
     #include <Arduino.h>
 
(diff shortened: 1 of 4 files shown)


src/common/debug_console.cpp
```diff
@@ -67,7 +68,7 @@ static std::thread reader_thread;
 
 /** @brief Used to signal the reader thread to halt.
  */
-static volatile bool reader_thread_is_running = false;
+static std::atomic<bool> reader_thread_is_running(false);
```
shihaocao, line 71: I think atomic and volatile do the same thing in this context, but tested with atomic so keeping atomic.

angelaloh09:
running tests on macOS, not linux: 

testing emptycase (without ni)
Screen Shot 2021-08-29 at 11 58 42 PM


testing emptycase (with ni)
Screen Shot 2021-08-29 at 11 59 09 PM

shihaocao:
Looks good on the mac side, but it seems like on an unrelated note that the GSW parser binaries are also still hanging around lmaoooo

angelaloh09 (approved):

knk38:
Looks good, lets see if this fixes the other sun vector branch, however its worth noting that the sun vector PR passes CI without the new test case.

src/common/debug_console.cpp
```diff
@@ -147,6 +148,17 @@ void debug_console::open() {
...
+
+            // a timer that will kill the reader thread if we are no longer running the reader
+            std::thread([&]
+            {
+                std::this_thread::sleep_for(std::chrono::milliseconds(500));
+
+                if (!reader_thread_is_running){
+                    std::terminate();
+                }
+            }).detach();
```
kylekrol, line 160: @shihaocao We're spawning a new thread here every single time that this loop runs...

src/common/debug_console.cpp
kylekrol: actually with this implementation we now do need this because we tossed locks.

src/common/debug_console.cpp
shihaocao: @kylekrol we sure are! I haven't noticed any performance issues because of it. This should also only spawn as fast as we write to a thread which is at most like ~20/30 times a control cycle.

Issue #804, opened 2021-08-31
GSW Close Out


knk38:
- Currently we have all of the layouts set up but they need to be completely filled out with bounds
- Widgets need to be set up as well
- Since we added new stuff to GSW lets update that on the docs

Duncan-McD:
## Layouts:

 - Meeting must be made to decide what "Green" bounds are for important telem fields (particularly ones with widgets)
 - We also should, as an additional feature, come up with a formula for the mission status as a whole (Ex. Whether everything important is "Green") Which would allow us to have a front page to MCT/GSW with a Green or Red indicator light to indicated if each satellite is nominal. (This is not as critical as other widgets but would make our live's more convenient in end)
 - After meeting I can update the layout json using the OpenMCT editor/GUI

 #### Files needed to be changed:
 1. spreadsheet to organize bounds
 2. Layout .json file

 ## Documentation:

 - All the new procedures for ground software should be added to the documentation. 
 These Should include both:
  1.  user-end procedures 
   - How to import layout json files
   - how to login to MCT from browser
   - Until browser security/SSL/HSTS issues are fixed, add procedures to complete a bypass
  2. Server-end procedures
   - How to start up each piece of the server stack
   - How to diagnose server issues on scaleway end
   - Once the scripts are made to ease up server procedures, update this as well.
   
 #### Files needed to be changed:
 1. GSW ReadMe's
 2. Relevent Pprocs/Documentation

PR #809, opened 2021-09-07
GroundSoftware Startup Scripts using TMUX

### LINKED ISSUE #806: Server Process Persistence and SSL Fixes
1. Process persistence

Using tmux to set up gsw stack. I can also write a shell script to start up the tmux instances, for ease of first time start up.

2. SSL/Certificate Fixes

Self-Signed certificates get blocked by many browsers, so it is best to get a signed one from a CA (Certificate Authority). Most of these authorities cost money, however we should be able to get a free cert from Cloudflare or Let's Encrypt.

Duncan-McD:
# GroundSoftware Startup Scripts using TMUX

Fixes #806 (The last unfinished part - the rest were server configuration changes)

### Summary of changes
- Added Script for Starting the whole GSW stack
- This script built off of the different components in seperate scripts so if one crashes, we can start it really quick

### Testing
I tested on Ubuntu 20.04 and bash which is what the server uses

### Documentation Evidence
-  I added a lot of documentation to explain what the script does and how to get it all set up on a machine

NOTE:

Currently to startup GSW  on server you would have to run ./FlightSoftware/tools/GSW.sh (which could be annoying to remember exactly) So I will make an alias on the server to run this script and call it StartGSW or something of the like


shihaocao (approved):

shihaocao:
This is looks good to me! Instead of aliasing, what about adding it to the GSW readme?

Duncan-McD:
> This is looks good to me! Instead of aliasing, what about adding it to the GSW readme?

Yeah  Sounds good, I am going to do that in the Documentation PR im working on right now (for #804)

PR #814, opened 2021-09-25
Prop HITL

### LINKED ISSUE #802: Prop System Firing HITL
- Figure out a way we can actually complete an irl firing on the satellite
- Create a new PTestCase/ take elements from the current prop test cases that can perform these firings
- Use this test as a way to check a few prop functionalities/ the firing schedules/ Govind’s new Node generator

govindchari:
Before opening the PR, think: 
- Is this PR bulletproof? Can I think of no ways to improve this PR beyond its current state?
- Is the PR short enough to ensure a quick but thorough review, but long enough to represent a complete feature? For large state machines, let's try opening PRs for only a few states at a time.

If the answer to either of these questions is "no", refactor your code and open your PR when the answer to both questions is "yes".

--------------------------

# Prop HITL

Fixes #802 

### Summary of changes
- Adds ptest case for prop HITL
- Adds hootl_hitl.json config

### DIFF
--- a/ptest/cases/base/dual_sat_case.py
+++ b/ptest/cases/base/dual_sat_case.py
@@ -121,3 +121,7 @@ class DualSatCase(PTestCase):
             #    device.scrape_uplink()
             if device.enable_auto_dbtelem:
                 device.dbtelem()
+    def rs(self, name, print=False):
+        return self.follower.rs(name, print=print)
+    def print_rs(self, name):
+        return self.follower.rs(name, print=True)
--- a/ptest/cases/propsystem_firing.py
+++ b/ptest/cases/propsystem_firing.py
@@ -1,0 +1,31 @@
+from .utils import Enums, TestCaseFailure
+import time
+
+from .mission import (
+    DualSatNearFieldCase
+)
+
+class PropFiringCase(DualSatNearFieldCase):
+
+    def __init__(self, *args, **kwargs): #allows for interaction with the terminal during ptesting
+        super().__init__(*args, **kwargs)
+        self.debug_to_console = True
+
+    def run(self):
+        self.follower.ws("prop.threshold_firing_pressure", 10)
+        propNum = self.rs("prop.state")
+        self.print_rs("orbit.control.valve1")
+        while(propNum != 7):
+            self.print_rs("prop.state")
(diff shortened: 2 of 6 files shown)


ptest/configs/hootl_hitl.json
```diff
@@ -0,0 +1,25 @@
...
+    "seed" : 0,
+    "devices" : [
+        {
+            "name": "FlightControllerLeader",
+            "run_mode": "native",
+            "pio_target": "fsw_native_follower_autotelem",
+            "http_port" : 8089,
+            "imei": "412222222",
+            "scrape_uplinks": false,
+            "enable_auto_dbtelem": false
```
knk38, line 11: Not that this really matters, but this case works fine with telemetry on the hootl designated unit? I think I remember some issues with this

ptest/cases/propsystem_firing.py
```diff
@@ -0,0 +1,31 @@
...
+        propNum = self.rs("prop.state")
+        self.print_rs("orbit.control.valve1")
+        while(propNum != 7):
+            self.print_rs("prop.state")
+            propNum = self.rs("prop.state")
+            self.print_rs("orbit.control.valve1")
+            self.print_rs("orbit.valid")
+            self.print_rs("rel_orbit.state")
+            self.print_rs("prop.cycles_until_firing")
+            #self.ws("prop.state", 6) #get prop state into await firing mode
```
knk38, line 25: why was this commented out? (if its not necessary i think we can delete it)

ptest/cases/base/dual_sat_case.py
```diff
@@ -121,3 +121,7 @@ def cycle(self):
             #    device.scrape_uplink()
             if device.enable_auto_dbtelem:
                 device.dbtelem()
+    def rs(self, name, print=False):
```
knk38, line 124: I think it would be good to note that we are designating follower to be the satellite that matters in this environment and the fact that we can only rs follower states

knk38 (changes requested):
Looks good, just a few doc stuff/ questions I had

shihaocao:
I think we can close this PR as we did complete this testing, but did not merge into final code base.

PR #815, opened 2021-09-26
Change to rs

shihaocao:
- Add print_rs and print_ws back in
- Change battery level check case so that it works again to use `.rs()`


knk38 (approved):
Yup, looks good

PR #817, opened 2021-09-28
Telemetry phases

fatimayousuf:
# Reorder Telemetry Flows if Faulted

### Summary of changes
- If. fault is signaled, the flow with all the faults in it (flow 7) will move to the third position in the flow order
- Once the fault is not faulted anymore, it will move back to its original position in the flow order

### Testing
- See the unit test in test_fsw_downlink_producer
- Signal a fault and check that flow 7 moves to the 3rd position
- Unsignal the fault and check that it goes back to its original position in the flow order


src/fsw/FCCode/DownlinkProducer.cpp
```diff
@@ -179,6 +201,38 @@ void DownlinkProducer::execute() {
...
+                        if (!faults_flow_shifted){
+                            fault_idx = idx; // store the id of the flow with faults
+                            shift_flow_priorities_idx(7, 2);
+                            faults_flow_shifted = true;
+                        }
+                    }
+                }
+            }
+        }
+    }
```
shihaocao, line 229: this method looks really big, and seems a bit confusing?

We know that all faults are going to always be in the same flow, so what if we wrote a function that returns true if any fault is faulted

Then if the above function returns true, we shift the fault flow to the top?

src/fsw/FCCode/DownlinkProducer.cpp
```diff
@@ -179,6 +201,38 @@ void DownlinkProducer::execute() {
...
+                    std::string field = fields[i]->name();
+
+                    if (!field.compare(fault->name())){
+                        found_fault = true; // Signal that we found the flow with the faults in it and can stop looping
+
+                        // If we haven't already, shift the flow towards the top of the flow data list
+                        if (!faults_flow_shifted){
+                            fault_idx = idx; // store the id of the flow with faults
+                            shift_flow_priorities_idx(7, 2);
+                            faults_flow_shifted = true;
```
shihaocao, line 223: Can this logic be separated into its own function? Say some loop that searches through all flows, and returns as soon as it finds the flow that contains the "key" fault. Where the "key" fault is the first fault in the fault flow (I don't think we'll be changing whatever that fault is)

src/fsw/FCCode/DownlinkProducer.cpp
```diff
@@ -293,4 +347,37 @@ void DownlinkProducer::shift_flow_priorities(unsigned char id1, unsigned char id
             std::swap(flows[i],flows[i+1]);
         }
     }
+}
+
+void DownlinkProducer::shift_flow_priorities_idx(unsigned char id, size_t idx) {
+    if(id > flows.size()) {
+        printf(debug_severity::error, "Flow with ID %d was not found when "
+                                      "trying to shift to index %d.", id, idx);
+        assert(false);
```
shihaocao, line 356: lets take out this debug code, or at least wrap it in a #IFNDEF flight or something

src/fsw/FCCode/DownlinkProducer.hpp
```diff
@@ -151,6 +158,11 @@ class DownlinkProducer : public TimedControlTask<void> {
      */
     void shift_flow_priorities(unsigned char id1, unsigned char id2);
 
+    /**
+     * @brief 
+     */
```
shihaocao, line 163: Can we fill this out? Just a few sentences is good haha thanks

shihaocao (changes requested):
Functionality looks good, just wondering if it could be refactored a bit.

Question though: what happens in the future if a state change and we change the flow ordering for states? I guess we don't have to bake it into this PR, but we should update the ticket with how this system will be easily upgradable and wont interfere with state changes n stuff.

src/fsw/FCCode/DownlinkProducer.cpp
fatimayousuf: The faults will always be in the same flow, that's true. but we also need to know the original index position of this flow for when the fault is eventually unfaulted and we have to move it back. that's why we have all this loop stuff going on.

Also this method doesn't return anything, it will shift the flow whenever a fault in the active_faults list is faulted.

src/fsw/FCCode/DownlinkProducer.cpp
```diff
@@ -293,4 +347,37 @@ void DownlinkProducer::shift_flow_priorities(unsigned char id1, unsigned char id
...
+    }
+    if(idx > flows.size()) {
+        printf(debug_severity::error, "Flow in index %d was not found when "
+                                      "trying to shift with flow ID %d.", idx, id);
+        assert(false);
+    }
```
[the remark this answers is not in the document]
fatimayousuf: ok should I do the same for `shift_flow_priorities()`

src/fsw/FCCode/DownlinkProducer.hpp
fatimayousuf: oops forgot about that sorry

fatimayousuf:
> Functionality looks good, just wondering if it could be refactored a bit.
> 
> Question though: what happens in the future if a state change and we change the flow ordering for states? I guess we don't have to bake it into this PR, but we should update the ticket with how this system will be easily upgradable and wont interfere with state changes n stuff.

yeah so Angela and I changed this a bit since our last discussion. now, before shifting a flow upwards in the flow order, we save the original _index_ of the flow with faults (not the id of the flow behind it). We also wrote a new function to shift the flow with a given flow id to a new index (as opposed to taking two flow ids as parameters). So this helps us get around changes in the telemetry flow ordering due to changes in mission state.

knk38:
Okay I'll wait until the next commit to add my review, but I agree with shihao's notes

src/fsw/FCCode/DownlinkProducer.cpp
```diff
@@ -13,6 +13,24 @@ DownlinkProducer::DownlinkProducer(StateFieldRegistry& r) : TimedControlTask<voi
     add_internal_field(snapshot_size_bytes_f);
 }
 
+void DownlinkProducer::init(){
+    active_faults = {
+        FIND_FAULT(adcs_monitor.wheel_pot_fault.base),  
+        FIND_FAULT(adcs_monitor.wheel3_fault.base),
+        FIND_FAULT(adcs_monitor.wheel2_fault.base),
+        FIND_FAULT(adcs_monitor.wheel1_fault.base), // this one isnt in flow_data.cpp
```
knk38, line 21: @shihaocao Any reason why this fault isnt in the flow_data.cpp vector? If not I think we should add lol

knk38 (approved):
looks good :)

shihaocao (approved):
Great stuff, love the refactor too

Issue #813, opened 2021-09-25
Bootcount is overwritten by EEPROM


shihaocao:
Two things to fix:
Lets loosen the restriction on MissionManager

Second lets change the computation of bootcount and its interaction with EEPROM so the statefield is still meaningful

shihaocao:
Post PR Verification:
(venv) shihao@shihao-T490V2:~/Code/PAN/FlightSoftware$ python -m ptest runsim -c ptest/configs/hitl.json -t EmptyCase
 _______     _       ____  _____                                
|_   __ \   / \     |_   \|_   _|                            
  | |__) | / _ \      |   \ | |                               
  |  ___/ / ___ \     | |\ \| |                              
 _| |_  _/ /   \ \_  _| |_\   |_                             
|_____||____| |____||_____|\____|                              
                                                                
Pathfinder for Autonomous Navigation                            
Space Systems Design Studio, Cornell University       
Console Software | Version 1.0

Opened connection to FlightController.
Running mission testcase EmptyCase.
[TESTCASE] Starting testcase.
[TESTCASE] Suppressing faults.
Turning off fault_handler
Turning off piksi_fh
Turning off qfh
Supressing gomspace.low_batt
Supressing gomspace.get_hk.base
Supressing prop.overpressured
Supressing prop.tank2_temp_high
Supressing prop.tank1_temp_high
Supressing adcs_monitor.functional_fault
Supressing adcs_monitor.wheel1_fault
Supressing adcs_monitor.wheel2_fault
Supressing adcs_monitor.wheel3_fault
Supressing adcs_monitor.wheel_pot_fault
Supressing attitude_estimator.fault
Supressing piksi_fh.dead
[TESTCASE] Auto-dbtelem INACTIVE
[TESTCASE] Scrape-uplink ACTIVE
[TESTCASE] Boot utility waiting to reach initial state.
Beginning console.
Type "help" for a list of commands.
NOTE: You are currently connected to the FlightController.
> [TESTCASE] Finished testcase.

> rs pan.bootcount
1                                               (Completed in 5596 us)
> exit
Exiting command line.
Stopping binary monitor thread...
Stopping uplink console...
Terminating 0 radio connection(s)...
Terminating 1 device connection(s)...
 - Terminating console connection to and saving logging/telemetry data for FlightController.
 - Finished terminating for FlightController.
[...]

PR #820, opened 2021-09-29
Bootcount

shihaocao:
Why?
- To address: #813 

On this PR:
- Updated boot count to be incremented in the init function
- Updated dispatch startup early return to be based solely on boot timer

Testing:
- WIP

### DIFF
--- a/src/fsw/FCCode/MissionManager.cpp
+++ b/src/fsw/FCCode/MissionManager.cpp
@@ -56,7 +56,6 @@ MissionManager::MissionManager(StateFieldRegistry &registry)
     add_writable_field(kill_switch_f);
 
     bootcount_fp = find_readable_field<unsigned int>("pan.bootcount", __FILE__, __LINE__);
-    bootcount_fp->set(bootcount_fp->get() + 1);
 
     static_cast<MainFaultHandler *>(main_fault_handler.get())->init();
 
@@ -94,6 +93,10 @@ MissionManager::MissionManager(StateFieldRegistry &registry)
     is_deployed_f.set(bootcount_fp->get() > 1);
     deployment_wait_elapsed_f.set(0);
     set(sat_designation_t::undecided);
+}
+
+void MissionManager::init(){
+    bootcount_fp->set(bootcount_fp->get() + 1);
 }
 
 void MissionManager::execute()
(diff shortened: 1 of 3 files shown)


src/fsw/FCCode/MissionManager.cpp
```diff
@@ -182,12 +185,10 @@ void MissionManager::dispatch_startup()
     }
 
     // Step 1. Wait for the deployment timer length. Skip if bootcount > 1.
-    if (bootcount_fp->get() == 1) {
-        if (deployment_wait_elapsed_f.get() < deployment_wait)
-        {
-            deployment_wait_elapsed_f.set(deployment_wait_elapsed_f.get() + 1);
-            return;
-        }
+    if (deployment_wait_elapsed_f.get() < deployment_wait)
```
tanishqaggarwal, line 188: Why would you remove this check?? If you do this, then every single time the spacecraft retarts it'll go through the deployment wait.

src/fsw/FCCode/MissionManager.cpp
```diff
@@ -182,12 +185,10 @@ void MissionManager::dispatch_startup()
     }
 
     // Step 1. Wait for the deployment timer length. Skip if bootcount > 1.
```
afoarce, line 187: I don't think this comment is updated

telemetry
```diff
@@ -4,7 +4,7 @@
         "attitude_estimator.mag_flag": 1,
         "cycle.auto": 1,
         "cycle.start": 1,
-        "pan.bootcount": 1000,
```
afoarce, line 7: what is this change doing/what's the reasoning?

src/fsw/FCCode/MissionManager.cpp
shihaocao: Deployment wait is still an EEPROM saved field, so I think we still get the behavior we want with regard to not waiting more than once.

Critically though, without restriction to bootcount, we make sure to wait for the full duration even if we reboot during startup.

knk38: @tanishqaggarwal  ^ i think this makes sense?

afoarce (approved):

PR #821, opened 2021-09-29
check initial state

Duncan-McD:
# Check Initial Conditions Property

Fixes #. N/A

### Summary of changes
- Adds Check initial state property to SingleSatCase and DualSatCase
- Added it as a check in while loop
- Added setting it to False in EmptyCase


shihaocao (approved):

shihaocao:
good stuff

PR #822, opened 2021-10-02
Full Mission Case

### LINKED ISSUE #811: Full Duration Checkout Case
A full-duration checkout case no speedups, check critical functionality like Piksi off, radio enabling, and getting to near field, all from the deployment state.

Duncan-McD:
# Full Mission Case

Fixes #811 

### Summary of changes
- Added a full mission case that starts from deployment and then goes through each state.
- Piksi_off is also validated post boot
- The radio is first disabled and then set to wait in near field to verify the radio can be disabled and enabled
- Deployment wait is not skipped/Speedup flag not on

### Testing

Ran in hootl (with deployment wait skipped) and it reached Near Field with no assertions being thrown.

### Documentation Evidence
- I added some comments to designate the flow of the ptest case and indicate the reasoning

### DIFF
--- a/ptest/cases/full_mission_case.py
+++ b/ptest/cases/full_mission_case.py
@@ -1,0 +1,108 @@
+from .base import DualSatCase, PSimCase
+from .utils import Enums, TestCaseFailure
+
+class FullMissionCase(DualSatCase, PSimCase):
+
+  firstStartup = True
+  firstDetumble = True
+  firstStandby = True
+  firstFarField = True
+  firstNearField = True
+
+  def __init__(self, *args, **kwargs):
+    super(FullMissionCase, self).__init__(*args, **kwargs)
+    self.leader_skip_deployment_wait = False
+    self.follower_skip_deployment_wait = False
+    self.psim_configs += ["truth/deployment"]
+    
+  def post_boot(self):
+    """
+      This prints whether the piksi_off is valid after booting
+    """
+    leader_piksi_off = self.leader.print_rs("gomspace.piksi_off")
+    self.soft_assert( leader_piksi_off == True, f"Leader Piksi_off is {leader_piksi_off} when it should be True")
+    follower_piksi_off = self.follower.print_rs("gomspace.piksi_off")
+    self.soft_assert( follower_piksi_off == True, f"Follower Piksi_off is {follower_piksi_off} when it should be True")
+
+  def run(self):
+    
+    # During each run, run different actions based on the current follower and leader states
+    self.full_mission_checkout(self.follower.rs("pan.state"), self.leader.rs("pan.state"))
+    self.cycle()
+
+    
+  def full_mission_checkout(self, followerState, leaderState):
+    
+    # 1. Startup
+    if followerState == 0 and leaderState == 0: 
+      if self.firstStartup:
+        self.logger.put("Follower and Leader Starting Up")
+        self.firstStartup = False
+
+    # 2. Detumble
+    elif followerState == 1 and leaderState == 1:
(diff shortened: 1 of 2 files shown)


ptest/cases/full_mission_case.py
```diff
@@ -0,0 +1,108 @@
+from .base import DualSatCase, PSimCase
+from .utils import Enums, TestCaseFailure
+
+class FullMissionCase(DualSatCase, PSimCase):
+
+  firstStartup = True
+  firstDetumble = True
+  firstStandby = True
+  firstFarField = True
+  firstNearField = True
```
knk38, line 10: We dont need these booleans since these were for the startup/safehold case where we would go through startup->standby twice. This case should only go through startup->near field once

ptest/cases/full_mission_case.py
```diff
@@ -0,0 +1,108 @@
...
+    # During each run, run different actions based on the current follower and leader states
+    self.full_mission_checkout(self.follower.rs("pan.state"), self.leader.rs("pan.state"))
+    self.cycle()
+
+    
+  def full_mission_checkout(self, followerState, leaderState):
+    
+    # 1. Startup
+    if followerState == 0 and leaderState == 0: 
+      if self.firstStartup:
```
knk38, line 38: Same thing here, we can remove the boolean logic that keeps track of the first shift to startup (same for the other states below)

ptest/cases/full_mission_case.py
```diff
@@ -0,0 +1,108 @@
...
+        self.firstStartup = False
+
+    # 2. Detumble
+    elif followerState == 1 and leaderState == 1:
+      if self.firstDetumble:
+        self.logger.put("Follower and Leader Detumbling")
+        self.firstDetumble = False
+
+    # 3. Standby
+    elif followerState == 3 and leaderState == 3:
```
knk38, line 49: Dont know if this matters too much, but in the case one sat detumbles faster than the other we would miss some cycles in standby with this logic.

knk38 (changes requested):
Looks pretty good, just some minor changes to get rid of code that isn't necessary

ptest/cases/full_mission_case.py
Duncan-McD: This is so it only prints "follower and leader are in _____" the first time and doesn't spam that out every iteration of run

ptest/cases/full_mission_case.py
Duncan-McD: Well this will run the first time it sees both of them are in standby

ptest/cases/full_mission_case.py
Duncan-McD: Same reasoning as other reply - it's so we don't spam print the console

ptest/cases/full_mission_case.py
knk38: LOL wait im dumb i literally wrote this code in a diff case

knk38 (approved):
Looks good! :) I think we should hitl test it today if possible and then go ahead and merge

shihaocao:
This case looks good. Can you just remove the commented-out commands, and upload the logs of this test case passing here?
 
Make sure to rebase as well.

shihaocao (approved):

shihaocao (changes requested):
Assert that radio is disabled during startup, clean up commented out commands, upload any logs if they exist of the older HITL tests if you have them.

Lastly, this test should definitely be added to the github workflow! :D

shihaocao:
We can merge this PR if you'd like @Duncan-McD

PR #824, opened 2021-10-02
Mission state telemetry

fatimayousuf:
# Reorder telemetry based on Mission State

### Summary of changes
- If we go into follower or follower close approach, then move the prop related information to the 13th position in the flow order
- Otherwise, just use default flow order

### Testing
Added a HOOTL unit test that checks if we move the flows around appropriately when changing mission state.


knk38 (approved):
changes make sense, looks good to me

angelaloh09:
> If you haven't already, I would strongly suggest that you guys test this in integration. Have you all tested reordering flows through operator commands before? There is a strong chance that reordering flows may break the downlink parser.

@tanishqaggarwal Downlink parser processes the flows based on ID. This PR just shifts the flows around (orders the flows that are produced differently), and doesn't change the fields corresponding to specific flow ids.

tanishqaggarwal:
FYI: We usually use "HOOTL" to refer to an integrated test with all of flight software running. I would just call the change to test_downlink_producer a "unit test", and the test that Angela posted a HOOTL.

What's the purpose of this change? This is a simple enough thing to just command in flight that I would recommend that you guys just do that, rather than making a C++ change. There is immense risk with code changes this late in the game.

PR #828, opened 2021-10-12
Piksi power cycle addition and gomspace checkout case fix

shihaocao:
# Piksi Power Cycle Addition and Gomspace Checkout Case Fix

### Problems, This PR identifies and fixes two bugs.
- Firstly the Piksi is never power cycled to be on after leaving the startup state, the fix is to request a power cycle of the Piksi on leaving startup.
- Secondly deployment elapsed was counted as an unsigned int, and therefore could not be restored from memory. The fix was to compress its range into an unsigned char, where an increment represents an elapsing of 17 seconds, or 100 control cycles.

Lastly, to fix the Gomspace Checkout case as a whole, it was refactored to be cleaner, and put in additional checks and waits to compensate for the Piksi turning on after startup.

### DIFF
--- a/ptest/cases/base/psim_case.py
+++ b/ptest/cases/base/psim_case.py
@@ -175,6 +175,12 @@ class PSimCase(PTestCase):
         else:
             sensors["adcs_monitor.ssa_mode"] = Enums.ssa_modes['SSA_FAILURE']
 
+        # Simulate wheel speeds
+        #
+        # As wheel speeds are required for magnetometer data offsets now,
+        # wheel speeds must be simmed
+        sensors['adcs_monitor.rwa_speed_rd'] = list(self.__sim[f'truth.{satellite}.wheels.w'])
+
         return sensors
 
     def rs_psim(self, name: str):


ptest/cases/gomspace_checkout_case.py
```diff
@@ -91,55 +125,61 @@ def run(self):
...
-        # start power cycle
-        power_cycle_output_cmd = [self.str_to_bool(self.write_state("gomspace.power_cycle_output"
-                                                               + str(i) + "_cmd", "true"))
-                                  for i in range(1, 7)]
-        self.cycle()
-        while int(self.read_state("pan.cycle_no")) % int(self.read_state("gomspace.powercycle_period")) != 0:
-            output = [self.str_to_bool(self.read_state("gomspace.output.output" + str(i)))
-                      for i in range(1, 7)]
-            self.logger.put("current output: " + str(output))
+        if self.flight_controller.is_teensy:
```
shihaocao, line 128: Only run these cycle output tests on teensy as they don't work in HOOTL

ptest/cases/piksi_checkout_case.py
```diff
@@ -8,6 +8,7 @@ def __init__(self, *args, **kwargs):
         super(PiksiCheckoutCase, self).__init__(*args, **kwargs)
 
         self.debug_to_console = True
+        self.check_initial_state = False
```
shihaocao, line 11: Allows the Piksi checkout case to be run regardless of the state of the spacecraft.

ptest/cases/reset_for_flight.py
```diff
@@ -3,6 +3,12 @@
 import math
 
 class ResetforFlight(SingleSatCase):
+    def __init__(self, *args, **kwargs):
```
shihaocao, line 6: These changes are so that we can reboot regardless of the initial state of the spacecraft.

src/fsw/FCCode/FieldCreatorTask.hpp
```diff
@@ -18,11 +18,11 @@
 class FieldCreatorTask : public ControlTask<void>
 {
 public:
-  ReadableStateField<unsigned int> bootcount_f;
+  ReadableStateField<unsigned char> bootcount_f;
```
shihaocao, line 21: changing bootcount to be a char since this is the largest we can store. We accept overflow because this is purely a data point field.

src/fsw/FCCode/FieldCreatorTask.hpp
```diff
@@ -18,11 +18,11 @@
...
 
   FieldCreatorTask(StateFieldRegistry &r)
       : ControlTask<void>(r),
-        bootcount_f("pan.bootcount", Serializer<unsigned int>(0xfffffff), 100)
+        bootcount_f("pan.bootcount", Serializer<unsigned char>(), 100)
```
shihaocao, line 25: changing to default serializer.

ptest/cases/gomspace_checkout_case.py
```diff
@@ -1,5 +1,9 @@
 # Gomspace test case. Gets cycle count purely for diagnostic purposes and logs
```
tanishqaggarwal, line 1: Can you post output from this case? It will be easier to review the changes if I can read the output which describes what is being done.

ptest/cases/gomspace_checkout_case.py
```diff
@@ -163,25 +203,32 @@ def run(self):
...
-            self.failed = True
-
-        gs_reboot_cmd = self.str_to_bool(
-            self.read_state("gomspace.gs_reboot_cmd"))
-        gs_reboot_cmd_updated = self.str_to_bool(self.write_state(
-            "gomspace.gs_reboot_cmd", not gs_reboot_cmd))
-        if gs_reboot_cmd == gs_reboot_cmd_updated:
-            self.logger.put("Could not update gs_reboot")
-            self.failed = True
+        # Dont run these sections as they cause the gomspace to power cycle or reboot
```
tanishqaggarwal, line 206: We absolutely want to be testing this, since it's a behavior we may need to command in flight. Are we testing it somewhere else, manually?

ptest/cases/base/dual_sat_case.py
```diff
@@ -68,11 +68,11 @@ def setup(self, devices, radios):
 
         if self.leader_skip_deployment_wait:
             self.logger.put("[TESTCASE] Skipping leader deployment wait.")
-            self.flight_controller_leader.write_state("pan.deployment.elapsed", "15000")
+            self.flight_controller_leader.write_state("pan.deployment.elapsed", "106")
```
tanishqaggarwal, line 71: I suggest writing out a quick python function to compute this number, because otherwise you might forget what it means.

ptest/cases/gomspace_checkout_case.py
```diff
@@ -25,11 +35,36 @@ def str_to_bool(self, string):
...
+        '''Get all the current output states, and return a list of boolean current states'''
+        outputs = [self.read_state(f'gomspace.power_cycle_output' + str(i) + '_cmd') for i in range(1, 7)]
+        outputs = [self.str_to_bool(x) for x in outputs]
+        self.logger.put(f'Outputs Commands: {outputs}')
+        return outputs
+
     def run(self):
-        self.flight_controller.write_state('gomspace.piksi_off', False)
+        self.ws('pan.state', 0)
+        self.ws('pan.deployment.elapsed', 106) # set it to post deployment
```
tanishqaggarwal, line 54: Don't hardcode this number

ptest/cases/gomspace_checkout_case.py
```diff
@@ -25,11 +35,36 @@ def str_to_bool(self, string):
...
+        self.ws('gomspace.get_hk.suppress', False)
```
tanishqaggarwal, line 55: Can you add a comment for why the suppression is necessary? What behavior results if you leave it unsuppressed?

PR #830, opened 2021-10-15
Exceed Covariance Norm 

samuelqian01:
Modified Attitude_estimator ptest case and supporting files for a scenario that has been constructed (nothing written inside) for exceeding covariance norm. Added logic in attitude_estimator suite of files to trigger a reset when state field attitude_estimator_reset_persistance_reached becomes true (which triggers exceed_persistence), which could first be tested by manually triggering it in the ptest case but then adding the logic within attitude_estimator to trigger upon reaching some covariance.

### DIFF
--- a/telemetry
+++ b/telemetry
@@ -2,7 +2,6 @@
     "eeprom_saved_fields": {
         "attitude_estimator.ignore_sun_vectors": 1,
         "attitude_estimator.mag_flag": 1,
-        "attitude_estimator.reset_persistance_reached": 1,
         "cycle.auto": 1,
         "cycle.start": 1,
         "pan.bootcount": 100,
@@ -1154,10 +1153,20 @@
             "type": "bool",
             "writable": true
         },
-        "attitude_estimator.reset_persistance_reached": {
-            "bitsize": 1,
+        "attitude_estimator.reset_persistance": {
+            "bitsize": 10,
             "flow_id": "undefined",
-            "type": "bool",
+            "max": 1000,
+            "min": 0,
+            "type": "unsigned int",
+            "writable": true
+        },
+        "attitude_estimator.safety": {
+            "bitsize": 20,
+            "flow_id": "undefined",
+            "max": 1000000,
+            "min": 0,
+            "type": "unsigned int",
             "writable": true
         },
         "attitude_estimator.valid": {


.github/workflows/hootl.yml
```diff
@@ -56,6 +56,7 @@ jobs:
         !(python -m ptest runsim -c ptest/configs/hootl_speedup.json -t FailingEmptyCase -ni)
         !(python -m ptest runsim -c ptest/configs/hootl_speedup.json -t FailingEmptySimCase -ni)
         python -m ptest runsim -c ptest/configs/hootl.json -t AttitudeNoSunVectorsInitializationCase -ni
+        python -m ptest runsim -c ptest/configs/hootl.json -t AttitudePersistanceExceededCase - ni
```
afoarce, line 59: remove space `-ni`

src/fsw/FCCode/Estimators/AttitudeEstimator.cpp
```diff
@@ -217,7 +221,7 @@ void AttitudeEstimator::_execute()
             gnc::attitude_estimator_reset(
                     _state, time_s, {0.0f, 0.0f, 0.0f, 1.0f});
         }
-        else if (adcs_ssa_valid)
+        else if ((adcs_ssa_valid) || (exceed_persistance))
```
afoarce, line 224: Is this the logic we want? a comment explaining the logic/reasoning would be helpful here

install.sh
```diff
@@ -0,0 +1,855 @@
+#!/bin/bash
```
afoarce, line 1: don't think we need this file

afoarce (changes requested):

src/fsw/FCCode/Estimators/AttitudeEstimator.cpp
samuelqian01: After a closer look at the code I realized I just wrote everything that had already been written. I'm probably going to end up deleting most of the changes to attitude_estimator.cpp because I can leverage adcs_ssa_valid/change it to accommodate for accounting the Frobenius norm.

src/fsw/FCCode/Estimators/AttitudeEstimator.cpp
```diff
@@ -43,6 +44,7 @@ AttitudeEstimator::AttitudeEstimator(StateFieldRegistry &registry)
       attitude_estimator_reset_cmd_f("attitude_estimator.reset_cmd", Serializer<bool>()),
       attitude_estimator_mag_flag_f("attitude_estimator.mag_flag", Serializer<bool>(), 1),
       attitude_estimator_ignore_sun_vectors_f("attitude_estimator.ignore_sun_vectors", Serializer<bool>(), 1),
+      attitude_estimator_reset_persistance_reached("attitude_estimator.reset_persistance_reached", Serializer<bool>(), 0),
```
knk38, line 47: why an eeprom save field value of zero here?

src/fsw/FCCode/Estimators/AttitudeEstimator.cpp
```diff
@@ -197,6 +201,33 @@ void AttitudeEstimator::_execute()
...
+
+    /* One if statement for no ssa covariance and one if statement for ssa valid covariance */
+
+    if (!adcs_ssa_valid && exceed_persistance) {
+        gnc::attitude_estimator_reset(_state, time_s, {0.0f, 0.0f, 0.0f, 1.0f});
+    }
+    else if (adcs_ssa_valid && exceed_persistance) {
+        gnc::attitude_estimator_reset(_state, time_s, orbit_pos, b_body, adcs_ssa);
+    }
+
```
shihaocao, line 230: @govindchari

samuelqian01:
Screen Shot 2021-10-16 at 6 08 34 PM
Screen Shot 2021-10-16 at 6 08 22 PM

We added persistence logic, created a writeable state field for the safety factor, and modified our ptest case to trigger the gnc::attitude_estimator_reset. The pictures above show the print statements we used to verify that a reset had occurred, and we tested by setting the safety factor extremely low to guarantee the triggering of exceed_persistance.

shihaocao:
We should close this PR as we unfortunately didn't have time to add this to the final FSW package.

PR #846, opened 2021-12-03
Security and Navigation of GSW Sites

Duncan-McD:
# Security and Navigation of GSW Sites

Completed Security features for server, as well as a front end page to link together all of the ground software pages

### Summary of changes

New Changes to the repo:

- Added a basic react application that serves as the core home page to all of the other pages, such as MCT, and the swagger endpoints. (I chose react as this will make it easy to add onto if we decide to add more pages or make it fancier looking in the future)
- I removed old ssl certificate sections in code as these are no longer handled in the individual processes but instead in the overall server (by the proxy)
- I removed old password authentication methods for MCT as a more secure method was available with the HAProxy Server that could be used across all processes.
- created a folder to hold the HAProxy configuration files necessary to handle all the domain names and page redirections.
- Added the most up to date version of the MCT Layouts to the MCT layouts folder so it is readily accessible for download.
As a note, we should make a more finalized version before launch as there are still some leftover features that we were hoping to get to in the current version that we have to decide on.

New Changes on the Server:

- Authentication will now be required before acessing each main process (each swagger endpoint, the home page, and mct) Once you log in once for a process you wont have to again in the browser session.) I will look into if there is an easy way to have it only require one login for all four of the processes, but for now this is good as it gives us the security we want despite being a bit repetitive.
- A reverse proxy server is now runnning using HAProxy. This should redirect all requests from certain domain names to their corresponding processes and backend servers running on the server.
- Let's Encrypt and Certbot is set up to automatically renew our SSL Certificates for all our domains, so we should be able to have HTTPS connections across all of our pages, including the swagger endpoints.

### Testing

I was able to successfully access all of the pages from their domain names and recieve both password protection and Secure Connections.

PANMISSIONCONTROL

### Documentation Evidence
- I am going to add some documentation for these changes in a bit, but I wanted to open the PR now, since it is important for Mission Rehearsals


shihaocao (approved):
Looks good to me! Awesome stuff :)

merge pending demo/testing

afoarce (approved):

Issue #862, opened 2022-02-09
Downlink Shift Crashes FSW on HITL and x86 Linux


shihaocao:
Using this specific uplink, because #863 is a confounding issue/issue ticket, we observe that certain downlink shifts cause a std::move to segfault the spacecraft.

**Note that shifting flow 36 with flow 2 does not cause a segfault!! Therefore, some parts of flow shifting work...
Note further that this issue affects HITL, and x86 Linux.**

Linux Version:
```
(venv) shihao@shihao-T490V2:~/Code/PAN/FlightSoftware$ lsb_release -a
No LSB modules are available.
Distributor ID:	Ubuntu
Description:	Ubuntu 20.04.3 LTS
Release:	20.04
Codename:	focal
(venv) shihao@shihao-T490V2:~/Code/PAN/FlightSoftware$ uname -r
5.10.0-1057-oem
(venv) shihao@shihao-T490V2:~/Code/PAN/FlightSoftware$ 
```
Observed on EDU running `fsw_teensy36_hitl_leader`

**THIS ISSUE DOES NOT AFFECT x86 MacOS???**

Example uplink that crashes FSW:
```
[
    {
    "field": "radio.max_wait",
    "value": 10420
    },
    {
    "field": "radio.max_transceive",
    "value": 10
    },
    {
    "field": "adcs_cmd.havt_disable18",
    "value": true
    },
    {
    "field": "gomspace.piksi_off",
    "value": true
    },
    {
    "field": "dcdc.disable_cmd",
    "value": true
    },
    {
    "field": "gomspace.power_cycle_output4_cmd",
    "value": true
    },
    {
    "field": "downlink.shift_id1",
    "value": 19
    },
    {
    "field": "downlink.shift_id2",
    "value": 2
    },
    {
    "field": "pan.state",
    "value": 1  
    },
    {
    "field": "adcs.state",
    "value": 5
    },
    {
    "field": "adcs_monitor.wheel1_fault.suppress",
    "value": true
    },
    {
    "field": "adcs_monitor.wheel2_fault.suppress",
    "value": true
    },
    {
    "field": "adcs_monitor.wheel3_fault.suppress",
    "value": true
    },
    {
    "field": "adcs_monitor.wheel_pot_fault.suppress",
    "value": true
    },
    {
    "field": "gomspace.low_batt.suppress",
    "value": true
    },
    {
    "field": "attitude_estimator.fault.suppress",
    "value": true      
[...]

PR #867, opened 2022-02-13
Replace swap with triple line tmp swap

shihaocao:
This is a hotfix to address #862. The problematic command in issue ticket 862 now works on x86 Linux.

## On this PR:
Replace the `std::swap` call with a triple line copy to a temporary variable.


shihaocao:
This screenshot shows a successful swap after swapping 19 up to position 2.
image

This uplink applied was the one linked in the issue ticket.

